PR #133, opened 2022-05-08
coverage tests p1

idiomatic-rust-com:
Hi,

This PR adds some code coverage tests.

Best regards,
Michal

### DIFF
--- a/coverage.sh
+++ b/coverage.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/sh
+
+if ! which grcov; then
+  echo "Error: grcov not found. Try |cargo install grcov|"
+  exit 1
+fi
+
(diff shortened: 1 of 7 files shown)


coverage.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/sh
```
juhaku, line 1: I would if not already latest from master, take a pull and move this file to the `scripts` folder :) after all that is the right place for these.

juhaku (commented):
I guess those tests will do for now, but generally tests that only exists for higher coverage report and actually tests nothing is not very useful IMO. And only makes future refactoring more painfull because there is unnecessary unit tests that need to suddenly be "fixed". 

Better approach is always to test behaviour and probably some of these could directly be tested through the the tests in `src/openapi.rs` or `src/openapi/schema.rs` etc. But if there are some behaviour that need to be tested with unit tests in module level its completely fine to have those tests there as well.

Issue #131, opened 2022-05-06
Support pkce with oauth2


jgramoll:
I am trying to use utoipa-swagger-ui with oauth2 (https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/oauth2.md). I need to be able to set the `usePkceWithAuthorizationCodeGrant` flag so that I can authorize with pkce. Is there currently a way to support this or something that could be added?

juhaku:
Yeah, unfortunately the configuration options via `utoipa-swagger-ui` is far limited at the moment. Currently there is no way to configure other than the url(s) for the api docs. 

Support for this could be added and is pretty straight forward to implement. I've had a plans to improve the Swagger UI configuration options but hadn't had time to work on this. I'll add a task to the project's kanban board about this so that it will be done some point in future. Currently there are already quite a few things to work on.

PR #134, opened 2022-05-09
feat(swagger-ui): add support for oauth config

jgramoll:
@juhaku not sure what you had in mind for supporting this. Welcome to any feedback

### DIFF
--- a/utoipa-swagger-ui/src/lib.rs
+++ b/utoipa-swagger-ui/src/lib.rs
@@ -229,19 +229,18 @@ impl SwaggerUi {
     ///
     /// Enable pkce with default client_id.
     /// ```rust
-    /// # use utoipa_swagger_ui::SwaggerUi;
+    /// # use utoipa_swagger_ui::{SwaggerUi, oauth};
     /// # use utoipa::OpenApi;
     /// # #[derive(OpenApi)]
     /// # #[openapi(handlers())]
     /// # struct ApiDoc;
     /// let swagger = SwaggerUi::new("/swagger-ui/{_:.*}")
     ///     .url("/api-doc/openapi.json", ApiDoc::openapi())
-    ///     .oauth(utoipa_swagger_ui::oauth::Config {
-    ///         client_id: Some(std::env::var("OKTA_CLIENT_ID").unwrap()),
-    ///         scopes: Some(vec![String::from("openid")]),
-    ///         use_pkce_with_authorization_code_grant: Some(true),
-    ///         ..Default::default()
-    ///     });
+    ///     .oauth(oauth::Config::new()
+    ///         .client_id("client-id")
+    ///         .scopes(vec![String::from("openid")])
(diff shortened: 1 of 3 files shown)


jgramoll:
Fixes #131

utoipa-swagger-ui/src/lib.rs
```diff
@@ -481,6 +517,7 @@ impl<'a> Config<'a> {
     pub fn new<I: IntoIterator<Item = U>, U: Into<Url<'a>>>(urls: I) -> Self {
         Self {
             urls: urls.into_iter().map(|url| url.into()).collect(),
+            oauth: None,
         }
     }
```
juhaku, line 522: I would add here a method `with_oauth_config` something like this.
```rust
pub fn with_oauth_config<I: IntoIterator<Item = U>, U: Into<Url<'a>>>(urls: I, 
   oauth_config: oauth::Config) -> Self {
        Self {
            urls: urls.into_iter().map(|url| url.into()).collect(),
            oauth: Some(oauth_config),
        }
}
```
This `Config` type is `non_exhaustive` for backwards compatibility purposes

>Outside of the defining crate, types annotated with non_exhaustive have limitations that preserve backwards compatibility when new fields or variants are added.
>
>Non-exhaustive types cannot be constructed outside of the defining crate:
>
>    Non-exhaustive variants ([struct](https://doc.rust-lang.org/reference/items/structs.html) or enum variant) cannot be constructed with a StructExpression (including with functional update syntax).
>    enum instances can be constructed.
https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute

It is in Rust a good idea to define outside looking API with `non_exhaustive` and instead provide constructors `new()` or / and accessor fields or builder pattern to construct objects. This way the API implementator is at control and not every change to the API nor adding / changing / deleting one field will be a breaking change for API consumer.

utoipa-swagger-ui/src/oauth.rs
```diff
@@ -0,0 +1,81 @@
+use std::collections::HashMap;
```
juhaku, line 1: This could also have some module level comments like
```rust
//! Implements Swagger UI [oauth configuration](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/oauth2.md) options.
```

juhaku (changes requested):
Yeah this is pretty much something I planned actually add. Later I will refactor this a bit and add the swagger-ui other configuration options as well but that is another PR then. 

Overall pretty good. Some minor things to check through though :)

jgramoll (commented):
Thanks for the great feedback

juhaku:
> Thanks for the great feedback

Welcome you can try running tests locally with `scripts/test.sh utoipa-swagger-ui` to get started with. And further use the commands found from that file to run specific tests if needed as well.

juhaku (approved):
Super :rocket:

PR #135, opened 2022-05-10
docs: fix spelling errors

jgramoll:


juhaku (approved):
Wow thanks. :rocket: :+1: This is the result (so many mistakes) with typing fast without proofreading. Well appreciated.

Issue #149, opened 2022-05-30
Inline schemas


kellpossible:
I'm dealing with an API that has a unique type for each response body, having them all as components is a bit of a headache, and they won't be re-used in other endpoints.  It would be nice to have a way to optionally inline schema definitions. Perhaps with a special wrapper like `inline()`, the `path` macro could insert the schema for a derived `Component` inline rather than attempting to reference it?

kellpossible:
For reference https://swagger.io/specification/#media-type-object `schema` property is optionally a `Schema Object` or a `Reference Object`. Perhaps there are some other areas of the `path` macro that would benefit from being able to inline schemas too.

kellpossible:
#153 implements this for the `path!()` macro, but additionally it would be good to have inline fields with the `Component` derive macro, based on a `#[component(inline)]` field attribute.

juhaku:
Yes thats true.

juhaku:
Closing this one since the PR has been already merged.

PR #153, opened 2022-06-01
Inline schemas for path macro

kellpossible:
+ Implement  #149 ability to specify component schemas inline in the `path!(...)` macro.
+ Fix bug with `ArrayBuilder` being instantiated with the wrong `ComponentType` by making the builder derive use the derived struct's `Default` implementation, instead of generating its own. This might also fix some unexpected bugs with other builders.

TODO:

- [x] Implement this for the `request_body` attribute. We might get it for free if I do this refactor https://github.com/juhaku/utoipa/pull/153#discussion_r886840821 because `RequestBodyAttr` contains `Type`.
- [x] Update documentation.
- [x] What happens with recursive data structures, perhaps leave a warning in the documentation about it? Update: it's impossible to have recursive currently because `Component` derive doesn't have the ability to inline yet.

### DIFF
--- a/utoipa-gen/src/component_type.rs
+++ b/utoipa-gen/src/component_type.rs
@@ -3,6 +3,7 @@ use std::fmt::Display;
 use quote::{quote, ToTokens};
 
 /// Tokenizes OpenAPI data type correctly according to the Rust type
+// TODO: see if this can't be unified with crate::Type
 pub struct ComponentType<'a, T: Display>(pub &'a T);
 
 impl<'a, T> ComponentType<'a, T>
--- a/utoipa-gen/src/component_type.rs
+++ b/utoipa-gen/src/component_type.rs
@@ -3,7 +3,6 @@ use std::fmt::Display;
 use quote::{quote, ToTokens};
 
 /// Tokenizes OpenAPI data type correctly according to the Rust type
-// TODO: see if this can't be unified with crate::Type
 pub struct ComponentType<'a, T: Display>(pub &'a T);
 
 impl<'a, T> ComponentType<'a, T>
--- a/utoipa-gen/src/lib.rs
+++ b/utoipa-gen/src/lib.rs
@@ -19,7 +19,7 @@ use proc_macro_error::{proc_macro_error, OptionExt, ResultExt};
 use quote::{quote, ToTokens, TokenStreamExt};
 use schema::into_params::IntoParams;
 
-use proc_macro2::{Group, Ident, Punct, Span, TokenStream as TokenStream2};
+use proc_macro2::{Group, Ident, Punct, TokenStream as TokenStream2};
 use syn::{
     bracketed,
     parse::{Parse, ParseBuffer, ParseStream},
@@ -1303,7 +1303,7 @@ impl Parse for Type<'_> {
             .parse::<ArrayOrOptionType>()
             .map_err(|error| {
                 syn::Error::new(
-                    Span::call_site(),
+                    error.span(),
                     format!("{}: {}", EXPECTED_TYPE_DEFINITION, error),
                 )
             })?
--- a/utoipa-gen/src/lib.rs
+++ b/utoipa-gen/src/lib.rs
@@ -1092,6 +1092,8 @@ impl Parse for Type<'_> {
 
 /// A place where a type defenition is required. The type schema can either be a referenced
 /// component or provided inline.
+// TODO: perhaps this could be rolled into `Type` with a new is_inline function.
+#[derive(Clone)]
 enum TypeDefinition<'a> {
     Component(Type<'a>),
     Inline(Type<'a>),
--- a/utoipa-gen/src/lib.rs
+++ b/utoipa-gen/src/lib.rs
@@ -7,7 +7,7 @@
 #![warn(missing_docs)]
 #![warn(rustdoc::broken_intra_doc_links)]
 
-use std::{borrow::Cow, mem, ops::Deref};
+use std::{borrow::Cow, mem, ops::Deref, any::type_name};
 
 use doc_comment::CommentAttributes;
 use schema::component::Component;
@@ -20,13 +20,13 @@ use quote::{quote, ToTokens, TokenStreamExt};
 #[cfg(feature = "actix_extras")]
 use schema::into_params::IntoParams;
 
-use proc_macro2::{Group, Ident, Punct, TokenStream as TokenStream2};
+use proc_macro2::{Group, Ident, Punct, TokenStream as TokenStream2, Span};
 use syn::{
     bracketed,
     parse::{Parse, ParseBuffer, ParseStream},
     punctuated::Punctuated,
     token::Bracket,
-    DeriveInput, ExprPath, ItemFn, Lit, LitStr, Token,
+    DeriveInput, ExprPath, ItemFn, Lit, LitStr, MetaList, Token, NestedMeta, Meta,
 };
 
 mod component_type;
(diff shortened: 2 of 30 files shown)


utoipa-gen/src/component_type.rs
```diff
@@ -3,6 +3,7 @@ use std::fmt::Display;
 use quote::{quote, ToTokens};
 
 /// Tokenizes OpenAPI data type correctly according to the Rust type
+// TODO: see if this can't be unified with crate::Type
```
kellpossible, line 6: See if this type can be unified with `utoipa_gen::Type`.

utoipa-gen/src/path/property.rs
```diff
@@ -1,59 +1,68 @@
...
+    component_type::{ComponentFormat, ComponentType},
+    TypeDefinition,
+};
 
 /// Tokenizable object property. It is used as a object property for components or as property
 /// of request or response body or response header.
-pub(crate) struct Property<'a, T: Display> {
-    pub(crate) is_array: bool,
-    pub(crate) component_type: ComponentType<'a, T>,
+// TODO: Switch this to support either references or inline definition.
```
kellpossible, line 13: Remove this TODO

utoipa-gen/src/path/response.rs
```diff
@@ -111,31 +101,37 @@ impl ToTokens for Response<'_> {
...
+                    } else {
+                        let default_type =
+                            self.resolve_content_type(None, &property.component_type());
+                        tokens.extend(quote! {
+                            .content(#default_type, #content.build())
+                        });
+                    }
+                }
+                TypeDefinition::Inline(_body_type) => {
+                    todo!()
```
kellpossible, line 133: implement inline schema rendering

utoipa-gen/src/lib.rs
```diff
@@ -1041,6 +1042,66 @@ impl Parse for Type<'_> {
     }
 }
 
+/// A place where a type defenition is required. The type schema can either be a referenced
+/// component or provided inline.
+// TODO: perhaps this could be rolled into `Type` with a new is_inline function.
```
kellpossible, line 1047: Resolve this TODO, perhaps `TypeDefinition` can be rolled into `Type`

utoipa-gen/src/path/property.rs
```diff
@@ -1,59 +1,80 @@
...
-                .to_array_builder()
-            });
+                    if type_definition_type.is_array {
+                        tokens.extend(quote! {
+                            .to_array_builder()
+                        });
+                    }
+                }
+                TypeDefinition::Inline(_) => tokens.extend(quote! {
+                    <#component_name_ident as utoipa::Component>::component()
```
kellpossible, line 75: What should happen here when `type_definition_type.is_array == true`?

kellpossible:
@juhaku pending updating the documentation for this feature, this PR is ready for review!

utoipa-gen/src/lib.rs
```diff
@@ -1081,11 +1142,38 @@ impl Parse for Type<'_> {
...
+            let t: Self = input.parse::<InlineType>()?.0;
+            return Ok(t);
+        }
+
+        // Try parsing as `Type`, `[Type]` or `Option<Type>`)
+        let t: Type = input
+            .parse::<ArrayOrOptionType>()
+            .map_err(|error| {
+                syn::Error::new(
+                    Span::call_site(),
```
juhaku, line 1170: Returned error will be a syn error which contains span, it focuses the error better for user. :)
```suggestion
                    error.span(),
```

utoipa-gen/src/path/property.rs
```diff
@@ -1,59 +1,82 @@
...
+    Type,
+};
 
 /// Tokenizable object property. It is used as a object property for components or as property
 /// of request or response body or response header.
-pub(crate) struct Property<'a, T: Display> {
-    pub(crate) is_array: bool,
-    pub(crate) component_type: ComponentType<'a, T>,
+pub(crate) struct Property<'a> {
+    type_definition: Type<'a>,
```
juhaku, line 14: I guess this could also be a reference to the `Type` like `&'a Type<'a>` or `Cow<'a, Type<'a>` so there is no need to clone the Type for the `Property`

utoipa-gen/src/path/response.rs
```diff
@@ -285,12 +273,12 @@ impl Parse for Header<'_> {
 
 impl ToTokens for Header<'_> {
     fn to_tokens(&self, tokens: &mut TokenStream2) {
-        if let Some(ref header_type) = self.value_type {
+        if let Some(header_type) = &self.value_type {
             // header property with custom type
-            let header_type = Property::new(header_type.is_array, &header_type.ty);
+            let header_type_property = Property::new(header_type.clone());
```
juhaku, line 278: This clone could probably be removed if the type was reference or Cow. :) See above :point_up:

utoipa-gen/src/path/response.rs
```diff
@@ -111,12 +100,11 @@ impl ToTokens for Response<'_> {
             utoipa::openapi::ResponseBuilder::new().description(#description)
         });
 
-        if let Some(ref body_type) = self.response_type {
-            let body_ty = &body_type.ty;
+        if let Some(response_type) = &self.response_type {
+            let property = Property::new(response_type.clone());
```
juhaku, line 104: This clone could probably be removed if the type was reference or Cow. :) See above :point_up:

utoipa-gen/src/path/request_body.rs
```diff
@@ -122,11 +122,11 @@ impl ContentTypeResolver for RequestBodyAttr<'_> {}
 
 impl ToTokens for RequestBodyAttr<'_> {
     fn to_tokens(&self, tokens: &mut TokenStream2) {
-        if let Some(ref body_type) = self.content {
-            let property = Property::new(body_type.is_array, &body_type.ty);
+        if let Some(body_type) = &self.content {
+            let property = Property::new(body_type.clone());
```
juhaku, line 126: This clone could probably be removed if the type was reference or Cow. :) See above :point_up:

utoipa-gen/src/path/parameter.rs
```diff
@@ -185,8 +185,8 @@ impl ToTokens for Parameter<'_> {
                 }
             }
 
-            if let Some(ref parameter_type) = parameter.parameter_type {
-                let property = Property::new(parameter_type.is_array, &parameter_type.ty);
+            if let Some(parameter_type) = &parameter.parameter_type {
+                let property = Property::new(parameter_type.clone());
```
juhaku, line 189: This clone could probably be removed if the type was reference or Cow. :) See above :point_down:

juhaku (commented):
Pretty awesome :+1: Didn't see test for simple request body inlined but should work as well though. Maybe one could be added. 

Also this could be rebased with the latest changes from the master. :)

utoipa-gen/src/path/property.rs
kellpossible: I think I ran into some weird compile errors when I tried to do that last time, but I'll try again now that this branch is feature complete!

kellpossible: I think maybe we can make it `&'a Type<'a>` because we need to produce some owning copies somewhere.

juhaku: Yeah it will do :+1:

kellpossible: Okay nice! We can resolve the remaining comments too for now then?

kellpossible:
> Didn't see test for simple request body inlined but should work as well though

Sorry I must have missed that one, I'll have a look in the docs for an example of simple request body. What do they look like?

juhaku:
> > Didn't see test for simple request body inlined but should work as well though
> 
> Sorry I must have missed that one, I'll have a look in the docs for an example of simple request body. What do they look like?

It's according this format `#[utoipa::path(... request_body = BodyEntity, ...)]`

utoipa-gen/src/lib.rs
```diff
@@ -534,17 +537,17 @@ pub fn derive_component(input: TokenStream) -> TokenStream {
...
+/// ```
+///
+/// Note that `MyParameters` can also be used in combination with the [tuples
+/// representation](#tuples) or other structs. **For example:**
+///
+/// ```text
+/// params(
+///     MyParameters1,
+///     MyParameters2,
+///     ("id" = String, path, deprecated, description = "Pet database id"),
```
juhaku, line 550: I guess there are still few places in docs where `path` should be represented as `Path` probably good idea to go through them. I can also check them as well and fix what I find.  
```suggestion
///     ("id" = String, Path, deprecated, description = "Pet database id"),
```

utoipa-gen/src/schema/component.rs
```diff
@@ -796,16 +796,21 @@ where
...
+                        // When users wishes to hinder the actual type with Any type render a generic `object`
+                        if component_part.is_any() {
+                            tokens.extend(quote! { utoipa::openapi::ObjectBuilder::new() })
                         } else {
-                            let name = &*self.component_part.ident.to_string();
-                            tokens.extend(quote! {
-                                utoipa::openapi::Ref::from_component_name(#name)
-                            })
+                            if is_inline {
+                                let component_name_ident = &self.component_part.ident;
```
juhaku, line 804: remove the `&self` so it will work with the `override type` as well which is defined beginning of the function :)
```suggestion
                                let component_name_ident = component_part.ident;
```

utoipa-gen/src/path/property.rs
juhaku: Yeah

PR #167, opened 2022-06-13
fix: escape raw identifier

abdulrahman1s:
#### Example:
```rs
#[derive(Serialize, Deserialize, Component)]
struct Test {
  r#type: String
}
```
#### The result without this PR:
```json
{"type":"object","properties":{"type":{"r#type":"string"}}}
```


juhaku:
On the second thought this could be addressed in `utoipa-gen` crate as it is the derive crate that holds the business logic for the types. the `utoipa` crate from src folder contains only types and should not contain business logic for type that alter the types. Users should have the ability and be in full control how the they form the types and what names they use as keywords.

Issue #162, opened 2022-06-11
AnyValue {} for third party types


RemiKalbe:
Hello again,

I need to use some types that comes from a third party library but obviously they don't implement `#[derive(Component)]`. I looked at the doc and tried to do something like this 
```rust
#[derive(Debug, Deserialize, Component)]
pub struct SampleStruct {
    ...
    #[component(value_type = Object)]
    pub rtp_capabilities: RtpCapabilities,
    ...
}
```
but sadly it stills tries to create a ref to something that doesn't exist
```json
"rtp_capabilities": {
  "$ref": "#/components/schemas/RtpCapabilities"
},
```
what I would like instead is a way to specify that I want any kind of Object in `#[component]`. Obviously it would be better to actually have the correct type, but this would at least create a correct OpenAPI file that I can use.
Or maybe I'm missing something?

juhaku:
Yeah, this is not supported by desing actually. There is no way currently define `object` type and reason for this is that it is not helpful for users to see when they browse the OpenAPI doc or Swagger UI that the endpoint will return `object` which could contain anything. This might be a bit furstrating but for API consumers is a lot more helpful to see the actual types instead of any. If the OpenAPI spec was to be written manually it still has the need to define all the fields of the object.

But a way to get around this is to define type as a `Map` e.g. `HashMap<String, String>` since maps are by OpenAPI spec rendered as `object` types.

Support for this though could be added, but defining just an object is cheating and re-creating a type which has a lot of fields feels annoying.

RemiKalbe:
Yeah I know it's cheating, I'm only using the generated OpenAPI spec to generate a rust client to use internally so it's not really important to me if I have some objects that aren't defined. I suggested this as a simpler solution for the time being; correct serialization of third-party struct would be better of course.

juhaku:
Sure, such a case overriding the actual type as `object` could be reasonable. This is something that need to be addressed in the desing itself. But should not be a big addition though. I'll add note of this to the projects kanban board.

juhaku:
#168 duplicate touches the same matter.

PR #183, opened 2022-06-20
Add enhanced value_type for Components

### LINKED ISSUE #180: Component derive override field reference name
It would be nice to be able to override the name chosen for a reference on a given field, because the import name used for the component provided to the parent OpenAPI might be different.

Perhaps something like this to rename the reference from `Bar` to `NewBar`:

```rust
#[derive(Component)]
struct Foo {
    #[component(rename_ref = (from = Bar, to = NewBar))
    field: Bar
}
```

The `from` and `to` syntax is to provide compatibility with types that may have multiple different reference types.

juhaku:
Fixes #180 #162 

Add enhanced `value_type` for Components. Now `value_type` can be used
to alter a reference of a field of the component or `Any` type can be
used to define field type as a generic `object`.

### DIFF
--- a/tests/component_derive_test.rs
+++ b/tests/component_derive_test.rs
@@ -1092,6 +1092,57 @@ fn derive_unnamed_struct_component_type_override_with_format() {
     }
 }
 
+#[test]
+fn derive_struct_override_type_with_any_type() {
+    let value = api_doc! {
+        struct Value {
+            #[component(value_type = Any)]
+            field: String,
+        }
+    };
+
+    assert_json_eq!(
+        value,
+        json!({
+            "type": "object",
+            "properties": {
+                "field": {
+                    "type": "object"
+                }
+            },
+            "required": ["field"]
+        })
+    )
+}
+
+#[test]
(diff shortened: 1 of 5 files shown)


kellpossible:
@juhaku what happens if a user has a type called `Any`? I wonder if `Value` might be better name, because if we're working with `json` then it's analogous to `serde_json::Value`.

juhaku:
> @juhaku what happens if a user has a type called `Any`? I wonder if `Value` might be better name, because if we're working with `json` then it's analogous to `serde_json::Value`? Or we could even have it explicitly `serde_json::Value`?

Good question, that indeed would not work and would be a bummer. Yeah if users has a type called `Value` as well it would not work either. Calling it `serde_json::Value` would be good, but users could use this library without `serde_json`  although it would not make sense. So tying names to types of certain library is not necessary the most optimal naming. If we call it just a `Object` but also same as `Value` someone could create a type called `Object` and it could then match to the `ComponentType::Object` sort of. But how likely it is to someone create type called `Value`, `Object` or `Any`? and what would be the least likely? We could change it to any of those though.

PR #187, opened 2022-06-25
Namespacing same types

### LINKED ISSUE #186: Support namespacing for types which have the same name
One of the main reasons for using namespacing is to separate types without requiring type aliases:

```rust
components(
    crate::mod1::SomeType,
    crate::mod2::SomeType
)
```

Currently this isn't possible with the `openapi` macro, however I can supply these names manually without using the macro. But referencing these components with paths in other macros is also not possible.

As I understand it, this gets a little complicated because, the use case in #152 was to avoid needing to import types, but it still references the component using the final ident in the path `SomeType` which in this situation will cause a clash. 

Perhaps we need to introduce some kind of path name resolution syntax, such as `crate::mod::SomeType as mod::SomeType` so that it does not discard the complete path? 

```rust
components(
    crate::mod1::SomeType as mod1::SomeType,
    crate::mod2::SomeType as mod2::SomeType,
)
```

To the left of the `as` is the path used to get the `Component` implementation, and to the right of the `as` is the path/name to use for the component in the API specification, and to be used for referencing in other macros using the `value_type = mod1::SomeType` attribute.

Related to #152 and #164.

kellpossible:
Fix #186 

### DIFF
--- a/tests/component_derive_test.rs
+++ b/tests/component_derive_test.rs
@@ -1207,7 +1207,7 @@ fn derive_struct_override_type_with_a_reference() {
 
     let value = api_doc! {
         struct Value {
-            #[component(value_type = custom::NewBar)]
+            #[component(value_type = NewBar)]
             field: String,
         }
     };
@@ -1366,11 +1366,20 @@ fn derive_component_with_generic_types_having_path_expression() {
         }
     };
 
-    assert_value! {ty=>
-        "properties.args.items.items.type" = r#""string""#, "Args items items type"
-        "properties.args.items.type" = r#""array""#, "Args items type"
-        "properties.args.type" = r#""array""#, "Args type"
-    }
+    let args = ty.pointer("/properties/args").unwrap();
+
+    assert_json_eq!(
+        args,
+        json!({
+            "items": {
+                "items": {
+                    "type": "string"
+                },
+                "type": "array"
+            },
+            "type": "array"
+        })
+    );
 }
 
 #[test]
--- a/tests/component_derive_test.rs
+++ b/tests/component_derive_test.rs
@@ -1126,7 +1126,7 @@ fn derive_struct_component_field_type_path_override() {
         .unwrap()
         .as_str()
         .unwrap();
-    assert_eq!(component_ref, "#/components/schemas/path::to::Foo");
+    assert_eq!(component_ref, "#/components/schemas/path.to.Foo");
 }
 
 #[test]
--- a/tests/component_derive_test.rs
+++ b/tests/component_derive_test.rs
@@ -1096,6 +1096,24 @@ fn derive_struct_component_field_type_override() {
 }
 
 #[test]
+fn derive_struct_component_field_type_path_override() {
+    let post = api_doc! {
+        struct Post {
+            id: i32,
+            #[component(value_type = path::to::Foo)]
+            value: i64,
+        }
+    };
+
+    let component_ref: &str = post
+        .pointer("/properties/value/$ref")
+        .unwrap()
+        .as_str()
+        .unwrap();
+    assert_eq!(component_ref, "#/components/schemas/path::to::Foo");
+}
+
+#[test]
 fn derive_struct_component_field_type_override_with_format() {
     let post = api_doc! {
         struct Post {
--- a/tests/request_body_derive_test.rs
+++ b/tests/request_body_derive_test.rs
@@ -11,7 +11,7 @@ macro_rules! test_fn {
         mod $name {
             #[derive(utoipa::Component)]
             /// Some struct
-            struct Foo {
+            pub struct Foo {
                 /// Some name
                 name: String,
             }
@@ -94,6 +94,7 @@ fn derive_request_body_option_array_success() {
         "paths./foo.post.requestBody.description" = r###"null"###, "Request body description"
     }
 }
+
 test_fn! {
     module: derive_request_body_primitive_simple,
     body: = String
@@ -373,4 +374,30 @@ fn derive_request_body_complex_primitive_array_success() {
         "paths./foo.post.requestBody.required" = r###"true"###, "Request body required"
         "paths./foo.post.requestBody.description" = r###""Create new foo references""###, "Request body description"
     }
+}
+
+test_fn! {
+    module: derive_request_body_primitive_ref_path,
+    body: = path::to::Foo
+}
+
+#[test]
+fn derive_request_body_primitive_ref_path_success() {
+    #[derive(OpenApi, Default)]
+    #[openapi(
+        handlers(derive_request_body_primitive_ref_path::post_foo),
+        components(derive_request_body_primitive_ref_path::Foo as path::to::Foo)
+    )]
+    struct ApiDoc;
+
+    let doc = serde_json::to_value(&ApiDoc::openapi()).unwrap();
+    let schemas = doc.pointer("/components/schemas").unwrap();
+    assert!(schemas.get("path::to::Foo").is_some());
+
+    let component_ref: &str = doc
+        .pointer("/paths/~1foo/post/requestBody/content/application~1json/schema/$ref")
+        .unwrap()
+        .as_str()
+        .unwrap();
+    assert_eq!(component_ref, "#/components/schemas/path::to::Foo");
 }
--- a/tests/request_body_derive_test.rs
+++ b/tests/request_body_derive_test.rs
@@ -392,12 +392,12 @@ fn derive_request_body_primitive_ref_path_success() {
 
     let doc = serde_json::to_value(&ApiDoc::openapi()).unwrap();
     let schemas = doc.pointer("/components/schemas").unwrap();
-    assert!(schemas.get("path::to::Foo").is_some());
+    assert!(schemas.get("path.to.Foo").is_some());
 
     let component_ref: &str = doc
         .pointer("/paths/~1foo/post/requestBody/content/application~1json/schema/$ref")
         .unwrap()
         .as_str()
         .unwrap();
-    assert_eq!(component_ref, "#/components/schemas/path::to::Foo");
+    assert_eq!(component_ref, "#/components/schemas/path.to.Foo");
 }
--- a/utoipa-gen/src/lib.rs
+++ b/utoipa-gen/src/lib.rs
@@ -21,8 +21,7 @@ use schema::into_params::IntoParams;
 
 use proc_macro2::{Group, Ident, Punct, TokenStream as TokenStream2};
 use syn::{
-    bracketed,
-    parse::{Parse, ParseBuffer, ParseStream},
+    parse::{Parse, ParseStream},
     punctuated::Punctuated,
     spanned::Spanned,
     AngleBracketedGenericArguments, DeriveInput, ExprPath, GenericArgument, ItemFn, Lit, LitStr,
@@ -1111,13 +1110,6 @@ where
 /// This is typically used in component `value_type` when type of the field is overridden by the user.
 #[cfg_attr(feature = "debug", derive(Debug))]
 struct ValueType(TypePath);
-
-impl ValueType {
-    /// Get the `Ident` of last segment of the [`syn::ExprPath`].
-    fn get_ident(&self) -> Option<&Ident> {
-        self.0.path.segments.last().map(|segment| &segment.ident)
-    }
-}
 
 impl Parse for ValueType {
     fn parse(input: ParseStream) -> syn::Result<Self> {
--- a/utoipa-gen/src/lib.rs
+++ b/utoipa-gen/src/lib.rs
@@ -1332,9 +1332,8 @@ impl Parse for ArrayOrOptionType<'_> {
                     return Err(syn::Error::new(
                         unsupported_type.span(),
                         format!(
-                            "Unsupported type {} {:?}. {}",
+                            "Unsupported type {}. {}",
                             unsupported_type.to_token_stream().to_string(),
-                            unsupported_type,
                             EXPECTED_TYPE_MESSAGE
                         ),
                     ))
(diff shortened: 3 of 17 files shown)


kellpossible:
Right now there is a failing test for:

```rust
#[derive(Component)]
struct Bar {
     args: Vec<std::vec::Vec<std::string::String>>
}
```

I'm wondering whether it's reasonable to even resolve primitive types when they have a path?

kellpossible:
It seems like I may have also broken some of the optional features format resolution with this.

tests/component_derive_test.rs
```diff
@@ -1112,6 +1112,24 @@ fn derive_struct_component_field_type_override() {
...
+            value: i64,
+        }
+    };
+
+    let component_ref: &str = post
+        .pointer("/properties/value/$ref")
+        .unwrap()
+        .as_str()
+        .unwrap();
+    assert_eq!(component_ref, "#/components/schemas/path::to::Foo");
```
juhaku, line 1130: I guess we need to think this namespacing over again. This kind of a name `"#/components/schemas/path::to::Foo");` is not valid OpenAPI spec name and will cause errors when rendered. 

The name should be just single word.

tests/request_body_derive_test.rs
```diff
@@ -374,3 +375,29 @@ fn derive_request_body_complex_primitive_array_success() {
...
+    let doc = serde_json::to_value(&ApiDoc::openapi()).unwrap();
+    let schemas = doc.pointer("/components/schemas").unwrap();
+    assert!(schemas.get("path::to::Foo").is_some());
+
+    let component_ref: &str = doc
+        .pointer("/paths/~1foo/post/requestBody/content/application~1json/schema/$ref")
+        .unwrap()
+        .as_str()
+        .unwrap();
+    assert_eq!(component_ref, "#/components/schemas/path::to::Foo");
```
juhaku, line 402: Same here :point_up:

utoipa-gen/src/lib.rs
```diff
@@ -1245,45 +1238,111 @@ struct ArrayOrOptionType<'a>(Type<'a>);
...
+                        if angle_bracketed.args.len() != 1 {
+                            return Err(syn::Error::new(type_segment.span(), "Option must have only a single generic parameter specified. e.g. Option<String>"));
+                        }
+
+                        let argument: &GenericArgument = angle_bracketed.args.first().expect(
+                            "Expected there to be 1 angle bracketed argument for Option<...>",
+                        );
+
+                        let argument_path: &TypePath = match argument {
+                            GenericArgument::Type(syn::Type::Path(path)) => path,
```
juhaku, line 1277: We probably could get rid of this whole custom type parsing and use the `ComponentPart`. This would allow users to use the regular Rust types as arguments.

utoipa-gen/src/schema.rs
```diff
@@ -56,38 +63,15 @@ impl<'a> ComponentPart<'a> {
...
-            .find_map(|pair| match pair {
-                Pair::Punctuated(_, _) => None,
-                Pair::End(segment) => Some(segment),
-            })
-            .unwrap();
-
-        if segment.arguments.is_empty() {
-            op(&segment.ident, segment)
+    /// Creates a [`ComponentPath`] from a [`TypePath`].
+    /// `op` is a function
```
juhaku, line 67: I guess now this is unnecessary line :)
```suggestion
    /// `op` is a function
```

utoipa-gen/src/schema.rs
```diff
@@ -100,7 +84,12 @@ impl<'a> ComponentPart<'a> {
             );
         };
 
-        let mut generic_component_type = ComponentPart::convert(&segment.ident, segment);
+        let path = TypePath {
+            qself: None,
+            path: syn::Path::from(segment.clone()),
+        };
+
+        let mut generic_component_type = ComponentPart::convert(Cow::Owned(path));
```
juhaku, line 92: Hmm.. interesting approach :thinking:

utoipa-gen/src/schema/into_params.rs
```diff
@@ -315,30 +315,30 @@ struct ParamType<'a>(&'a ComponentPart<'a>);
...
+                    let format = ComponentFormat(&*part.path);
                     if format.is_known_format() {
                         tokens.extend(quote! {
                             .format(Some(#format))
                         })
                     }
                 }
                 ValueType::Object => {
-                    let name = ty.ident.to_string();
+                    let name = format_path_ref(&part.path.to_token_stream().to_string());
```
juhaku, line 341: Rust guidelines usually tells to call free standing functions with the path. This gives user more context where the function is located and user will not mix it with local functions. :)
```suggestion
                    let name = component::format_path_ref(&part.path.to_token_stream().to_string());
```

juhaku (changes requested):
I refactored the `value_type` functionality for `Component` and added it to `IntoParams` as well. This PR needs rebase from latest master. 

There are couple of things need a bit of rethinking. One being the namespacing. At least we cannot have it namespaced with the Rust path which will result invalid OpenAPI spec. Maybe better way is to allow users to use `as MyAnotherType` but this might not be feasible because the question relies who to reference such type which uses `as MyType` syntax.

Another thing is that I have been planning on refactoring the `component()` function to return a name of the component as well along side with the component. This would allow users to use  serde's `rename` on container level. But with this as well the hot question is how to reference such component which has been renamed. 

Lastly is it too cumbersome to support sort of namespacing? It would be great to have some solution to this but after all if one was to write a OpenAPI spec manually he/she would need to write distinct type names for each type. So there would not be a clash in such scenario. Point being use uniques type names for `Component`s as well. :joy:

kellpossible:
> At least we cannot have it namespaced with the Rust path which will result invalid OpenAPI spec. 

Is this true? It seems to work fine in swagger, and I can't see anything in the spec which says that component references can't contain `::` in them.

> Lastly is it too cumbersome to support sort of namespacing? It would be great to have some solution to this but after all if one was to write a OpenAPI spec manually he/she would need to write distinct type names for each type. So there would not be a clash in such scenario. Point being use uniques type names for Components as well.

Let's say you have module `alpha` and inside you have `Query`, and inside module `beta` you have `Query`. You could rename them to `AlphaQuery` and `BetaQuery` but the Rust community errs against this, https://rust-lang.github.io/rust-clippy/master/#module_name_repetitions . I could rename them with a `use alpha::Query as AlphaQuery` but this makes mapping the objects back to the code more cumbersome.

tests/component_derive_test.rs
kellpossible: Sorry, it works fine in Swagger, and RapiPDF, and I can't find anything in the spec yet which says that the name cannot contain a `::`. Where is the reference for this?

kellpossible: From what I could find out openapi 3.0 uses the json pointer specification for these references: https://datatracker.ietf.org/doc/html/draft-ietf-appsawg-json-pointer-04 there is also no mention in here about colons being an illegal character.

juhaku:
>> At least we cannot have it namespaced with the Rust path which will result invalid OpenAPI spec.

> Is this true? It seems to work fine in swagger, and I can't see anything in the spec which says that component references can't contain :: in them. 

It actually works but not without caveats. If you try out the spec with https://editor.swagger.io/ and try to use `::` referenced names here it will work but same time it will complain you about invalid names. 
image
image
While this may not be a deal breaker for all but certainly most of people want to have spec validated and linted correctly.

If we want to have namespaces implemented there. I guess we need to do manual mangling of the type name according to the rules of Swagger. 

> Let's say you have module `alpha` and inside you have `Query`, and inside module `beta` you have `Query`. You could rename them to `AlphaQuery` and `BetaQuery` but the Rust community errs against this, https://rust-lang.github.io/rust-clippy/master/#module_name_repetitions . You could rename them with a use `alpha::Query as AlphaQuery` but this makes mental mapping the objects back to the code more cumbersome.

Yeah I agree it's not optimal solution not necessary fan of implementing such thing too though :smile:

tests/component_derive_test.rs
juhaku: Sorry forgot to add the details previously. Added them now down there :point_down:

kellpossible:
@juhaku 

> While this may not be a deal breaker for all but certainly most of people want to have spec validated and linted correctly.
> 
> If we want to have namespaces implemented there. I guess we need to do manual mangling of the type name according to the rules of Swagger

That sounds reasonable. How about this proposal, we replace `::` with `.`?

juhaku:
> @juhaku
> 
> > While this may not be a deal breaker for all but certainly most of people want to have spec validated and linted correctly.
> > If we want to have namespaces implemented there. I guess we need to do manual mangling of the type name according to the rules of Swagger
> 
> That sounds reasonable. How about this proposal, we replace `::` with `.`? And if people don't like it, nobody is forcing them to reference components using paths, but at least there is the option, because the alternative effectively requires a bunch of renaming or awkward imports in a pre-existing codebase.

I reckon we could go with the dot `.` instead of the colon`:`. Though it kind of sticks out from Rust syntax but I guess that is best we can go with so far. Underscore `_` would look too out of the place I wonder? Yeah we can offer a way to use references but its not necessary to them to use it.

 Also another thing I am wondering could we use sort of special syntax like `components(path(foo::bar::MyType))` if we wanted to have the full path seen in the OpenAPI doc? This might be also too much and just add to the complexity which would be unnecessary for the users point of view though?

kellpossible:
> I reckon we could go with the dot `.` instead of the colon`:`. Though it kind of sticks out from Rust syntax but I guess that is best we can go with so far. Underscore `_` would look too out of the place I wonder? Yeah we can offer a way to use references but its not necessary to them to use it.

Yeah I think `.` is definitely better than an underscore, it's more commonly used in programming to delineate context.

> Also another thing I am wondering could we use sort of special syntax like `components(path(foo::bar::MyType))` if we wanted to have the full path seen in the OpenAPI doc? This might be also too much and just add to the complexity which would be unnecessary for the users point of view though?

I think I agree that this is unnecessary at this stage if we have the `.` it should be good enough and simple to implement.

juhaku:
Yeah good

juhaku:
We probably need to strip the generics out of the path. The error in build is because the path has generic params 
` "#/components/schemas/Foo < 'bar >"`

Failing test: derive_component_with_complex_enum_lifetimes

kellpossible:
> We probably need to strip the generics out of the path. The error in build is because the path has generic params ` "#/components/schemas/Foo < 'bar >"`
> 
> Failing test: derive_component_with_complex_enum_lifetimes

Thanks! Yes I didn't have time to fix that one yet, I'll fix it next week and the other comments, and hopefully ready to merge soon so it doesn't require another large merge conflict resolution.

juhaku:
> hopefully ready to merge soon so it doesn't require another large merge conflict resolution.

Yeah sorry about that previously :bow: . Hopefully not, I'm working on the extension part which is getting some refinment there and the impact should not be too big.

utoipa-gen/src/lib.rs
kellpossible: Unfortunately it fails to parse things like `[Foo]` and `Option<[Foo]>`

kellpossible: Perhaps we could do a breaking change, and make it so people need to use `Vec<Foo>` and `Option<Vec<Foo>>` instead?

kellpossible:
@juhaku I can't figure out why the `utoipa-gen` tests are failing on CI, it's not happening locally for me

juhaku:
There seems to be broken rocket related tests. You can run all tests with command `./tests/test.sh utoipa` and see what goes wrong. Seems like rocket integration breaks for some reason. :thinking: 
```
Test: resolve_get_with_multiple_args
Test: resolve_get_with_optinal_query_args
Error: Query items parameter type: [0].schema.items.type expected to be: "string" but was: null', tests/path_derive_rocket.rs:65:5
```

kellpossible:
@juhaku I had a look into it today, it seems that the parsing for rocket arguments types (and perhaps actix too?) is pretty limited, ideally we should share code with the component derive macro?

kellpossible:
I'm thinking maybe to quickly fix this bug with a work around, and lodge a new issue to propose a refactor in this regard

kellpossible:
I'm also thinking maybe we should merge this with the `[Type]` and `Option<[Type]>` syntax parsing route, we've got almost the same logic in about 3 different places.

utoipa-gen/src/schema.rs
kellpossible: Do you think there is a better way to do it?

juhaku:
> @juhaku I had a look into it today, it seems that the parsing for rocket arguments types (and perhaps actix too?) is pretty limited, ideally we should share code with the component derive macro?

Yeah those use cases are quite specific. And there is surely room to improve and refactor the implementation. The axum integration will party address this issue. But will not fully "fix" it. My plan is to address the steps in #217 regarding this matter.

> I'm thinking maybe to quickly fix this bug with a work around, and lodge a new issue to propose a refactor in this regard

Yeah that is okay for now. 

> I'm also thinking maybe we should merge this with the [Type] and Option<[Type]> syntax parsing route, we've got almost the same logic in about 3 different places.

I also commented there about this issue.

juhaku (commented):
Super. :rocket:  Few comments only and soon this can be merged :)

juhaku (approved):

kellpossible:
@juhaku thanks for reviewing this PR, it was a bit of a mammoth in the end!

radix:
Is there some documentation for how to deal with name conflicts? I have noticed that utoipa seems to just silently drop types if they have the same name so they don't end up in my openapi output. I would much prefer an error in this case, with a pointer to docs on how to avoid the name conflict. I've tried to find some information in the associated ticket and this PR about how to do this, but it's not clear to me.

PR #188, opened 2022-06-25
Into params inline

kellpossible:
Add inline option to params(...) struct field attribute for `#[derive(IntoParams)]`

This allows the schema for a type of a struct field to be inlined with
the `IntoParams` derive.

### DIFF
--- a/examples/todo-warp/src/main.rs
+++ b/examples/todo-warp/src/main.rs
@@ -112,12 +112,22 @@ mod todo {
         value: String,
     }
 
+    #[derive(Debug, Deserialize, Component)]
+    #[serde(rename_all = "snake_case")]
+    pub enum Order {
+        AscendingId,
+        DescendingId,
+    }
+
     #[derive(Debug, Deserialize, IntoParams)]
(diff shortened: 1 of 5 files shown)


juhaku (approved):
Pretty awesome :1st_place_medal:  Is the inline supported also for `Vec<..>` types here?

kellpossible:
@juhaku good question! If there's not already a test for that I can add one to check, and if it's not I could add it

kellpossible:
Question, how would deserializing Vec be supported by a query parameter in a standard manner?

juhaku:
> @juhaku  how would deserializing Vec be supported by a query parameter in a standard manner? 

Vec in query parameters is typically rendered as comma separated list of items. 

> edit: nevermind, there was a bug with Vec, but I added a test for it and fixed the problem

I merged that code to the main.

kellpossible:
> I merged that code to the main.

You already merged this branch?

juhaku:
> > I merged that code to the main.
> 
> You already merged this branch?

The other PR which fixes the Vec support for inline.

Issue #192, opened 2022-06-27
Encoding for Content


kellpossible:
The specification allows for `encoding` to be specified for content https://swagger.io/specification/#media-type-object but this feature is currently missing from [`ContentBuilder`](https://docs.rs/utoipa/latest/utoipa/openapi/content/struct.ContentBuilder.html)

PR #193, opened 2022-06-27
Implement Encoding for Content. 

kellpossible:
Implements #192


juhaku (approved):
Awesome :+1:

src/openapi/encoding.rs
```diff
@@ -0,0 +1,100 @@
...
+from!(Encoding EncodingBuilder content_type, headers, style, explode, allow_reserved);
+
+/// Builder for [`Encoding`] with chainable configuration methods to create a new [`Encoding`].
+#[derive(Default)]
+pub struct EncodingBuilder {
+    content_type: String,
+    headers: HashMap<String, Header>,
+    style: Option<ParameterStyle>,
+    explode: Option<bool>,
+    allow_reserved: Option<bool>,
```
juhaku, line 67: Add some serde optionality so not all will be serialized.
```suggestion
    #[serde(skip_serializing_if = "Optional::is_none")]
    allow_reserved: Option<bool>,
```

src/openapi/encoding.rs
```diff
@@ -0,0 +1,100 @@
...
+
+from!(Encoding EncodingBuilder content_type, headers, style, explode, allow_reserved);
+
+/// Builder for [`Encoding`] with chainable configuration methods to create a new [`Encoding`].
+#[derive(Default)]
+pub struct EncodingBuilder {
+    content_type: String,
+    headers: HashMap<String, Header>,
+    style: Option<ParameterStyle>,
+    explode: Option<bool>,
```
juhaku, line 66:
```suggestion
    #[serde(skip_serializing_if = "Optional::is_none")]
    explode: Option<bool>,
```

src/openapi/encoding.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+from!(Encoding EncodingBuilder content_type, headers, style, explode, allow_reserved);
+
+/// Builder for [`Encoding`] with chainable configuration methods to create a new [`Encoding`].
+#[derive(Default)]
+pub struct EncodingBuilder {
+    content_type: String,
+    headers: HashMap<String, Header>,
+    style: Option<ParameterStyle>,
```
juhaku, line 65:
```suggestion
    #[serde(skip_serializing_if = "Optional::is_none")]
    style: Option<ParameterStyle>,
```

src/openapi/encoding.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+from!(Encoding EncodingBuilder content_type, headers, style, explode, allow_reserved);
+
+/// Builder for [`Encoding`] with chainable configuration methods to create a new [`Encoding`].
+#[derive(Default)]
+pub struct EncodingBuilder {
+    content_type: String,
+    headers: HashMap<String, Header>,
```
juhaku, line 64:
```suggestion
    #[serde(skip_serializing_if = "HashMap::is_empty")]
    headers: HashMap<String, Header>,
```

src/openapi/encoding.rs
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+}
+
+from!(Encoding EncodingBuilder content_type, headers, style, explode, allow_reserved);
+
+/// Builder for [`Encoding`] with chainable configuration methods to create a new [`Encoding`].
+#[derive(Default)]
+pub struct EncodingBuilder {
+    content_type: String,
```
juhaku, line 63: This could be optional as it is not a required field by the spec. But could also be a `String` as it does not allocate while being empty. Either I would add the suggestion below so it will not serialize it is not present. Or I would make it `Option` similar to other fields with the `serde` attribute.
```suggestion
    #[serde(skip_serializing_if = "String::is_empty")]
    content_type: String,
```

juhaku (changes requested):
Taking closer look a couple of things could be added. :) Previously I have written the doc comments basing on the once existing in the Swagger docs. Yeah I also am not sure how big of a deal it would be. There now are already in utoipa-swagger-ui crate some comments basically "copy pasted". I would go according to the once own conscience.

juhaku:
> 
>     * [ ]  Does content_type need to be optional?

Propbably not, but could also be since it is not a required field.

juhaku (approved):
Super, though we could use the `builder!` macro since it does not have example which required conditional field parameters like some of the components does. But I can change that though.

PR #215, opened 2022-07-11
Response references, IntoResponseComponent and IntoResponses

### LINKED ISSUE #201: IntoResponses trait
It would be nice to have an `IntoResponses` trait that works in a similar way to the `IntoParams` trait. 

```rust
trait IntoResponses {
    fn responses() -> Vec<utoipa::openapi::response::Response>
}
```

Personally I would like to create a custom derive of `IntoResponses` for enums that represent different response types on a platform that I'm documenting, and be able to just plug it into the `path` attribute like:

```rust
#[derive(my_macros::IntoResponses)]
enum Response {
    Ok,
    NotFound,
    Forbidden,
}

#[path(
    ...
    responses(Response)
)]
````

This would also work fine for structs which can return a single response:

```rust
#[derive(my_macros::IntoResponses)]
struct Ok;

#[derive(my_macros::IntoResponses)]
struct NotFound;

#[derive(my_macros::IntoResponses)]
struct Forbidden;

#[path(
    ...
    responses(
        Ok,
        NotFound,
        Forbidden,
    )
)]
````

### PR DESCRIPTION
**Pull request by kellpossible**
Close #201

TODO:
- [x] ~Implement the `responses( (code = "200", response = MyResponse) )` as per https://github.com/juhaku/utoipa/issues/201#issuecomment-1180771424 if we decide we want it.~ this can be done in a separate PR

### CODE REVIEW

File: src/lib.rs [hunk 1]
```diff
@@ -540,3 +543,62 @@ pub trait IntoParams {
...
+/// impl ResponseComponent for MyResponse {
+///     fn response() -> (String, Response) {
+///         (
+///             "MyResponse".to_string(),
+///             ResponseBuilder::new().description("My Response").build(),
+///         )
+///     }
+/// }
+/// ```
+pub trait ResponseComponent {
```
juhaku, line 601: Just wondering should this actually just be `Response` instead of a `ResponseComponent`? Component response seems quite long name feels like little bit stuttering? Although if the name is just a `Response` it easily will conflict with other crates which offer types with same name. 

Also wondering should the `Component` derive trait be renamed to `Schema` but it might be too big breaking change for people already using the library. Basically they would have to change it to every type they previously had `Component`

utoipa-gen/src/openapi.rs
```diff
@@ -269,45 +288,48 @@ impl ToTokens for OpenApi {
...
-                let assert_ty_generics = if component.has_lifetime_generics() {
-                    Some(quote! {<'static>})
-                } else {
-                    Some(ty_generics.to_token_stream())
-                };
-                let assert_component = format_ident!("_AssertComponent{}", component_name);
-                tokens.extend(quote_spanned! {span=>
-                    struct #assert_component where #path #assert_ty_generics: utoipa::Component;
-                });
-
```
juhaku, line 295: Seems like the assert component is being removed here? I had previously added these assert components for better error reports when someone is missing the `Component` derive from the item.

utoipa-gen/src/path/response.rs
```diff
@@ -138,26 +158,33 @@ impl ToTokens for Response<'_> {
...
 
                 acc
-            });
+            },
+        ));
 
-            tokens.extend(quote! {
-                utoipa::openapi::Responses::from_iter([#responses])
-            });
-        }
```
juhaku, line 160: I guess we cannot use similar approach here since there might be `into_responses()` call for responses and same time a `Response` object. The nice thing with this `from_iter()` is that it would make only one allocation of exact size. 

When looping over items without creating a map or a vec with extact size with `::with_capacity()` would make the result collection to reallocate when items where added.

utoipa-gen/src/schema/component.rs
```diff
@@ -795,13 +795,8 @@ where
...
                             tokens.extend(quote! { utoipa::openapi::ObjectBuilder::new() })
                         } else if is_inline {
-                            let assert_component = format_ident!("_Assert{}", name);
                             tokens.extend(quote_spanned! {component_ident.span()=>
-                                {
-                                    struct #assert_component where #component_ident : utoipa::Component;
-                                    
-                                    <#component_ident as utoipa::Component>::component()
-                                }
+                                <#component_ident as utoipa::Component>::component()
```
juhaku, line 799: Same here as above, the assert stuct was there to provide better errors for users which is also a pattern demonstrated in syn crate's docs.

juhaku (commented):
Awesome work. :trophy: just a few comments to clarify and this could also be merged, supposely. Also there is a need to take a rebase from the upstream. :)

src/lib.rs
kellpossible: I was also tempted by this change, it would be good to make for the sake of clarity and consistency, but perhaps too big of a break? At least with renames it's usually not too difficult to fix with a grep + sed style find replace.

kellpossible: > Just wondering should this actually just be Response instead of a ResponseComponent? Component response seems quite long name feels like little bit stuttering? Although if the name is just a Response it easily will conflict with other crates which offer types with same name.

Sure I would be happy to rename it. If it's a conflict people can import it with a rename or with a module scope :slightly_smiling_face:

utoipa-gen/src/openapi.rs
kellpossible: Sorry I should have mentioned this in the description for the PR. I found that the `<#component as utoipa::Component>::component()` in a spanned quote, the message is pretty clear, there doesn't appear to be any benefit of having the assert, or maybe I'm missing something?

 If I remove the `Component` derive for `Todo` in the `todo-warp` example project:

With the assert:

```text
error[E0277]: the trait bound `Todo: utoipa::Component` is not satisfied
  --> src/main.rs:24:20
   |
24 |         components(todo::Todo),
   |                    ^^^^^^^^^^ the trait `utoipa::Component` is not implemented for `Todo`
   |
   = help: see issue #48214

For more information about this error, try `rustc --explain E0277`.
```

Without assert, in this PR:

```text
error[E0277]: the trait bound `Todo: utoipa::Component` is not satisfied
  --> src/main.rs:24:26
   |
24 |         components(todo::Todo),
   |                          ^^^^ the trait `utoipa::Component` is not implemented for `Todo`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `todo-warp` due to previous error
```

utoipa-gen/src/path/response.rs
kellpossible: Thanks I didn't think of that. [It looks like there is no `::with_capacity()` for `BTreeMap`](https://users.rust-lang.org/t/btreemap-with-capacity/39770/3) which is [used by us to store responses](https://github.com/juhaku/utoipa/pull/215/files#diff-eea3ccce264a9afded89ef39496101cb1a5bf8c1fedc4cd4ddf9221e4744d922R27).

It also doesn't make use of the size hint/length anywhere with its `FromIterator` that I can find:
+ [BTreeMap FromIterator impl](https://stdrs.dev/nightly/x86_64-unknown-linux-gnu/src/alloc/collections/btree/map.rs.html#2089-2099)
+ [BTreeMap internal bulk_push](https://stdrs.dev/nightly/x86_64-unknown-linux-gnu/src/alloc/collections/btree/append.rs.html#38-90) (looks like it doesn't use the size hint anywhere)

We could switch to using a Vec<(String, Response)> internally for `Responses` but we would lose the deduplication guarantees. What do you think?

utoipa-gen/src/schema/component.rs
kellpossible: https://github.com/juhaku/utoipa/pull/215#discussion_r921544433

utoipa-gen/src/path/response.rs
kellpossible: Anyhow I can add a `From<IntoResponses>` impl for `Responses` for people who want it.

kellpossible: The other thing about this is that currently this macro supports interleaving both `IntoResponses` and regular tuple defined responses (just like `IntoParams` for `params()`), which is why `responses.extend()` is used in the impl for  `ResponsesBuilder::responses_from_into_responses()`:

```rust
responses(
    Response,
    (status = 404, description = "some extra response"),
)
```

kellpossible: Another thing I'm wondering is that our status codes are currently strings, but should we perhaps enforce them as `u16`? Or is the http status code fair game for any string in practice? It doesn't seem like the openapi spec itself enforces this.

utoipa-gen/src/openapi.rs
juhaku: Actually thats great. :slightly_smiling_face: Then we can remove these assert occurences there completely. There are few other's as well too.

utoipa-gen/src/schema/component.rs
juhaku: We can remove these assert structs then. Less code anyway better :smile:

utoipa-gen/src/path/response.rs
juhaku: > We could switch to using a Vec<(String, Response)> internally for Responses but we would lose the deduplication guarantees. Or we could switch to using a HashMap<String, Response> but we would lose the ordering. These types make use of the size hint in their FromIterator to avoid allocations as I understand it. What do you think?

I think we can keep it as is with the `BTreeMap`. I don't think the benefits would out weight the ordering support and be that significant. So I guess we can keep it as is? Any thoughts?

juhaku: > We would have to first collect all of these together in a single data structure, and then use FromIterator for Responses. Even then I'm not sure if we could compose it down to a single allocation.

Yeah might just be a waste of effort to try to reduce allocations in this case

kellpossible: > I think we can keep it as is with the BTreeMap. I don't think the benefits would out weight the ordering support and be that significant. So I guess we can keep it as is? Any thoughts?

I think it sounds reasonable to keep as is, we can find a way to optimise it later if we need.

> We could internally use it as u16 might actually not be a bad idea. I originally thought to use a number type for this. But leaned back to String because within OpenAPI document the responses must string=response map.

Nice I'll make a new issue for this idea :slightly_smiling_face:

src/lib.rs
kellpossible: Let's do it in a subsequent PR, seeing as it it is a rather large change!

juhaku: Good

utoipa-gen/src/path/response.rs
juhaku: Sounds reasonable.

src/lib.rs
kellpossible: > Sure I would be happy to rename it. If it's a conflict people can import it with a rename or with a module scope

The problem here is that it's always a conflict because to implement this trait you must always use `openapi::Response`, the same problem happens with the `Component` trait.  I'm thinking it's still good to keep this as `ResponseComponent`, and rename the `Component` trait to `SchemaComponent` in the future.

utoipa-gen/src/path/response.rs
kellpossible: > Nice I'll make a new issue for this idea slightly_smiling_face, and perhaps implement it in this PR because otherwise we'll have to break the interface again.

I think I'll abandon this idea, it requires a custom serializer/deserialzer implementation, can implement it in a future PR

src/lib.rs
juhaku: I'd still prefer them to be one word if possible. Currently the `Component` trait conflicts with the `Component` type. But it is hardly visible to the users. Not sure whether the the `Component` struct should be renamed to `Schema` derivent after renaming `Component` trait. Using `ResponseComponent`or `SchemaComponent`seems little bit of stuttering to me and superfluous to repeat the `Component` word everywhere as well it is unnecessary long to type. :thinking: Is it that it will conflict with the types from the warp or some other?

juhaku: > The problem here is that it's always a conflict because to implement this trait you must always use openapi::Response, the same problem happens with the Component trait.

Typically this won't show up to the users and is not an issue for end users. :thinking:  Since mostly noone will implement manually those traits. Or we could have the traits named as `ResponseComponent` and `SchemaComponent` but the derive name would be `Response` and `Schema`? Was this what you were after? 

> and rename the Component trait to SchemaComponent in the future.

Yeah lets rename that in separete PR so this will not grow overly large.

Also I was considering whether the openapi was defined like this
```rust
#[derive(OpenApi)]
#[openapi(
    schemas(...),
    responses(...),
)]
```
Instead of this. It would be more straight forward but this follows better the OpenAPI spec. 
```rust
#[derive(OpenApi)]
#[openapi(
    components(
        schemas(...),
        responses(...),
    )
)]
```

Also the `handlers` in `#[openapi]` should be renamed to `paths` to follow the spec as well. But that can be done in future PR:s as well.

kellpossible: >  Or we could have the traits named as ResponseComponent and SchemaComponent but the derive name would be Response and Schema?

Sure I can support that :slightly_smiling_face: Having some distinguishing feature in the name other than one belonging to the `openapi` module and the other not, because that's not enough for me personally to guess at a glance that they represent different concepts. 

> Instead of this. It would be more straight forward but this follows better the OpenAPI spec.

That looks good, do you think it might be good to represent this structure in the underlying data types using something like a `ComponentsBuilder`?

> Also the handlers in #[openapi] should be renamed to paths to follow the spec as well. But that can be done in future PR:s as well.

Great idea!

kellpossible: > Or we could have the traits named as ResponseComponent and SchemaComponent but the derive name would be Response and Schema?

Actually on second thoughts, shouldn't the derive name match the trait? That seems to be standard practice in the Rust community.

kellpossible: In a sense, the trait `Component` isn't actually a component, it's a component generator, it generates openapi components. Perhaps `ComponentGenerator` or something with a similar meaning would be a better name for it?

juhaku: > Actually on second thoughts, shouldn't the derive name match the trait? That seems to be standard practice in the Rust community.

Yeah I was wondering the same actually so better name them same :) 

> In a sense, the traits Schema (as proposed) isn't actually a , it's a schema generator, it generates openapi schema on demand. Perhaps SchemaGenerator or something with a similar meaning would be a better name for it? Same deal for Response trait.

Maybe `ToSchema` and `ToResponse` this would then follow same convention as syn crate's `ToTokens`?

juhaku: > That looks good, do you think it might be good to represent this structure in the underlying data types using something like a ComponentsBuilder?

If something then the `Comonent` references should be probably renamed to `Schema`s. And similarly the function `components_from_iter` and `component` should be renamed to `schemas_from_iter` and to `schema`. Then it would follow the spec more closely.

juhaku: > Perhaps `SchemaGenerator` or something with a similar meaning

Yeah the `SchemaGenerator` indicate that it creates a generator object. Perhaps `GenerateSchema` but again it would be unnecessary long or `GenSchema`? Like serde's `Serialize` or `Deserialize` what would be if possible one word or as short as possible descriptive action that it would perform. Rust prefers naming traits as one verb but also `or` and `er` endings are somewhat common too. I guess `ToSchema` and `ToResponse` would be not too far fetched and sound quite  reasonable?

kellpossible: > components_from_iter and component should be renamed to schemas_from_iter and to schema. Then it would follow the spec more closely.

I've made that change in the PR for `ComponentsBuilder`, because it stores the components in a variable called `schemas` already, so I made it match.

kellpossible: > Yeah the `SchemaGenerator` indicate that it creates a generator object. Perhaps `GenerateSchema` but again it would be unnecessary long or `GenSchema`? Like serde's `Serialize` or `Deserialize` what would be if possible one word or as short as possible descriptive action that it would perform. Rust prefers naming traits as one verb but also `or` and `er` endings are somewhat common too. I guess `ToSchema` and `ToResponse` would be not too far fetched and sound quite reasonable?

I think using the full word is usually better than shortcuts because it makes the code more accessible to those who don't speak English natively (and are using translation to understand), or those who are using screen readers. We should optimize for reading. But `ToSchema` and `ToResponse` seem like a reasonable option to me, let's go with that. I've renamed the `ResponseComponent` to `ToResponse`. I can do the `Component` to `ToSchema` rename in the next PR. 

Is this PR ready now do you think?

juhaku (approved):
Awesome.. :+1:

src/lib.rs
juhaku: Yeah you are right.. Let's get the ball rolling and get this thing to the upstream. Thanks for your work :heart:

PR #218, opened 2022-07-13
Implement additionnal properties for objects

traxys:
This allows to specify the component to be used for additionnal properties.

### DIFF
--- a/utoipa-gen/src/schema.rs
+++ b/utoipa-gen/src/schema.rs
@@ -93,7 +93,18 @@ impl<'a> ComponentPart<'a> {
                 {
                     None
                 } else {
-                    Some(ComponentPart::get_generic_arg_type(0, angle_bracketed_args))
+                    match generic_component_type.generic_type {
+                        Some(GenericType::Map) => Some(ComponentPart::get_generic_arg_type(
+                            0,
+                            angle_bracketed_args,
+                            true,
+                        )),
+                        _ => Some(ComponentPart::get_generic_arg_type(
+                            0,
+                            angle_bracketed_args,
+                            false,
+                        )),
+                    }
                 }
             }
             _ => abort!(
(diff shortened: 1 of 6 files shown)


traxys:
I think this should be automatically infered or at least available using the Component derive, but I don't really know what would be the best interface, or if it should be tried for all HashMap

juhaku (commented):
How about adding a test which would show and then assert that the generated json is actually correct :) You can use the `assert_json_eq!` macro to assert json and then look examples from `openapi.rs` `schema.rs` or `security.rs` for writing tests. Or from any of the integration tests in `tests` folder.

juhaku:
> I think this should be automatically infered or at least available using the Component derive, but I don't really know what would be the best interface, or if it should be tried for all HashMap

Yes this should be inferred from users using any `map` type e.g `HashMap` or `BTreeMap` etc. I guess these two maps are sufficient since they are provided by the standard library. The infer should happen in `utoipa-gen` library. 

Currently the maps are just treated as generic objects. So there would be some changes to be done to the behaviour so that the types can be inferred from the maps. One of the places need changes is `ComponentPart` struct which currently does not infer types for `maps` correctly (because there has not been a need for).

traxys:
I added the test in the first commit, and I am looking into the derive macro, but I need some time to understand the code

traxys:
I think I have a functionnal implementation, but as I don't really understand the code I may have made a mistake

utoipa-gen/src/schema.rs
```diff
@@ -107,13 +118,20 @@ impl<'a> ComponentPart<'a> {
         generic_component_type
     }
 
-    fn get_generic_arg_type(index: usize, args: &'a AngleBracketedGenericArguments) -> &'a Type {
+    fn get_generic_arg_type(
+        index: usize,
+        args: &'a AngleBracketedGenericArguments,
+        skip_first: bool,
```
juhaku, line 124: Pretty nifty solution this `skip_first`. I originally thought more complicated solution but this actually is fine too. It can be enhanced later. I was considering refactoring this a bit anyways.

utoipa-gen/src/schema/component.rs
```diff
@@ -689,9 +689,18 @@ where
...
+                    self.component_part.child.as_ref().unwrap().as_ref(),
+                    self.comments,
+                    self.attrs,
+                    self.deprecated,
+                    self.xml,
+                );
+
                 tokens.extend(quote! {
-                    utoipa::openapi::ObjectBuilder::new()
+                    utoipa::openapi::ObjectBuilder::new().additional_properties(Some(#component_property))
```
juhaku, line 703: I guess this implementation need to be added to `into_params.rs` for request parameters.

juhaku (commented):
Super, just one thing to test and add the same functionality to `into_params.rs` and it should be good to go :trophy:

traxys:
I have rebased the branch on the current master, and have added the `into_param` implementation & test

juhaku (approved):
Ok :ok_hand: :medal_military:

Ryan-DL:
@juhaku Has this been implemented in the newest release? I'd love to use this 😄

juhaku:
> @juhaku Has this been implemented in the newest release? I'd love to use this smile

This has been implemented and is currently in master but has not been yet released to the crates. This will be available in next 2.0.0 release once released. Alternatively you may use the git master branch as a source in your Cargo.toml file but be aware that breaking changes might occur still before the actual release is out. 
```toml
utoipa = { git = "https://github.com/juhaku/utoipa", branch = "master" }
```

Issue #231, opened 2022-07-25
(Master) the name `_AssertComponentSomeComponent` is defined multiple times


RemiKalbe:
I'm trying to do the following
```rust
#[openapi
(
    ...
    components(
        handlers::some_mod::another_one::one_last::request::body::SomeComponent as request::body::SomeComponent,
        handlers::some_mod::another_one::one_last::response::SomeComponent as response::SomeComponent,
    )
)]
```
but this gives me the following error
```rust
error[E0428]: the name `_AssertComponentSomeComponent` is defined multiple times
   --> src/main.rs:215:89
    |
214 | ...ast::request::body::SomeComponent as request::body::SomeComponent,
    |                         -------------------- previous definition of the type `_AssertComponentSomeComponent` here
215 | ...ast::response::SomeComponent as response::SomeComponent,
    |                    ^^^^^^^^^^^^^^^^^^^^ `_AssertComponentSomeComponent` redefined here
    |
    = note: `_AssertComponentSomeComponent` must be defined only once in the type namespace of this block
```
I thought https://github.com/juhaku/utoipa/pull/187 solved this issue?

juhaku:
Yeah it originally suppose to fix it. It might be regression caused by me because of the assert structs that was removed and added back and forth. But now on they will be completely removed since they are superfluous and are not providing more context to the errors that Rust compiler might emit. :)

PR #232, opened 2022-07-25
Use the complete path to construct AssertComponent

RemiKalbe:
Fix https://github.com/juhaku/utoipa/issues/231

### DIFF
--- a/utoipa-gen/src/openapi.rs
+++ b/utoipa-gen/src/openapi.rs
@@ -1,5 +1,6 @@
 use proc_macro2::Ident;
 use proc_macro_error::ResultExt;
+use std::ops::Not;
 use syn::{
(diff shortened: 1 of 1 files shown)


juhaku (commented):
Nice, But actually this is superfluous PR since the assert structs will be remove in sort future as they are deemed unnecessary.

RemiKalbe:
Yeah I did ask myself what was the point of that 😆
But at least in the meantime it's possible to have components with the same name?

juhaku:
> Yeah I did ask myself what was the point of that laughing But at least in the meantime it's possible to have components with the same name?

Well it could be merged. But I will now remove all assert instances that still exists in the codebase :) as we very next PR

PR #233, opened 2022-07-25
Make layout configurable between Standalone and Base

ju6ge:
Hey,

I added an option to the Config in Swagger to enable toggling between Standalone and Base Layout.

Kind regards,
ju6ge

### DIFF
--- a/utoipa-swagger-ui/build.rs
+++ b/utoipa-swagger-ui/build.rs
@@ -105,7 +105,8 @@ fn replace_default_url_with_config(target_dir: &str) {
     .iter()
     .collect::<PathBuf>();
 
-    let swagger_initializer = fs::read_to_string(&path).unwrap();
+    let mut swagger_initializer = fs::read_to_string(&path).unwrap();
+    swagger_initializer = swagger_initializer.replace("layout: \"StandaloneLayout\"", "nothing: false");
 
     let replaced_swagger_initializer = RE.replace(&swagger_initializer, "{{config}},");
 
--- a/utoipa-swagger-ui/build.rs
+++ b/utoipa-swagger-ui/build.rs
@@ -106,7 +106,7 @@ fn replace_default_url_with_config(target_dir: &str) {
     .collect::<PathBuf>();
 
     let mut swagger_initializer = fs::read_to_string(&path).unwrap();
-    swagger_initializer = swagger_initializer.replace("layout: \"StandaloneLayout\"", "nothing: false");
+    swagger_initializer = swagger_initializer.replace("layout: \"StandaloneLayout\"", "");
 
     let replaced_swagger_initializer = RE.replace(&swagger_initializer, "{{config}},");
 
--- a/utoipa-swagger-ui/src/lib.rs
+++ b/utoipa-swagger-ui/src/lib.rs
@@ -390,7 +390,7 @@ impl<'a> From<Cow<'static, str>> for Url<'a> {
 
 pub const SWAGGER_STANDALONE_LAYOUT: &str = "StandaloneLayout";
 pub const SWAGGER_BASE_LAYOUT: &str = "BaseLayout";
-fn swagger_default_layout() -> String { SWAGGER_STANDALONE_LAYOUT.to_string() }
+fn swagger_default_layout() -> &'static str { SWAGGER_STANDALONE_LAYOUT }
 
 /// Object used to alter Swagger UI settings.
 ///
@@ -546,7 +546,7 @@ pub struct Config<'a> {
 
     /// [ layout ] the layout of Swagger UI uses, default is "StandaloneLayout"
     #[serde(default = "swagger_default_layout")]
-    layout: String,
+    layout: &'a str,
 }
 
 impl<'a> Config<'a> {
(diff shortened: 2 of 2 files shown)


utoipa-swagger-ui/build.rs
```diff
@@ -105,7 +105,8 @@ fn replace_default_url_with_config(target_dir: &str) {
     .iter()
     .collect::<PathBuf>();
 
-    let swagger_initializer = fs::read_to_string(&path).unwrap();
+    let mut swagger_initializer = fs::read_to_string(&path).unwrap();
+    swagger_initializer = swagger_initializer.replace("layout: \"StandaloneLayout\"", "nothing: false");
```
juhaku, line 109: Maybe just directly replace it to "Nothing" instead :)
```suggestion
    swagger_initializer = swagger_initializer.replace("layout: \"StandaloneLayout\"", "");
```

utoipa-swagger-ui/src/lib.rs
```diff
@@ -388,6 +388,10 @@ impl<'a> From<Cow<'static, str>> for Url<'a> {
     }
 }
 
+pub const SWAGGER_STANDALONE_LAYOUT: &str = "StandaloneLayout";
+pub const SWAGGER_BASE_LAYOUT: &str = "BaseLayout";
+fn swagger_default_layout() -> String { SWAGGER_STANDALONE_LAYOUT.to_string() }
```
juhaku, line 393: I would try to use `&str` here since the are know values and thus no need to allocat them if just possible
```suggestion
fn swagger_default_layout() -> &'static str { SWAGGER_STANDALONE_LAYOUT }
```

utoipa-swagger-ui/src/lib.rs
```diff
@@ -539,6 +543,10 @@ pub struct Config<'a> {
     /// [`oauth::Config`] the Swagger UI is using for auth flow.
     #[serde(skip)]
     oauth: Option<oauth::Config>,
+
+    /// [ layout ] the layout of Swagger UI uses, default is "StandaloneLayout"
+    #[serde(default = "swagger_default_layout")]
+    layout: String,
```
juhaku, line 549: If possible the layout will live at least as long as the config does.
```suggestion
    layout: &'a str,
```

utoipa-swagger-ui/src/lib.rs
```diff
@@ -758,6 +766,24 @@ impl<'a> Config<'a> {
...
+    /// # Examples
+    ///
+    /// Allow operation id to be shown.
+    /// ```rust
+    /// # use utoipa_swagger_ui::Config;
+    /// let config = Config::new(["/api-doc/openapi.json"])
+    ///     .use_base_layout();
+    /// ```
+    pub fn use_base_layout(mut self) -> Self {
+        self.layout = SWAGGER_BASE_LAYOUT.to_string();
```
juhaku, line 782: And just a reference here :)
```suggestion
        self.layout = SWAGGER_BASE_LAYOUT;
```

utoipa-swagger-ui/src/lib.rs
```diff
@@ -758,6 +766,24 @@ impl<'a> Config<'a> {
         self
     }
 
+    /// Set 'layout' to 'BaseLayout' to only use the base swagger layout without a search header.
+    ///
+    /// Default value is 'StandaloneLayout'.
+    ///
+    /// # Examples
+    ///
+    /// Allow operation id to be shown.
```
juhaku, line 775: I guess this is a copy paste after effect. Maybe "Change Swagger UI to BaseLayout"?

juhaku (changes requested):
Nice addition :+1: I would add this new function to the test with name `format_swagger_config_with_multiple_fields` and update the test to cover this `use_base_layout()` case as well :)

utoipa-swagger-ui/src/lib.rs
ju6ge: True that 😅

ju6ge:
I also updated the tests now, but I somehow cannot run the test suit because the build of utoipa-gen fails:
```
error[E0369]: binary operation `==` cannot be applied to type `Option<Cow<'_, TypePath>>`
  --> utoipa-gen/src/schema.rs:30:5
   |
27 | #[derive(PartialEq)]
   |          --------- in this derive macro expansion
...
30 |     pub path: Option<Cow<'t, TypePath>>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the derive macro `PartialEq` (in Nightly builds, run with -Z macro-backtrace for more info)
```
Which is weird because it works when I compile when importing the code from another project 🤨

juhaku (approved):
Code does look fine :1st_place_medal: :)

juhaku:
That's really weird. It should compile though. If you run the `cargo test -p utoipa-swagger-ui --features actix-web,axum,rocket -- --show-output` command it ends up in following error? 

I ran the command above and even with nightly `cargo +nighly test ...` and it passes. Just updated nightly to latest. It should work anything above the 1.60. though :open_mouth:

ju6ge:
Anyway tests are fixed now, that I could actually run them :D …

juhaku:
> But just using cargo test fails …

Right. Most of the tests needs some flags in order to work. And cannot use `--all-features` because then there are clashing implemenetations for different frameworks. 

> Anyway tests are fixed now, that I could actually run them :D …

Awesome.. :trophy:

juhaku (approved):

PR #237, opened 2022-07-29
Use IndexMap to preserve insertion order for content_types

### LINKED ISSUE #235: Order is not preserved for `response.content_types`
Given the following path,

```
#[utoipa::path(
    request_body(content=MyRequest, content_type="application/json"),
    responses(
        (status = 200, content_type=["application/json", "application/cbor"], description = "RPC call was successful", body = FindByIdResponse),
        (status = 400, content_type=["application/json", "application/cbor"], description = "Bad Request", body = RpcError),
        (status = 401, content_type=["application/json", "application/cbor"], description = "Forbidden", body = RpcError),
        (status = 500, content_type=["application/json", "application/cbor"], description = "Internal Server Error", body = RpcError)
    ),
)]
```

The resulting spec will not be serialized in the same order it is specified. I see `application/cbor` come first in the list. 

This matters because in Swagger UI, the first one in the list is the default example, which should be JSON because it is not possible to play with CBOR in the browser.

```
        "responses": {
          "200": {
            "description": "RPC call was successful",
            "content": {
              "application/cbor": {
                "schema": {
                  "$ref": "#/components/schemas/FindByIdResponse"
                }
              },
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/FindByIdResponse"
                }
              }
            }
          },
          "400": {
            "description": "Bad Request",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/RpcError"
                }
              },
              "application/cbor": {
                "schema": {
                  "$ref": "#/components/schemas/RpcError"
                }
              }
            }
          },
          "401": {
            "description": "Forbidden",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/RpcError"
                }
              },
              "application/cbor": {
                "schema": {
                  "$ref": "#/components/schemas/RpcError"
                }
              }
            }
          },
          "500": {
            "description": "Internal Server Error",
            "content": {
              "application/cbor": {
                "schema": {
                  "$ref": "#/components/schemas/RpcError"
                }
              },
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/RpcError"
                }
              }
            }
          }
        },
        ```

okdistribute:
Fixes #235

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -68,8 +68,10 @@ jobs:
           cargo test --test component_derive_test --features chrono,decimal,uuid
           cargo test --test component_derive_test --features chrono_with_format
           cargo test --test path_derive_rocket --features rocket_extras,json
+          cargo test --test path_derive_axum_test --features axum_extras,json
+          cargo test --test utoipa_gen_test --features json,yaml
         elif [[ "${{ matrix.testset }}" == "utoipa-gen" ]] && [[ ${{ steps.changes.outputs.gen_changed }} == true ]]; then
           cargo test -p utoipa-gen --features utoipa/actix_extras
         elif [[ "${{ matrix.testset }}" == "utoipa-swagger-ui" ]] && [[ ${{ steps.changes.outputs.swagger_changed }} == true ]]; then
-          cargo test -p utoipa-swagger-ui --features actix-web,rocket
+          cargo test -p utoipa-swagger-ui --features actix-web,rocket,axum
         fi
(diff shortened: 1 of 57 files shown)


juhaku (commented):
Nice. At first I thought to implement this behavior manually as struct with having inner type `Vec` holding items in the insertion order and then manually implementing serde functionality. But seems like this `indexmap` does not affect the performance thus we could use it too though.

I'd also add a mention here to the docs https://github.com/juhaku/utoipa/blob/master/utoipa-gen/src/lib.rs#L453 that or order is preserved for multiple types. "When the multiple content types are defined the insertion order is preserved in OpenAPI doc." Or something like this.

PR #238, opened 2022-07-29
Only enable the "deflate" feature of the zip crate

ETCaton:
This saves ~3-4 seconds on release builds on my local machine per `cargo build --timings`


juhaku (approved):
Awesome :1st_place_medal:  This indeed really good optimization. thanks :)

PR #240, opened 2022-08-01
Feature `openapi_extensions` with helper functions for JSON responses and requestBodies

erewok:
Following our [discussion](https://github.com/juhaku/utoipa/discussions/236) on adding some helper functions for documenting a JSON `requestBody` or `response`, this PR includes the following, all gated by the feature-flag suggested in the discussion, `openapi_extensions`:

- New Trait `RequestBodyExt` for easily documenting JSON `requestBody` content (implemented for `RequestBodyBuilder` ).
- New Trait `ResponseExt` for easily documenting JSON `response` content (implemented for `ResponseBuilder`).
- Includes tests for new traits and methods
- Includes documentation for these traits

In addition, the following changes are also included:
- Adds `from_response_name` for `Ref` struct for `"#/components/responses/SomeObject"` references
- Adds `-F openapi_extensions` to `cargo tests` github actions workflow

## Questions

Having traits in this case may be unnecessary because these methods are only implemented in each case for a single `struct`. We could do one of the following [with these traits]:

- Create a _single_ trait that is shared by `RequestBodyBuilder` and `ResponseBuilder` (instead of two separate traits). The trait could include common shared things, but _may_ result in people using `response` objects in requests or otherwise blending the two different types. (A single shared trait is originally how I did this in my own project.)
- Drop the traits and implement these as methods directly on `RequestBodyBuilder` and `ResponseBuilder`.
- Alternatively, I could keep the two new traits but also implement `RequestBodyBuilderExt` (probably renaming the trait) for `RequestBody` and `ResponseBuilderExt` for `Response` (also renaming the trait) instead of just for the builders and then having traits would probably be justified.

Happy to implement one of the above or leave this as-is. Also, happy to take any feedback or make any changes requested.

### DIFF
--- a/src/openapi/response.rs
+++ b/src/openapi/response.rs
@@ -152,15 +152,19 @@ impl ResponseBuilder {
 ///
 #[cfg(feature = "openapi_extensions")]
 pub trait ResponseExt {
+    /// Add [`Content`] to [`Response`] referring to a schema
+    /// with Content-Type `application/json`.
     fn json_component_ref(self, ref_name: &str) -> Self;
+
+    /// Add [`Content`] to [`Response`] referring to a response
+    /// with Content-Type `application/json`.
     fn json_response_ref(self, ref_name: &str) -> Self;
 }
 
 
 #[cfg(feature = "openapi_extensions")]
 impl ResponseExt for Response {
-    /// Add [`Content`] to [`Response`] referring to a component-schema
-    /// with Content-Type `application/json`.
+
     fn json_component_ref(mut self, ref_name: &str) -> Response {
         self.content.insert(
             "application/json".to_string(),
@@ -169,8 +173,6 @@ impl ResponseExt for Response {
         self
     }
 
-    /// Add [`Content`] to [`Response`] referring to a component-response
-    /// with Content-Type `application/json`.
     fn json_response_ref(mut self, ref_name: &str) -> Response {
         self.content.insert(
             "application/json".to_string(),
@@ -182,8 +184,7 @@ impl ResponseExt for Response {
 
 #[cfg(feature = "openapi_extensions")]
 impl ResponseExt for ResponseBuilder {
-    /// Add [`Content`] to [`Response`] referring to a component-schema
-    /// with Content-Type `application/json`.
+
     fn json_component_ref(self, ref_name: &str) -> ResponseBuilder {
         self.content(
             "application/json",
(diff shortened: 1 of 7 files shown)


juhaku:
> * Create a single trait that is shared by RequestBodyBuilder and ResponseBuilder (instead of two separate traits). The trait could include common shared things, but may result in people using response objects in requests or otherwise blending the two different types. (A single shared trait is originally how I did this in my own project.)
> * Drop the traits and implement these as methods directly on RequestBodyBuilder and ResponseBuilder.

I think the two seprate traits are good since then it allows having possible separete set of helpers for each type which is a good thing for separating consern.

> * Alternatively, I could keep the two new traits but also implement RequestBodyBuilderExt (probably renaming the trait) for RequestBody and ResponseBuilderExt for Response (also renaming the trait) instead of just for the builders and then having traits would probably be justified.

Yeah these could be implemented for `Response` and `RequestBody` as well thus naming the traits as `ResponseExt` and `RequestBodyExt` is reasonable.

juhaku (approved):
This looks awesome :+1: can't wait to get more utilities to the library.

erewok:
Sounds great, @juhaku! Thanks for the suggestions. 

I have now made the changes discussed above:

- Rename traits: `ResponseExt` and `RequestBodyExt`
- Implement `RequestBodyExt` for `RequestBody`
- Implement `ResponseExt` for `Response`

I think this PR is ready for one more review and if you are satisfied with it, we can merge. Alternately, I am happy to make any other changes you would prefer.

src/openapi/response.rs
```diff
@@ -114,14 +114,178 @@ impl ResponseBuilder {
...
+pub trait ResponseExt {
+    fn json_component_ref(self, ref_name: &str) -> Self;
+    fn json_response_ref(self, ref_name: &str) -> Self;
+}
+
+
+#[cfg(feature = "openapi_extensions")]
+impl ResponseExt for Response {
+    /// Add [`Content`] to [`Response`] referring to a component-schema
+    /// with Content-Type `application/json`.
```
juhaku, line 163: Not sure if this doc comment will be shown in docs actually when it is placed over the implementation.  Same for the other instances below :point_down:  If you run this command `cargo doc --workspace --no-deps --features actix_extras,json,debug` and see from the docs whether it is there. I think this doc comment needs to be placed on the trait function itself [line 155](https://github.com/juhaku/utoipa/pull/240/files#diff-eea3ccce264a9afded89ef39496101cb1a5bf8c1fedc4cd4ddf9221e4744d922R155).

juhaku (approved):
Super. :1st_place_medal:

PR #248, opened 2022-08-17
fix(utoipa-gen): FromStr impl for serde's PascalCase in rename_all

CBenoit:
Serde's value for pascal case is "PascalCase"


juhaku (approved):
Sweet :candy:

PR #250, opened 2022-08-18
update SwaggerUi to version 4.14.0

ju6ge:


juhaku (commented):
Nice, :+1:  did you try to run some of the examples with the newest swagger? Or any of your own projects?

ju6ge:
Hey I am using it for a private project. It does seem to work, but sadly the bug I had hoped that it would fix, still persists. But that is probably due to the fact that it is an upstream problem. Anyway from what I can tell it works just as well as the previous version with some extra features.

juhaku:
Great, lets have it updated, so the next release will also get updated swagger :slightly_smiling_face:

PR #273, opened 2022-09-08
Add caution on export rule

komi1230:
Hello,

Thank you for developing such an awesome project.
I'm developing an application with Rust and Axum and I've found this crate is very helpful for documenting API spec.

When I started to use `utoipa`, I was stuck in how to export handler function since I put the handler function in different file and exported this.
And finally I solved this with exporting a function named with `__path_` suffix.

When developing a big project, many people can split modules and cut off handler function in different file.
So I think this message would be very helpful for those people.

Thank you checking !

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,9 @@ mod pet_api {
 }
 ```
 
+This attribute macro substantially create another function named with `__path_` prefix.
+So when you implement `some_handler` function in different file and want to export this, make sure `__path_some_handler` in the module can also be accessible from the root.
+
 Tie the `Schema` and the endpoint above to the OpenApi schema with following `OpenApi` derive proc macro.
 
 ```rust


juhaku:
Yeah, indeed, the handler with `path` macro and essentially the `__path_` prefixed struct implementing the path for the handler need to be visible to the `OpenApi` trait to register them. There is planned effort to improve this a bit for bigger projects wherein people tend to split the functionality to separate modules. By allowing users to define multiple `OpenApi` trait implementing structs and merge them in root level. This way developers only need to export the struct to the root and can leave the implementation private to the module. 

There already are people doing so but currently there is no built in support for this. There is an existing discussion related to this might be worth linking here: https://github.com/juhaku/utoipa/discussions/208

README.md
```diff
@@ -142,6 +142,9 @@ mod pet_api {
 }
 ```
 
+This attribute macro substantially create another function named with `__path_` prefix.
```
juhaku, line 145:
```suggestion
This attribute macro substantially will create another struct named with `__path_` prefix + handler function name.
```

juhaku (approved):
This is :ok_hand:  Just a minor edit to the text :slightly_smiling_face:

juhaku:
> Thank you for quick review !

Anytime :slightly_smiling_face:

Issue #271, opened 2022-09-06
Support for the repr attribute


Naymoll:
I have this type:
```
#[derive(Debug, Clone, Serialize_repr, Deserialize_repr)]
#[derive(ToSchema)]
#[repr(u8)]
pub enum OsrmVersion {
    Osrm4 = 0,
    Osrm5,
}
```
Generated output: 
```
"OsrmVersion":{
    "type":"string",
    "enum":[
        "Osrm4",
        "Osrm5" 
    ]
}
```
Expected output: 
```
"OsrmVersion":{
    "type":"integer",
    "enum":[
        0,
        1
    ]
}
```
Is this possible to add support for the repr attribute? Or did I miss something?

juhaku:
Currently there is no support for `repr` but it could be investigated whether the support for this could be added and how much effort it would take. This is pretty nice idea though. I assume that the support should not be too difficult but cannot make any statements nor promises beforehand.

juhaku:
I'll close this now since the PR has been merged to master.

PR #282, opened 2022-09-14
Support for repr attribute

Naymoll:
I tried to write a simple impl support of repr attribute. Right now there is no support for serde tags.

If we have something like this:
```
#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, ToSchema)]
#[repr(u8)]
pub enum OsrmVersion {
    Osrm4 = 0,
    Osrm5,
}
```

Macro generates the following output:
```
impl utoipa::ToSchema for OsrmVersion {
    fn schema() -> utoipa::openapi::schema::Schema {
            .schema_type(utoipa::openapi::SchemaType::Integer)
            .enum_values(Some([Self::Osrm4 as u8, Self::Osrm5 as u8]))
            .into()
    }
}
```
`*.json` file
```
"OsrmVersion": {
    "type": "integer",
    "enum": [
        "0",
        "1"
}
```

Related to: https://github.com/juhaku/utoipa/issues/271


Naymoll:
Also, I haven't checked how schema attributes affect macro generation.

juhaku:
This is a pretty good start :muscle: but support for serde attributes would be beneficial,  but the rename of enum at container level is no prio since throughout the lib there is no such support and implementing it needs more planning. And tests would be be useful. There are plenty in `schema_derive_test.rs` file :slightly_smiling_face:

Naymoll:
I think the `tag` and `skip` attributes need to be supported from serde. The others do not make sense. I will try to write an implementation. Tests as well.

Naymoll:
I noticed one thing. Even if the enum type is marked as `integer`, the value itself is still a string. So when I make a request to the server, it sends this:
```
"configuration": {
    ...
    "osrm_version": "0",
    ...
}
```
Which is incorrect:
```
{"error":"Json deserialize error: invalid type: string \"0\", expected u8 at line 7 column 27"}
```

I've already changed `enum_values`, so you can pass whatever that impl `ToString`. But I think that it's wrong. And more changes are needed. Any advice?

juhaku:
Keep in mind that the OpenAPI specification does not allow integer type of enums. https://swagger.io/specification/ If you browse to the spec and search for `enum` you find that it is always used as array of `string` types.

Naymoll:
Didn't know that. Then all that's left is to write the tests.

Naymoll:
I think that's all for now.

juhaku:
> Keep in mind that the OpenAPI specification does not allow integer type of enums. https://swagger.io/specification/ If you browse to the spec and search for enum you find that it is always used as array of string types.

Sorry, I eat my words, I tested it in editor.swagger.io and seems like the enum can also be represented according the type of the field. e.g.
```yaml
components:
  schemas:
    Order:
      type: object
      properties:
        status:
          type: integer
          description: Order Status
          enum:
            - 0
            - 1
            - 2
```

juhaku:
> I've already changed enum_values, so you can pass whatever that impl ToString. But I think that it's wrong. And more changes are needed. Any advice?

Actually to have it support dynamic enum values there is a need to change the `enum_values` representation within the `Object` type in `utoipa` crate.
```rust
pub struct Object {
// ...
    #[serde(rename = "enum", skip_serializing_if = "Option::is_none")]
    pub enum_values: Option<Vec<String>>,
```
The string needs to be changed to `serde_json::Value` see the `example` field. And then of course fix all the existing usages which currently expect it to be `string`.

juhaku (commented):
Overall this is okay there are 2 things still address here :slightly_smiling_face: 
- [ ] Add this `repr` create support behind a `repr` feature flag, so people can enable it when they want.
- [ ] Add docs about this new functionality to `utoipa-gen/src/lib.rs` and list the new feature also in `README.md` and in `utoipa/src/lib.rs`

You can generate the doc with, and then use browser to see the output. 
```bash
cargo doc --workspace --no-deps --features utoipa/json,utoipa/actix_extras,utoipa/openapi_extensions
```

juhaku (approved):
:trophy: pretty good, let's get it merged

Issue #297, opened 2022-09-25
Define default response


jacob-pro:
Hi :)
thanks for this great crate.

I'm wondering if it is possible to define a default response?

Default responses are documented here: https://swagger.io/docs/specification/describing-responses/, for example in an API spec:

```yaml
      responses:
        '200':
          description: Success
        'default':
          "$ref": "#/components/responses/ApiError"
```

I was expecting to be able to use something like this, but it doesn't seem to work:

```
    responses(
        (status = 200, description = "Success"),
        (status = "default", description = "Error")
    ),
```

juhaku:
Hi, anytime :slightly_smiling_face: 

> I'm wondering if it is possible to define a default response?

That is actually unimplemented feature. One way to do it would  be to allow not to define the status what would then make it the "default" response. Or add another attribute `default` which can be provided in place of the status when the default response is defined. 

This is definitely something that could be added in the future. And should not be too hard to make. PR:s are also welome :slightly_smiling_face:

jacob-pro:
Hi @juhaku ,

Yeah I'm happy to look at writing a PR...

I'm wondering though if it would make more sense for status to just be a string?

The spec also allows for range matches as well, such as "4XX" and "5XX"

So maybe the simplest would be to allow doing?

```
    responses(
        (status = "200", description = "Success"),
        (status = "4XX", description = "Bad Request"),
        (status = "5XX", description = "Server Error"),
        (status = "default", description = "everything else")
    ),
```

juhaku:
> I'm wondering though if it would make more sense for status to just be a string?

To be honest, originally it was just a `String` but then it was changed to a number  be more efficient via not using that much space. So it was turned to a int internally. :smile: But true, it could be a string as you proposed. Also it could be an enum internally which would provide some validation over allowed types e.g. 
```rust
enum StatusCode {
  Number(i32),
  Default,
  1XX,
  2XX,
  // .. etc
}
```
And this enum above could be parsed from the strings represented in your comment.

This change should be fairly even with tests.

jacob-pro:
Looks like it is already a string internally?
https://github.com/juhaku/utoipa/blob/84f969e7f14b819d5f54f1fd4914ac39e6c0d21c/utoipa/src/openapi/response.rs#L25-L29

juhaku:
[...]

PR #298, opened 2022-09-25
#297 - Allow using default and range matched status responses

jacob-pro:
To hopefully close https://github.com/juhaku/utoipa/issues/297

Changes the macro to allow passing strings for the response status code.

This is necessary to use things like "default" or "5XX"

### DIFF
--- a/utoipa-gen/src/path/response.rs
+++ b/utoipa-gen/src/path/response.rs
@@ -48,6 +48,9 @@ pub struct ResponseValue<'r> {
 impl Parse for ResponseValue<'_> {
     fn parse(input: syn::parse::ParseStream) -> syn::Result<Self> {
         const EXPECTED_ATTRIBUTE_MESSAGE: &str = "unexpected attribute, expected any of: status, description, body, content_type, headers";
+        const VALID_STATUS_RANGES: &[&str] = &["default", "1XX", "2XX", "3XX", "4XX", "5XX"];
+        const INVALID_STATUS_RANGE_MESSAGE: &str = "Invalid status range, expected one of:";
+
         let mut response = ResponseValue::default();
 
         while !input.is_empty() {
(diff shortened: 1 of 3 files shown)


utoipa-gen/tests/path_response_derive_test.rs
```diff
@@ -58,7 +58,9 @@ test_fn! {
         (status = 200, description = "success"),
         (status = 401, description = "unauthorized"),
         (status = 404, description = "not found"),
-        (status = 500, description = "server error")
+        (status = 500, description = "server error"),
+        (status = "5XX", description = "all other server errors"),
+        (status = "default", description = "default")
```
juhaku, line 63: This is ok, but would be more beneficial if asserts were also added under line 83.

utoipa-gen/src/path/response.rs
```diff
@@ -62,8 +62,18 @@ impl Parse for ResponseValue<'_> {
...
+                            // And also strings for "default" "4XX" etc.
+                            let lookahead = input.lookahead1();
+                            if lookahead.peek(LitInt) {
+                                input.parse::<LitInt>()?.base10_parse()
+                            } else if lookahead.peek(LitStr) {
+                                Ok(input.parse::<LitStr>()?.value())
+                            } else {
+                                Err(lookahead.error())
+                            }
+                        })?
```
juhaku, line 76: While this is okay, I'm also wondering that this does not limit the the user of defining a value as "sflksfjl" just random string, and when such is defined user will only see error when he opens the browser with the documentaiton.

juhaku (commented):
Docs also should be updated regarding changed behaviour. https://github.com/juhaku/utoipa/blob/84f969e7f14b819d5f54f1fd4914ac39e6c0d21c/utoipa-gen/src/lib.rs#L505

You can generate the docs with: `cargo doc --workspace --no-deps --features utoipa/json,utoipa/actix_extras,utoipa/openapi_extensions` and then use your favourite browser to view them :slightly_smiling_face:

utoipa-gen/tests/path_response_derive_test.rs
jacob-pro: thanks yep I wasn't entirely sure, how these tests worked, have added now

utoipa-gen/src/path/response.rs
jacob-pro: Ok I have changed it to either accept integer literals, or strings from a validated list

juhaku: Awesome :+1:

juhaku (approved):
Super :trophy:

Issue #300, opened 2022-09-25
Use component response in path responses list


jacob-pro:
So I have defined a re-usable response using the `OpenApi` macro:

```
        responses(
            api::error::ApiError,
        )
```

Where I have implemented:

```rust
impl ToResponse for ApiError {
    fn response() -> (String, Response) {
         ...
    }
}
```

I would now like to be able to use it in my paths, e.g. something like:

```rust
#[utoipa::path(
    post,
    responses(
        (status = 200, description = "Success", body = ListObjects),
        (status = "default", response = ApiError),
    ),
)]
```

As far as I can see this isn't currently possible right now? (Unless I have to manually create a custom `IntoResponses` for every single path?)

juhaku:
Yeah, there is a already closed discussion regarding this. The derive implementation is unimplemented for responses at the moment.  See more here https://github.com/juhaku/utoipa/issues/201. Though the traits are in place somehow, but might need some tuning once the actual derive implementation gets there. 

I believe there are some people how use the existing traits to build macro rules to imitate the derive functionality partly. 

Eventually when this gets implemented, it should be feasible to just define the response itself within the `responses(...)` attribute.
```rust
#[utoipa::path(
    post,
    responses(
        (status = 200, description = "Success", body = ListObjects),
        ApiError
    ),
)]
```

jacob-pro:
Sorry I'm not sure I quite follow... The status code is still needed though?

E.g. the end spec would look like:

```yaml
  /users/{id}:
    get:
      summary: Gets a user by ID.
      response:
        '200':
          description: OK
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/User'
        '401':
          $ref: '#/components/responses/Unauthorized'   # <-----
        '404':
          $ref: '#/components/responses/NotFound'       # <-----
```

so we would need some syntax of mapping a code to the response?

If you have an idea of a syntax which would be acceptable I would be happy to look at writing a PR?

juhaku:
Supporting derivation of responses is not a small feature and I hope to be able to refactor he codebase a bit before adding them to mitigate furhter fragmentation of the code and reduce unnecessary duplication in some areas. Like now there is almost 2 similar implementations `IntoParams` and `ToSchema` but are still different since they serve different purpose, but both of them share a same ideology with the new `IntoResponses` or also perhaps `ToResponse` derive implementation. 

While they would not be completely same, but similary structure from the main parts would be present. 

[...]

PR #303, opened 2022-09-26
Allow using a response reference in the list of responses

jacob-pro:
First of all this updates the Responses struct to have:

```rust
pub responses: BTreeMap<String, RefOr<Response>>,
```

Which should now correctly match the way the OpenAPI specs work:

```yaml
      responses:
        '200':
          description: Success
        '400':
          "$ref": "#/components/responses/BadRequest"
```

And then it enables you to use a reference in the `path` macro:

```
    responses: (
        (status = 200, description = "Success"),
        (status = 400, response = crate::BadRequestResponse),
    )
```

where `crate::BadRequestResponse` is a type that implements `ToResponse` and is used in:

```rust
#[openapi(
    components(
        responses(
            crate::BadRequestResponse,
        )
    ),
)]
```

(Note I had to use `response ` instead of just `ref` since it is a reserved keyword)

This should close https://github.com/juhaku/utoipa/issues/300

### DIFF
--- a/utoipa-gen/src/lib.rs
+++ b/utoipa-gen/src/lib.rs
@@ -551,7 +551,7 @@ pub fn derive_to_schema(input: TokenStream) -> TokenStream {
 ///
 /// ```text
 /// responses(
-///     (status = 200, response_ref = ReusableResponse)
+///     (status = 200, response = ReusableResponse)
 /// )
 /// ```
 ///
--- a/utoipa-gen/src/lib.rs
+++ b/utoipa-gen/src/lib.rs
@@ -545,6 +545,16 @@ pub fn derive_to_schema(input: TokenStream) -> TokenStream {
 /// )
 /// ```
 ///
+/// **Reference a reusable response type:**
+///
+/// `ReusableResponse` must be a type that implements [`ToResponse`][to_response_trait]
+///
+/// ```text
+/// responses(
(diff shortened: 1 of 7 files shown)


jacob-pro:
Hey @juhaku  - please can you let me know what you think of this?

juhaku:
> Hey @juhaku - please can you let me know what you think of this?

Hi @jacob-pro I will try to check this out today / tomorrow. I am bit busy during these days unfortunately. Thanks for patience. :slightly_smiling_face: . 

I took a quick look and overall looks quite good. One thing I am wondering whether the `response_ref` should actually just be a `response`. This would work as well as a first go. But with `response` we could introduce possibility to inline the response as well. 

Also the `Parse` implementation is currently implemented to `ResponseValue` struct. But the response could also be changed to an enum like.
```rust
enum Response {
  Value(ResponseValue),
  Ref(ResponseRef),
}
```

jacob-pro:
@juhaku No problem, thanks for all the quick responses!

I have updated the PR as requested - hopefully I have understood correctly?

juhaku (approved):
Looks good, awesome :1st_place_medal:

juhaku:
> I have updated the PR as requested - hopefully I have understood correctly?

Yeah pretty much yes :slightly_smiling_face:

PR #304, opened 2022-09-29
recognize `http::status::Statuscode` paths in status

### LINKED ISSUE #301: Allow the the statuscode in the `path` macro to be a `http::StatusCode`?
Currently, the statuscode in the `utoipa::path` macro is either an integer literal, or a wildcard string of the form `"5XX"`. Could we support `http::StatusCode`s too?

Present:

```
#[utopia::path(
    responses(status = 200, description = "Success")
)]
```

What I hope to introduce:

```
use http::StatusCode;

#[utopia::path(
    responses(status = StatusCode::OK, description = "Success")
)]
```

---

`http::StatusCode` supports `ToString` as well as `Into<u16>` which can be used to implement the conversions. I'd be willing to pick this up, if this is something that `utoipa` would like to have!

nerdypepper:
i have pasted the mapping verbatim from the `http` crate (i don't think they are bound to change), and added support for idents or paths in the `status` field. this allows for the following patterns:

```
    responses(
        (status = OK),
        (status = StatusCode::BAD_REQUEST),
        (status = http::StatusCode::INTERNAL_SERVER_ERROR),
    ),

```

note that this does not vaildate the presence of the `http` crate, so above are valid patterns even if those paths are not in scope. 

---

couple of questions:

- would you want this behind a feature flag?
- where should this be documented?

Resolves #301


juhaku:
> i have pasted the mapping verbatim from the http crate (i don't think they are bound to change)

Yeah they are quite static in form thus I reckon its fine. At first I thought that maybe changing the `status_code` field to TokenStream would be sufficient which would allow us to define the status code like so `quote! { http::HttpStatus::ACCEPTED.as_str() }`. But this approach does not validate the status codes which makes this approach not smart enough.

> would you want this behind a feature flag?

Maybe it is not necessary after thinking this again.

> where should this be documented?

It should be documented here: https://github.com/juhaku/utoipa/blob/ab555d6cc272baaa49bb5d816ed718aa92a2ce71/utoipa-gen/src/lib.rs#L505 
And possibly an example could  be added below as well.

And perhaps here and in README could mention something like "Utoipa has a builtin support for http crate's StatusCode" https://github.com/juhaku/utoipa/blob/3920a7494021316c84aaa14479ab912fa30f0905/utoipa/src/lib.rs#L80

utoipa-gen/src/path/response.rs
```diff
@@ -64,24 +68,42 @@ impl Parse for ResponseValue<'_> {
...
-                                if !VALID_STATUS_RANGES.contains(&value.as_str()) {
-                                    return Err(Error::new(
-                                        input.span(),
-                                        format!("{} {}", INVALID_STATUS_RANGE_MESSAGE, VALID_STATUS_RANGES.join(", ")),
-                                    ))
-                                }
-                                Ok(value)
-                            } else {
-                                Err(lookahead.error())
+                    response.status_code = parse_utils::parse_next(input, || {
```
juhaku, line 71: I would change the `status_code` to `Cow<'r, str>` if possible. This will allow to use less allocations. see below :point_down:

utoipa-gen/src/path/response.rs
```diff
@@ -64,24 +68,42 @@ impl Parse for ResponseValue<'_> {
...
+                            input
+                                .parse::<ExprPath>()?
+                                .path
+                                .segments
+                                .last()
+                                .and_then(|segment| {
+                                    STATUS_CODES
+                                        .into_iter()
+                                        .find(|&(_, path)| segment.ident == path)
+                                        .map(|&(code, _)| code.to_owned())
```
juhaku, line 98: Here if `status_code` was `Cow<'r, str>` we could just use:
```suggestion
                                        .map(|&(code, _)| Cow::Borrowed(code))
```

juhaku (commented):
Pretty okay, just a few things to check.

You can generage doc with following command: `cargo doc --workspace --no-deps --features utoipa/json,utoipa/actix_extras,utoipa/openapi_extensions` and then use any browser to browse files.

Issue #276, opened 2022-09-11
Status and Vision for Next few months


juhaku:
# Status

Some time back, a disaster in family happend, thus forcing me to focus on managing things related to it. In Finland we have the saying "ei onnettomuus kulje kello kaulassa" transliterated to english "no accident goes around  clock haning on neck" and probably the closest thing to come by with equivalent in english would be "Accidents don’t happen out of bad luck always". Truly sometimes life is unpredictable and that is the name of the game. This evidently leaves me with very little to no time to further develop this library on the scale I would wish to. I'm not abandoning the library but the speed of development will decrease for time being and only small features / fixes / issues and production bugs stalling people from using the library will be put to the first priority. 

# Timeline

The slow down in effort put to this library will be effective for foreseeble future approximately 1 - 3 months. After processing the matter at hand is finished it can be expected for me to resume putting the regular effort for maintaining the library. 

# Vision and what's there to do

At least few things to expect in future releases when the time allows me to work on the library.
- Serde support for  `IntoParam` as it is in `ToSchema`
- `IntoRespones` and maybe `ToResponse` derive implementations and perhaps also serde support as well.
- `OpenApi` derive macro improvements
- `OpenApi` improvements in general, e.g. merging OpenAPI docs allowing users to split the documentation to separate modules which by now is also possible but there is no built in support for merging the docs.
- Generalization of utoipa-gen internals (not visible to users)
- Maybe redocly support
- Maybe book and other docs and pages etc. 
- More examples e.g. a large api example with utoipa, saving openapi doc to file for later usage e.g. generating clients from the generated openapi doc

Furhtermore if you are eager to have some of these quicker in the library or even if there is some other feature that you would perceive missing or would be needed, feel free too participate. I'll write hopefully soon the code of conduct and contributing guidelines for futher assistance to get started.

juhaku:
So what exactly happened, and how things are going. I thought I could give a short update on the things. My father had passed away what led me to focus on dealing bureaucracy that followed after. Now that almost all matters are passed and they are not taking a big portion of my time anymore I can again give more focus toward the library. This should give a boost for the progress but still is totally developed during freetime. It seesm there are some rough edges that needs a bit of polish, some functionalities are not as feature complete as they could be and then of course items listed above. I try to prioritise the work by what seems most beneficial at the moment. Thank you for understanding. :heart:

PR #309, opened 2022-10-04
allow renaming `IntoParams` field names

fluxxu:
Use case:

```rust
#[derive(Debug, Deserialize, ToSchema, IntoParams)]
pub struct GetInventorySupplyInput {
  pub warehouse_id: String,
  #[param(rename = "sku_list[]")]
  pub sku_list: Option<Vec<String>>,
}
```

Without the rename, `sku_list` is serialized as `sku_list=1&sku_list=2`, but to use with `axum::Query` we need `sku_list[]=1&sku_list[]=2`

### DIFF
--- a/utoipa-gen/src/component/into_params.rs
+++ b/utoipa-gen/src/component/into_params.rs
@@ -273,9 +273,15 @@ impl ToTokens for Param<'_> {
                     .unwrap_or_abort()
             });
 
-        tokens.extend(quote! { utoipa::openapi::path::ParameterBuilder::new()
-            .name(#name)
-        });
+        if let Some(IntoParamsFieldParamsAttr { rename: Some(ref name), ..}) = field_param_attrs {
+            tokens.extend(quote! { utoipa::openapi::path::ParameterBuilder::new()
+                .name(#name)
+            });
+        } else {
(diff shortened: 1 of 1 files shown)


DerZwergGimli:
much needed please add this in a new version - the fork is working!

i need it for renaming 'type' to something else!

```rs
#[derive(Debug, Deserialize, IntoParams)]
pub struct SearchParams {
    #[param(style = Form, example = "FOOD")]
    query: Option<String>,
    #[param(rename = "type")]
    stype: Option<String>,
    exchange: Option<String>,
    limit: Option<i32>,
}
```

juhaku:
For these few recent days, I haven't had a time to do much regarding this project. I have had my hands full of other things explained a bit here: #276.

I can add this to the master, but I will not create a release before this feature has been rewritten with serde support #313 thus the rename will then be only temporarily within `param(rename = ...)` and will be removed when the support is written for `serde(rename = ...)`.

juhaku (approved):
This is pretty okay. 

Though it is missing docs and tests. But since this will be only a temporary functionality those can be omitted I suppose. The support should be implemented with `serde` as described here:
https://github.com/juhaku/utoipa/pull/309#issuecomment-1272371307 and here: #313

Issue #307, opened 2022-10-02
SchemaFormat should be an open value


jacob-pro:
According to the spec:
https://swagger.io/docs/specification/data-models/data-types/#string

> However, format is an open value, so you can use any formats, even not those defined by the OpenAPI Specification, such as:
    - email
    - uuid
    - uri
    - hostname
    - ipv4
    - ipv6
    - and others

I would like to be able to use some of these custom format using the `#[schema(format="")]` annotation, however SchemaFormat appears to currently limited to a [hardcoded enum](https://docs.rs/utoipa/latest/utoipa/openapi/schema/enum.SchemaFormat.html)?

juhaku:
Right, this true, currently it is limited to the hard coded enum, which does not have custom options. Adding possibility to use a custom format would not be a massive enhancement to do. The enum itself could have `Custom` option and parsing of `format` could be made conditional either literal string which would become `Custom` and otherwise any or hard coded values like now.
```rust
enum SchemaFormat {
    Custom(String),
}
```

jacob-pro:
My concern with this is maintaining de-serialization compatibility - because as far as I can see deserializing other values is not currently supported in serde itself: https://github.com/serde-rs/serde/issues/912

There is some discussion and work-arounds here: https://stackoverflow.com/questions/57469527/how-can-i-support-an-unknown-or-other-value-for-a-serde-enum

One way to do it is to write our own `From<String>` and use the `serde(from="String")` annotation - would that be ok with you?

jacob-pro:
same problem for serialization as well I think

juhaku:
Yeah, I think this needs an experiment. I believe there is a way which need to be tested though. Key point in example below is to define `untagged` to the enum representation.
```rust
#[derive(Serialize, Deserialize, Clone)]
#[cfg_attr(feature = "debug", derive(Debug))]
#[serde(rename_all = "lowercase", untagged)]
pub enum SchemaFormat {
    // ... other format options
    /// Custom schema format which is not one of the defaults.
    Custom(String),
}
```
If that does not avail perhaps we need to introduce another `enum` with the current `SchemaFormat`. This `Format` enum would be `untagged` instead of the `SchemaFormat` being untagged. The order of the variants within the `Format` is important since that is the visit order for visitor implementations generated by the serde. We want it first to try out to parse `SchemaFormat` and then on error just parse any string as `Custom`. 
```rust
#[derive(Serialize, Deserialize)
#[serde(untagged)]
[...]

PR #315, opened 2022-10-09
Allow custom SchemaFormats

jacob-pro:
As per the OpenAPI spec, allows custom values to be used as a format.

Closes https://github.com/juhaku/utoipa/issues/307


juhaku (approved):
:1st_place_medal: Awesome

Issue #314, opened 2022-10-07
Support for nullable fields


jacob-pro:
Currently it doesn't seem there is support for the [nullable](https://swagger.io/docs/specification/data-models/data-types/#null) attribute on a property.

Not required means the field may be omitted completely from the JSON. Whereas nullable means it can be present but have a null value. This distinction is useful for partial update APIs where there is a distinction between making no changes, versus modifying an optional field.

See [double_option](https://docs.rs/serde_with/latest/serde_with/rust/double_option/index.html) for more info on usage.

At a minimum we should be able to manually mark fields as nullable. Even better would be to detect double Options and mark them as non-required and nullable.

juhaku:
Nullable indeed is not supported since the Rust itself does not have null values. The absence is measured via `Option`. If a value is an `Option` type it will be marked as non-required. Alternatively currently if field value is decorated with `#[serde(default)]` it will be non-required as well. 

It might not be a too big of a stretch to support double `Option` and mark the field nullable. Or at least add derive parameter `nullable` for a field. However if double `Option` is considered a nullable it should not be done so automatically but instead behind a feature flag. Not all might want to enable such behavior. 

In Rust world I reckon the partial updates works really well without double `Option` as well. Values that are `None` can be easily handled in application so that these values are not updated to whatever that update target might be e.g. database. I know that there are lot of languages e.g. Java and TypeScript which are heavily in the world of `null` which is a curse word and a cancer within languages. And in these languages the `null` is used to represent absence. Exception in TypeScript where absence is measured with undefined and null value is a null value which is another annoying issue where to me  both undefined and null are used to "solve" same problem.

Thinking this further the serde itself will actually serialize all Option fields with `None` value as `null` in JSON by default. Only when field is decorated with `#[serde(skip_serializing_if = "Option::is_none")]` the field will not be present in the JSON. Perhaps current logic could also be enhanced so that if the field is decorated with serde's `skip_serializing_if` attribute the field will be just non-required field and otherwise the field would actually also be nullable by default.

jacob-pro:
Let's say you have an API that allows updating a user account, it is a partial update API, so only properties that are included will be changed, e.g.  `PUT /users/123`

The input `{ "phone_number": null }` would remove the phone number from the user account, whereas an input of `{ }` would simply make no changes to the phone number.

So as far as Serde is concerned:

```rust
pub struct OptionalVsNull {
    
    // When serialized it will be completed omitted from the JSON
[...]

PR #319, opened 2022-10-12
Allow manually setting field as nullable

jacob-pro:
This is a partial solution to https://github.com/juhaku/utoipa/issues/314

This adds support to the schema object for the `nullable` field, and allows manually setting it via the `nullable` attribute.

This does not yet automatically enable it when there are double Options / inspect the types and serde annotations etc. - that would be a separate task...


juhaku (approved):
Awesome, great work :+1: This paves the way eventually for more advanced cases e.g. automatic nullable resolvation.

Issue #312, opened 2022-10-05
Serializing and then Deserializing is incomplete


ju6ge:
Hey,

I have stumbled upon a problem. The yetserialization of "Schema" types fails. To showcase what I mean I have written the following test:

```rust
#![cfg(all(feature = "yaml", feature = "json"))]

use std::assert_eq;
use serde_json;
use serde_yaml;

use utoipa::openapi::{*, Schema, ObjectBuilder};
use utoipa::openapi::schema::RefOr;

#[test]
pub fn serialize_deserialize_schema() {

	let ref_or_schema = RefOr::T(
		Schema::Object(
			ObjectBuilder::new()
				.property("test", RefOr::T(Schema::Array(
				ArrayBuilder::new()
					.items(RefOr::T(
						Schema::Object(
							ObjectBuilder::new()
								.property("element", RefOr::Ref(Ref::new("#/test")))
								.build()
						)
					))
					.build()
			)))

				.build()
		));

	let yaml_str = serde_yaml::to_string(&ref_or_schema).expect("");
	println!("----------------------------");
	println!("{yaml_str}");

	let deserialized: RefOr<Schema> = serde_yaml::from_str(&yaml_str).expect("");

	let yaml_de_str = serde_yaml::to_string(&deserialized).expect("");
	println!("----------------------------");
	println!("{yaml_de_str}");

	assert_eq!(yaml_str, yaml_de_str);
}
```

This produces an assertion failed with the following std output:
`cargo test --features yaml --features json --test serde_schema -- --nocapture`

```
    Finished test [unoptimized + debuginfo] target(s) in 0.15s
     Running tests/serde_schema.rs (target/debug/deps/serde_schema-8fd331298e547711)

running 1 test
----------------------------
type: object
properties:
  test:
    type: array
    items:
      type: object
      properties:
        element:
          $ref: '#/test'

----------------------------
type: object
properties:
  test:
    type: array

thread 'serialize_deserialize_schema' panicked at 'assertion failed: `(left == right)`
  left: `"type: object\nproperties:\n  test:\n    type: array\n    items:\n      type: object\n      properties:\n        element:\n          $ref: '#/test'\n"`,
 right: `"type: object\nproperties:\n  test:\n    type: array\n"`', utoipa-gen/tests/serde_schema.rs:41:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test serialize_deserialize_schema ... FAILED

failures:

failures:
[...]

PR #323, opened 2022-10-17
Custom Deserializer for Schema Type

ju6ge:
This PR adds the tests described in #312 as well as implements a custom deserializer for the Schema Type to address the issue.

Thoughts and Improvements welcome ;)

### DIFF
--- a/utoipa/src/openapi/schema.rs
+++ b/utoipa/src/openapi/schema.rs
@@ -271,7 +271,10 @@ impl<'de> Deserialize<'de> for Schema {
                                     Ok(o) => Ok(Schema::AllOf(o)),
                                     Err(msg) => panic!("{}", msg)
                                 }
-                            } else {
+                            } else if s == "string" {
+                                Ok(Schema::Object(ObjectBuilder::new().schema_type(SchemaType::String).build()))
+                            }
+                            else {
                                 println!("{v:#?}");
                                 panic!("Can not deserialize unsupported schema type '{}'!", s)
                             }
--- a/utoipa/src/openapi/schema.rs
+++ b/utoipa/src/openapi/schema.rs
@@ -272,6 +272,7 @@ impl<'de> Deserialize<'de> for Schema {
                                     Err(msg) => panic!("{}", msg)
                                 }
                             } else {
+                                println!("{v:#?}");
                                 panic!("Can not deserialize unsupported schema type '{}'!", s)
                             }
                         }
--- a/utoipa/src/openapi/schema.rs
+++ b/utoipa/src/openapi/schema.rs
@@ -2,6 +2,7 @@
 //! used to define field properties, enum values, array or object types.
 //!
 //! [schema]: https://spec.openapis.org/oas/latest.html#schema-object
+use serde::Deserializer;
 use std::collections::{BTreeMap, HashMap};
 
 use serde::{Deserialize, Serialize};
@@ -213,7 +214,7 @@ impl ComponentsBuilder {
 ///
 /// [schemas]: https://spec.openapis.org/oas/latest.html#schema-object
 #[non_exhaustive]
-#[derive(Serialize, Deserialize, Clone)]
+#[derive(Serialize, Clone)]
 #[cfg_attr(feature = "debug", derive(Debug))]
 #[serde(untagged, rename_all = "camelCase")]
 pub enum Schema {
@@ -234,6 +235,60 @@ pub enum Schema {
     ///
     /// [composite]: https://spec.openapis.org/oas/latest.html#components-object
     AllOf(AllOf),
+}
+
+impl<'de> Deserialize<'de> for Schema {
+    fn deserialize<D>(deserializer: D) -> Result<Schema, D::Error>
+    where
+        D: Deserializer<'de>
+    {
+        let v: Value = serde::Deserialize::deserialize(deserializer)?;
+        match &v {
+            Value::Object(ob) => {
+                if ob.contains_key("type") {
+                    match ob.get("type").unwrap() {
+                        Value::String(s) => {
+                            if s == "object" {
+                                let ro: Result<Object, serde_json::Error> = serde::Deserialize::deserialize(v);
+                                match ro {
+                                    Ok(o) => Ok(Schema::Object(o)),
+                                    Err(msg) => panic!("{}", msg)
+                                }
+                            } else if s == "array" {
+                                let ro: Result<Array, serde_json::Error> = serde::Deserialize::deserialize(v);
+                                match ro {
+                                    Ok(o) => Ok(Schema::Array(o)),
+                                    Err(msg) => panic!("{}", msg)
+                                }
+                            } else if s == "oneof" {
+                                let ro: Result<OneOf, serde_json::Error> = serde::Deserialize::deserialize(v);
+                                match ro {
+                                    Ok(o) => Ok(Schema::OneOf(o)),
+                                    Err(msg) => panic!("{}", msg)
+                                }
+                            } else if s == "allof" {
+                                let ro: Result<AllOf, serde_json::Error> = serde::Deserialize::deserialize(v);
+                                match ro {
+                                    Ok(o) => Ok(Schema::AllOf(o)),
+                                    Err(msg) => panic!("{}", msg)
+                                }
+                            } else {
+                                panic!("Can not deserialize unsupported schema type '{}'!", s)
+                            }
+                        }
+                        _ => {
+                            panic!("Object `type` attribute needs to of type string!")
+                        }
+                    }
+                } else {
+                    panic!("Deserialized object has no key named `type`, can not derive Schema Type!")
+                }
+            }
+            _ => {
+                panic!("Deserializing Schema requires that the input is of type Object!")
+            }
+        }
+    }
 }
 
 impl Default for Schema {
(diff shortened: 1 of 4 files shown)


ju6ge:
For me this PR fixes Issue #312 without needing to change the serde tags or structure. But I could use some help/guidance on how to deal with the dependencies within the crate. Since openapi is build around json I don't think that `serde_json` should be an optional dependency. This just introduces unnecessary bloat and makes the code less readable. Implementing the custom deserialization function makes serde_json an requirement to build the code.

jayvdb:
Just noting that with this PR, I get errors like below, for models like:

```rust
#[derive(Deserialize, ToSchema)]
pub struct Image {
    pub id: Uuid,
    #[schema(example = "image/png")]
    pub mime_type: String,
    ...
}
```

```rust
error[E0277]: the trait bound `std::string::String: std::convert::From<serde_json::Value>` is not satisfied
   --> project/src/models/model_name.rs:6:23
    |
6   | #[derive(Deserialize, ToSchema)]
    |                       ^^^^^^^^ the trait `std::convert::From<serde_json::Value>` is not implemented for `std::string::String`
    |
    = help: the following other types implement trait `std::convert::From<T>`:
              <&'a str as std::convert::From<mime::Name<'a>>>
              <&'a str as std::convert::From<webpki::name::dns_name::DnsNameRef<'a>>>
              <&'t str as std::convert::From<regex::re_unicode::Match<'t>>>
              <std::string::String as std::convert::From<&mut str>>
              <std::string::String as std::convert::From<&std::string::String>>
              <std::string::String as std::convert::From<&str>>
              <std::string::String as std::convert::From<Box<str>>>
              <std::string::String as std::convert::From<Cow<'a, str>>>
            and 5 others
    = note: required because of the requirements on the impl of `Into<std::string::String>` for `serde_json::Value`
note: required by a bound in `ObjectBuilder::example`
   --> /home/jayvdb/.cargo/git/checkouts/utoipa-a72e072de3d00100/ed2a123/utoipa/src/openapi/schema.rs:916:23
    |
916 |     pub fn example<I: Into<String>>(mut self, example: Option<I>) -> Self {
    |                       ^^^^^^^^^^^^ required by this bound in `ObjectBuilder::example`
    = note: this error originates in the derive macro `ToSchema` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
```

ju6ge:
Yes the same test is also failing in the CI pipeline. I have not come around to fixing it yet. I think the problem lies within the way serde_json traits are optionally like so included: 
```rust
    /// Default value which is provided when user has not provided the input in Swagger UI.
    #[cfg(feature = "serde_json")]
    pub default: Option<Value>,

    /// Default value which is provided when user has not provided the input in Swagger UI.
    #[cfg(not(feature = "serde_json"))]
    pub default: Option<String>,
```
This also applies to `use` statements.

There must be a reason to be doing so? So I don't want to just remove it. But the question also is if I optionally include the fix does that mean, that the bug persists in versions where somebody explicitly removes `serde_json`. Which makes me wonder if there is any harm in making it a hard requirement?

utoipa/src/openapi/schema.rs
```diff
@@ -236,6 +236,59 @@ pub enum Schema {
...
+                                    Ok(o) => Ok(Schema::AllOf(o)),
+                                    Err(msg) => panic!("{}", msg)
+                                }
+                            } else {
+                                // per default it is tried to deserialize as an object
+                                let ro: Result<Object, serde_json::Error> = serde::Deserialize::deserialize(v);
+                                match ro {
+                                    Ok(o) => Ok(Schema::Object(o)),
+                                    Err(msg) => panic!("{}", msg)
+                                }
```
juhaku, line 274: For `array` and to the `object` this _type_ field check might work just fine. But for `oneOf` and for `allOf` there is no type field to check. Instead they just are formed like so:
```yaml
oneOf:
  - #/schemas/components/Foobar
---
allOf:
  - #/schemas/components/Foobar
```
Probably there is something more sophisticated that need to be done in order to cover these as well.

juhaku:
> There must be a reason to be doing so? So I don't want to just remove it. But the question also is if I optionally include the fix does that mean, that the bug persists in versions where somebody explicitly removes serde_json. Which makes me wonder if there is any harm in making it a hard requirement?

The original reason for having `serde_json` as optional dependency was to of course have as little dependencies as possible for users. Also to allows users to use the macros and types without the `serde_json` dependency for example in case where users could define the types in common crate which does not have a dependency to the `serde_json`. One could define the OpenAPI types in another crate without depending on the actual serialization logic but just the types. Then of course the exmaple field must be a `string` format. However when the types would be consumed in the web app crate you need to have a `serde_json` as a dependency in order to allow them to be serialized to JSON.

> This just introduces unnecessary bloat and makes the code less readable. Implementing the custom deserialization function makes serde_json an requirement to build the code.

This optional dependency indeed adds to the complexity of the code. This might be a valid point to have that consideration whether the `serde_json` should be lifted being an optional dependency. It should not cause any errors but for people that already might be using the library without default dependencies like described above in a crate which does not have dependencies to the `serde_json` this might be a problematic. If this is to be changed most likely safest choice would be to make a major release for people to have heads up that there might be something that breaks. Along with that then of course the unnecessary conditional `serde_json` feature checked code could be removed completely from `utoipa` making the types more clear indeed.

ju6ge:
Yes I see the benefit for people how would not like to use json. For me the question is from a practical perspective how much time should be invested on finding a different solution given that probably most people will be using the json part of it anyway, since basically everybody else supports json. If you use utoipa to deserialize an api spec you want json support and if you want to export a rust schema to openapi you will probably want to support every format you can. Unless you are operating in a very resource constrained environment and really need to save space.

The issue is basically that you need to "look ahead into the object tree", to be better able to know what type of object you are deserializing. This is done by first deserializing into a "Value" type (`serde_json::Value`). It also exists for other crates (`serde_yaml::Value`), maybe there is a better way to proxy this value type, without introducing more include guard hell. Or we refactor to make serde_json a requirement, which would probably only impact library size not functionality.

utoipa/src/openapi/schema.rs
ju6ge: Yes, but nothing to complicated. Maybe I will have some time tomorrow to take a look at this.

juhaku: That would be awesome. Thank you for your time investment with this library :slightly_smiling_face: I am also little by little getting back to the routine since the personal things are settled more or less.

juhaku:
Yes, I can do a quick test hopefully tomorrow regarding this how much it would affect. But most likely the impact is unnoticeable. If it only increses the library size a bit perhaps we can live it it. At least we can then remove quite a big part of code from `utoipa` crate which would make it a lot smaller in place. :smile:

ju6ge:
Well it is all about tradeoffs in the end :laughing:

utoipa/src/openapi/schema.rs
ju6ge: Hey so I have some changes that fix this now on my personal verison, will push it soon … but now I am a bit puzzled about the `OneOf` and `AllOf` structures of Utoipa. if `oneOf` or `allOf` are supposed to be serialized/deserialized to/from an array, that is not compatible with the other keys present in the Struct like `description` and so forth.

ju6ge:
So I fixed the issue of the `allOf` and `oneOf` schema types having a different format. In the process I think there are even more issues here. The `AllOf` and `OneOf` structures contain fields that can not be handled with the array type, and I would suspect that serializing these types also yields an incorrect result.

I have also written some tests for the deserialization routine. Testing for correctness is a bit of a pain, because the utoipa types do not implement `PartialEq`.  Adding support for `PartialEq` would make testing a lot easier!

juhaku:
Great, Yeah.. I have been thinking this on side as well. And since the `OneOf` and `AllOf` are actually schema objects like the `Object` but have those extra fields like `oneOf`. As they should support all the same attributes as the `Object` does I think this needs a bit different approach with the types in the end since now they do not support and to make them support there is a need to manually add more fields to the types which itself is kinda naah.. :smile: Need to experiment a bit with the types. 

We could add the `PartialEq` for those types though :thinking:

juhaku:
I created experiemental PR for this issue, I tried with few test cases and by just changing the order of `Object` and `Array` seemingly fixed the issue. :open_mouth:

ju6ge:
PartialEq would be great. That enables better testing capabilities. In order to be sure that the Schema Type can handle all desired cases, we should create a lot more tests using real world openapi definitions as input and validate, that they deserialize to the correct structure.

I am a bit busy with other stuff right now. But will take a look as soon as I have some more time.

juhaku:
Sure that's fine. I added couple of more test cases. But sure it would be good to have few real world schemas to test it with. Currently the schema is missing some of the possible configuration options so not any schema will actually work. But schema what are created with the utoipa should work indeed.

juhaku:
I'll first merge #356 since it seems to help with the symptoms at least. So I can create a release as well of currently available new things.

juhaku:
Just a reference here: now there is a PR for actually removing the conditional code from utoipa and simultaneously making `serde_json` a mandatory dependency. Anyway it is needed to serialize the OpenAPI spec correctly. https://github.com/juhaku/utoipa/pull/378

ju6ge:
Sorry I have not had the time to look at this further, because I had other things to do as part of my day job. Since nobody else is complaining it is probably fine. As soon as I rebase my fork back onto master I will let you now if everything went smoothly ;)

Issue #120, opened 2022-04-28
#[serde(flatten)] may not be supported (yet)


bhoudebert:
If we design a struct with the special instruction `flatten` of `serde` it is not flatten during schema creation.

```Rust
#[derive(Serialize, Deserialize, Debug, Clone, Component)]
pub struct SomePayload {
  some_id: String,
  
  #[serde(flatten)]
  extras: ExtraContent,
}
```

So would it be possible to tackle that point, maybe like the example, format, write_only?

```Rust
#[derive(Serialize, Deserialize, Debug, Clone, Component)]
pub struct SomePayload {
  some_id: String,

  #[serde(flatten)]
  #[component(nested = true)]
  extras: ExtraContent,
}
```

juhaku:
Yes `flatten` is not supported at least yet. To support this  there is actually couple of things could be done and there is no need to have `#[component(nested = true)]` necessarily. 

a) If component have `#[serde(flatten)]` attribute the component could be transformed to `AllOf` type which then contains the own schema inlined and the references to the other schemas. But then it means that the `ExtraContent` need to be `Component` as well. 

b) If component will have the `#[serde(flatten)]` then propably runtime ther should be custom modifier which does handle this nested mangling of types by copying the properties from nested type to the actual type etc. And similarly the `ExtraContent` should be a `Component` as well. 

I would prefer the `a)` since its simplier solution and works well with the compile time design. `b)` at the other hand needs some hacks and takes a toll with performing some schema mangling at runtime, but would not use `AllOf` composite type and all properties would be directly under the `SomePayload` itself.

bhoudebert:
I would be ok with anything solving this but sure `a)` detection would be really great and for the side note I already deal with `ExtraContent` as a `Component` which to me is not an issue.

PR #325, opened 2022-10-21
Add support for serde flatten

kw7oe:
Address issue: https://github.com/juhaku/utoipa/issues/120

It's still an early draft and probably missing impl on other applicable areas and test cases. Do let me know if I'm heading towards the right direction.

### DIFF
--- a/utoipa-gen/src/component/schema.rs
+++ b/utoipa-gen/src/component/schema.rs
@@ -230,52 +230,17 @@ impl ToTokens for NamedStructSchema<'_> {
                 let name = &rename_field(&container_rules, &mut field_rule, field_name)
                     .unwrap_or_else(|| String::from(field_name));
 
-                let type_tree = &mut TypeTree::from_type(&field.ty);
-
-                if let Some((generic_types, alias)) = self.generics.zip(self.alias) {
-                    generic_types
-                        .type_params()
-                        .enumerate()
-                        .for_each(|(index, generic)| {
-                            if let Some(generic_type) = type_tree.find_mut_by_ident(&generic.ident)
-                            {
-                                generic_type.update_path(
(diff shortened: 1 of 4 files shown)


juhaku:
Took a quick look and looks promising :+1:  :tada:  Indeed missing at least docs. Also eventually this need to be implemented to the `IntoParams` in `into_params.rs` but there is no serde support at all currently but that can be another story later on.

kw7oe:
Thanks for the review, I'll add another test case where we have nested flatten and also looking into updating the docs as well. For `IntoParams`, I can take a quick look into it and see if I can add the support as well.

kw7oe:
Hey @juhaku I have update the docs _(I could be missing something as well...)_, added some test cases and I think is ready to be review.

For the `IntoParams`, since I have less context on it since I haven't need to use it so far. While I would still like to tackle on this, would you mind to sharing more details to me? For example:
- is the goal to be supporting `param(flatten)`? 
- where should I be adding the test cases if I implemented it? I see a couple of files that might be related to this work like `path_derive*.rs` and `path_parameter_derive*.rs`. Having a pointer would be great hahaha

And I totally understand that you are occupied with personal things, so take all the time you need. Nothing is urgent here. I would just like to contribute some work back to upstream.

juhaku (approved):
Looks good to me :100:

juhaku:
> is the goal to be supporting param(flatten)?

Yes

> where should I be adding the test cases if I implemented it?

`utoipa-gen/tests/path_derive.rs` Seems like the place.. There already is some IntoParams tests. Essentially the functionality is the same. It `IntoParams` is used to create openapi request parameters from rust structs.

juhaku:
The `IntoParams` implementation can be a separate PR. I can merge this one to master.

PR #344, opened 2022-11-10
Use OUT_DIR in build.rs

dkellner:
From the docs [0]: "Build scripts may save any output files or intermediate artifacts in the directory specified in the OUT_DIR environment variable. Scripts should not modify any files outside of that directory."

[0] https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script

---

I've discovered this issue when I tried to build a crate depending on utoipa using Nix. In this case the current "fallback" directory ("../target") is not writable and executing `build.rs` fails.

### DIFF
--- a/utoipa-swagger-ui/build.rs
+++ b/utoipa-swagger-ui/build.rs
@@ -1,6 +1,6 @@
 use std::{
     cmp::Ordering,
-    env::{self, VarError},
+    env,
     fs::{self, File},
     io,
     path::PathBuf,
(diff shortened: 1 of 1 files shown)


utoipa-swagger-ui/build.rs
```diff
@@ -18,17 +18,7 @@ fn main() {
...
-        .or_else(|_| -> Result<String, VarError> {
-            let manifest_dir = env::var("CARGO_MANIFEST_DIR")?;
-            if manifest_dir.contains("/package") {
-                Ok("target".to_string())
-            } else {
-                Ok("../target".to_string())
-            }
-        })
-        .unwrap();
+    let target_dir = env::var("OUT_DIR").unwrap();
```
juhaku, line 21: The reason for logic behind is that the docs.rs uses custom target dir for the artifacts. And we cannot use `OUT_DIR` for the artifacts because it wont be constant place thus `rust-embed` will break (at least previously). This would need some throughout testing with actual API using dev build and release build to see whether it actually works. 

There is also branch: https://github.com/juhaku/utoipa/tree/test-swagger-build-attr for fixing  the target dir assumption many are facing. But not merged yet.

juhaku: :ok_hand: Good, two things I experienced before was. 
1. docs.rs has custom builder setup which relies on custon target dir location otherwise the build will completely fail there. 
2. When I used `OUT_DIR` the build out dir was not constant thus building once the app and then second time it did not produce output dir resulting empty pages  / or did not find anything to show via RustEmbed. This happened with older Rust versions though.

juhaku: I think this `OUT_DIR` can be tried out. But it need to be released alone as single change and first as a pre-release in case there are some funny outcomes in docs.rs etc. 

I don't have the error reports any more and docs.rs does not keep that long history of builds. And in my understanding the only place to find out incompatibilities is to browse the build logs of docs.rs.

juhaku:
I'll try with the OUT_DIR and see the results.

dkellner:
Thanks! Let's see if this simple solution is good enough. :crossed_fingers:

PR #345, opened 2022-11-11
Update axum to 0.6 RC

BBaoVanC:
Probably best not to merge this until 0.6 is a stable release, but review would be appreciated

## BREAKING

Because of this change in axum 0.6:

> breaking: The request /foo/ no longer matches /foo/*rest. If you want
to match /foo/ you have to add a route specifically for that ([#1086](https://github.com/tokio-rs/axum/pull/1086))

... that means that we have to manually specify a second route without `*tail` at the end. So I decided the most reasonable thing would be to make the user remove the `*tail` from the URL in `SwaggerUi::new()`. Then we can just register the two routes easily by appending `*tail` to the end of the second one.

### DIFF
--- a/utoipa-swagger-ui/src/axum.rs
+++ b/utoipa-swagger-ui/src/axum.rs
@@ -1,5 +1,6 @@
 #![cfg(feature = "axum")]
 
+use std::borrow::Cow;
 use std::sync::Arc;
 
 use axum::{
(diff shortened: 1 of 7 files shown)


utoipa-swagger-ui/src/axum.rs
```diff
@@ -38,19 +39,24 @@ where
...
+        let config = Arc::new(config);
 
         router.route(
-            swagger_ui.path.as_ref(),
-            routing::get(serve_swagger_ui).layer(Extension(Arc::new(config))),
+            &swagger_ui.path,
+            routing::get(serve_swagger_ui).layer(Extension(config.clone())),
+        ).route(
+            &format!("{}*tail", swagger_ui.path),
+            routing::get(serve_swagger_ui).layer(Extension(config.clone())),
```
juhaku, line 49: No need to `.clone()` here since it is the last usage, thus it can be moved.
```suggestion
            routing::get(serve_swagger_ui).layer(Extension(config)),
```

juhaku (approved):
Overall seems solid :+1: 

Yeah, I agree, I wouldn't merge this yet because `axum` hasn't released yet the `0.6` stable and most of users are still using the `0.5` version according to the crates statistics. I'd go with the version has most downloads in crates.

PR #350, opened 2022-11-17
Enable syn's extra-traits feature when using debug

### LINKED ISSUE #349: debug feature fails to build
This Cargo.toml fails to build with rust 1.65:

```
[package]
name = "schematest"
version = "0.1.0"
edition = "2021"

[dependencies]
utoipa = { version = "2", features = ["debug", "repr"] }
serde = { version = "~1.0", features = ["derive"] }
```

with errors such as "`syn::Path` cannot be formatted using `{:?}` because it doesn't implement `Debug`".
This is because the "extra-traits" feature is missing from syn.

I don't understand why, but adding tracing or tracing-attributes (which depends on syn with extra-traits in its dependencies) makes it work:

```
[package]
name = "schematest"
version = "0.1.0"
edition = "2021"

[dependencies]
# syn = { version = "1.0", features = ["full", "parsing", "printing", "visit", "visit-mut", "clone-impls", "extra-traits", "proc-macro"] }
utoipa = { version = "2", features = ["debug", "repr"] }
serde = { version = "~1.0", features = ["derive"] }
tracing-attributes = { version = "0.1.23" }
```

Adding syn with extra-traits to my dependencies doesn't make it work, I have to add it to build-dependencies instead:

```
[package]
name = "schematest"
version = "0.1.0"
edition = "2021"

[dependencies]
utoipa = { version = "2", features = ["debug", "repr"] }
serde = { version = "~1.0", features = ["derive"] }

[build-dependencies]
syn = { version = "1.0", features = ["full", "parsing", "printing", "visit", "visit-mut", "clone-impls", "extra-traits", "proc-macro"] }
```

nkovacs:
Fixes #349

### DIFF
--- a/utoipa-gen/Cargo.toml
+++ b/utoipa-gen/Cargo.toml
@@ -36,7 +36,7 @@ assert-json-diff = "2"
 time = { version = "0.3.11", features = ["serde-human-readable"] }
 
 [features]
-debug = []
+debug = ["syn/extra-traits"]
 actix_extras = ["regex", "lazy_static", "syn/extra-traits"]
 chrono = []
 json = []


juhaku (approved):
:+1: Somehow I remembered that this was there already, but might be regression from some commits down to history. Anyways good addition.

PR #370, opened 2022-11-25
Fix broken swagger-ui on axum 

simmsb:
The new trailing slash handling of axum broke the swagger ui page.

This PR fixes it by introducing handlers for `<doc page>/` and `<doc page>/*tail` and introducing a redirect from `<doc page>` to `<doc page>/`

I'm unsure if this is the best way though as I don't have much experience with axum.


juhaku (approved):
I believe this is feasible solution for this. I am not super expert on axum either but I see there is some internal changes on how they handle urls and especially tail maching in handlers.

PR #380, opened 2022-12-01
Fix the order of responses

RossyWhite:
Use BTreeMap for `responses` in `Components`to keep fixed order (like other fields).


juhaku (approved):
Awesome :+1:

PR #382, opened 2022-12-02
feat: make schema_type field pub on Object struct

kyle-mccarthy:
I've been playing around with generating a client for the frontend for code that uses Utopia; however, I have to use a fork since the `schema_type` property isn't public.

Would you consider making this field public? I think it could be useful for others as well.


juhaku (approved):
Well sure why not. :slightly_smiling_face: 

The reason for it being private is that when it is public users could basically break their generated OpenAPI types e.g. changing the type to something else what would not make sense in terms of OpenAPI. But that of course would totally in up to the user. And with that said since there are types like `Array`, `Object`, `OneOf` `AllOf` thus there is no practical reason for manually changing the type. However if it is deemed useful to be public we can surely make it so.

kyle-mccarthy:
Thank you, I appreciate it!

Issue #281, opened 2022-09-13
Add support for extra NaiveDate Chrono type


JasterV:
I'm trying to use this library along with the `chrono` and I use the `NaiveDate` on lots of places but it is not supported by `utoipa`
 
> Could not resolve reference: Could not resolve pointer: /components/schemas/NaiveDate does not exist in document

JasterV:
@juhaku Thanks for your fast reply :D We were able to solve the errors on the schemas by using the `value_type` as you are saying.

But now we have another problem, we use `NaiveDate` as a param attribute on an endpoint and we are getting the same error as before.

juhaku:
Anytime :slightly_smiling_face: If you are using `IntoParams` then you are able to use the same `value_type` attribute for param as well. See the docs here: https://docs.rs/utoipa/latest/utoipa/derive.IntoParams.html

JasterV:
Cool! @juhaku

This is my current solution:

```
#[derive(Deserialize, IntoParams, Debug)]
pub struct Params {
    #[param(allow_reserved, value_type = String, example = "2022-01-22")]
    pub claim_date: NaiveDate,
    #[param(example = "Victor")]
    pub name: String,
}
```

I would love to be able to add a description for each field but I can't find the way to do it. Is there a way I can support both having the NaiveDate & having a description? 
Also being able to specify the format of the String such as the Date would be awesome :smile:

juhaku:
Description is actually just a doc comment above the field. While it might not be most explicit but it should be described somewhere at the docs, but might be missing. Seems by a quick look the details regarding the description is missing from the docs. Description is only supported by primitive fields or inlined objects. `$ref` fields to other schemas does not have description support. 
```rust
#[derive(Deserialize, IntoParams, Debug)]
pub struct Params {
    /// This first line here at the doc comment will end up being description in the OpenAPI doc
    /// 
    /// More additional details to comment
    #[param(allow_reserved, value_type = String, example = "2022-01-22")]
    pub claim_date: NaiveDate,
    #[param(example = "Victor")]
    pub name: String,
}
```

JasterV:
@juhaku Perfect thank you so much! Then I think we can close this issue :smile: 

Pd: As said, could it be possible to specify the format of the `value_type`? Like `format = Date`. 
We've done it when defining components but it can't be done here

juhaku:
[...]

PR #404, opened 2022-12-09
Add support for chrono NaiveDate

juhaku:
Add support for chrono NaiveDate to be recognized as Date in OpenAPI spec.


juhaku:
Resolves #281

PR #436, opened 2023-01-14
Add `preserve_order` feature to preserve property order

### LINKED ISSUE #428: [Feature Request] Serialize component properties in order of fields in struct definition
In certain use-cases, the order of fields in the request body (listed in the documentation) matters and having them sorted alphabetically (reference: #220) may not always be the best choice. Some fields might have to included first because they have more value compared to other fields. 

## Example

If we were to extend the `Todo` struct in the `todo-actix` example ([source code here](https://github.com/juhaku/utoipa/blob/e50da565e3b6afd5f1c2febea738b3d6057171e9/examples/todo-actix/src/todo.rs#L31-L42)) with a few more fields like so:

```rust
pub(super) struct Todo {
    id: i32,
    value: String,
    checked: bool,
    reminder: Option<OffsetDateTime>,
    tags: Vec<String>,
    repeat: Option<Duration>,
}
```

In this case, the `value` is an important field as it contains what the to-do list entry is about, while the other fields are optional and need not be provided. But sorting them alphabetically would cause them to be listed in the order: `[checked, id, reminder, repeat, tags, value]` moving the most important field to the extreme end of the list. This is just one of those cases where documenting fields in order of struct definition would have been beneficial compared to alphabetically sorted order.

## Feature Description

It would be great if the fields/properties documented for a specific struct followed the order of fields in the struct definition. Just to clarify, my request does not involve the order in which the components are listed (in the generated documentation), but only the order in which the properties of each component are listed. This should still generate meaningful and deterministic diffs that @Sytten mentioned in #210.

Let me know what you feel about this and if you'd like this change to be made, I can open a PR for it. Someone at work requested for this and I quickly made a hacky solution that just works for us ([commit](https://github.com/SanchithHegde/utoipa/commit/4ca5604f94484c7a5a7852972c9b6e68f33a871e)). My change involves storing the fields in a `Vec` of 2-tuples and serializing as a map of field-schema pairs.

SanchithHegde:
As discussed in #428, this PR introduces a cargo feature named `preserve_order` to retain property order when serializing a component schema.

Closes #428.

### DIFF
--- a/utoipa/Cargo.toml
+++ b/utoipa/Cargo.toml
@@ -32,6 +32,7 @@ time = ["utoipa-gen/time"]
 smallvec = ["utoipa-gen/smallvec"]
 openapi_extensions = []
 repr = ["utoipa-gen/repr"]
+preserve_order = []
 
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
--- a/utoipa/src/openapi/schema.rs
+++ b/utoipa/src/openapi/schema.rs
@@ -508,6 +508,11 @@ impl From<AllOfBuilder> for RefOr<Schema> {
 
 component_from_builder!(AllOfBuilder);
 
+#[cfg(not(feature = "preserve_order"))]
+type ObjectPropertiesMap<K, V> = BTreeMap<K, V>;
+#[cfg(feature = "preserve_order")]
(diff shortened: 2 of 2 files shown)


juhaku (approved):
Nice, :+1:

PR #438, opened 2023-01-16
support serde `skip_serializing`

Xenfo:


juhaku (changes requested):
This is good, but there is a test missing, and docs as well.

Docs should be updated here: https://github.com/juhaku/utoipa/blob/master/utoipa-gen/src/lib.rs#L179

And here:
https://github.com/juhaku/utoipa/blob/master/utoipa-gen/src/lib.rs#L1569

I suppose this `skip_serializing` should work in both `IntoParams` and in `ToSchema`

Xenfo:
I couldn't find where the tests are for the original `skip` and I couldn't find anything for the original `skip` related to `IntoParams`. Can you point them out for me please?

juhaku:
Sure here is the example test for skip for `ToSchema`
https://github.com/juhaku/utoipa/blob/13a3aaec932369544f791e17b53f8153314758ce/utoipa-gen/tests/schema_derive_test.rs#L2645-L2682

Seems like the skip support is not implemented for `IntoParams`. I remembered wrong. So we can ignore it now. That can be another PR because it needs more changes for `IntoParams` than just updating docs and adding tests.

So just creating a test for for `skip_serializing` attribute is enough in `tests/schema_derive_test.rs`

juhaku:
Also instead of using `assert_value!` macro you can use `assert_json_eq!` as well.

juhaku (approved):

PR #450, opened 2023-01-24
Fix spelling

jayvdb:


utoipa-gen/src/openapi/info.rs
```diff
@@ -100,18 +100,18 @@ impl Parse for Info<'_> {
                     info.description =
                         Some(parse_utils::parse_next(input, || input.parse::<Str>())?)
                 }
-                "licence" => {
-                    let licence_stream;
-                    parenthesized!(licence_stream in input);
-                    info.license = Some(licence_stream.parse()?)
+                "license" => {
```
jayvdb, line 103: I believe this was a bug in https://github.com/juhaku/utoipa/pull/400 , but there hasnt been a release since then.

juhaku: Yes indeed, nice catch

juhaku (approved):
Awesome, :1st_place_medal: :100: This is super. Thanks for hard work fixing my broken english :smile: 

Priceless contribution.

Issue #451, opened 2023-01-24
#[get(..)] and #[utoipa::path(..)] fails without path =  


jayvdb:
With `post` and `put` and likely others, I can do

```rust
#[utoipa::path(
    request_body  = MyStruct,
    responses(
        (status = 201, description = "Done something")
    ),
    security(
        ("api_key" = [])
    ),
)]
#[post("/my_endpoint")]
pub async fn foobar(req: HttpRequest, body: Json<MyStruct>) -> HttpResponse { .. }
```

However the same with `get` fails.
```rust
#[utoipa::path(
    responses(
        (status = 200, description = "Done successfully", body = MyStruct)
    ),
    security(
        ("api_key" = [])
    ),
)]
#[get("/get_props")]
pub async fn get_props() -> MyStruct { .. }
```
results in
```rust
error[E0277]: the trait bound `MyStruct: Responder` is not satisfied
   --> src/foobar.rs:68:14
    |
67  | #[get("/get_props")]
    | -------------------- required by a bound introduced by this call
68  | pub async fn get_props() -> MyStruct {
    |              ^^^^^^^^^ the trait `Responder` is not implemented for `MyStruct`
    |
    = help: the following other types implement trait `Responder`:
              &'static [u8]
              &'static str
              &std::string::String
              (R, StatusCode)
              BytesMut
              Cow<'_, str>
              CustomizeResponder<T>
              Either<L, R>
            and 13 others
note: required by a bound in `actix_web::Resource::<T>::to`
   --> /home/jayvdb/.cargo/registry/src/github.com-1ecc6299db9ec823/actix-web-4.2.1/src/resource.rs:226:20
    |
226 |         F::Output: Responder + 'static,
    |                    ^^^^^^^^^ required by this bound in `actix_web::Resource::<T>::to`

For more information about this error, try `rustc --explain E0277`.
```

To get it working, I need to "copy" the `#[get]` info into `#[utoipa::path(..)`, like below.  This inconsistency feels quite odd.
```rust
#[utoipa::path(
    get,
    path = "/get_props",
    responses(
        (status = 200, description = "Done successfully", body = MyStruct)
    ),
    security(
        ("api_key" = [])
    ),
)]
pub async fn get_props() -> MyStruct { .. }
```

The `#[get("/get_props")]` can still be in there, but it is superfluous, at best duplicating the info, and maybe can even get out of sync which would be confusing.

[...]

PR #452, opened 2023-01-24
Fix doctests

jayvdb:
Related to https://github.com/juhaku/utoipa/issues/451

### DIFF
--- a/utoipa-gen/src/lib.rs
+++ b/utoipa-gen/src/lib.rs
@@ -915,6 +915,8 @@ pub fn derive_to_schema(input: TokenStream) -> TokenStream {
 ///
 /// /// Get Pet by id
 /// #[utoipa::path(
+///     get,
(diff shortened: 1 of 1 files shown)


juhaku:
Most likely unnecessary "fix" because such feature is only supported in actix-web with `actix_extras` feature flag. See [comment here in #451.](https://github.com/juhaku/utoipa/issues/451#issuecomment-1401793014)

jayvdb:
Ah.  So can we make these doctest only enabled when `actix_extras` feature flag is enabled?  I'll look into that.

juhaku:
> Ah. So can we make these doctest only enabled when actix_extras feature flag is enabled? 

Yes, Not sure whether it is possible to enable only certain doc tests via feature flags. That's why framework independent tests are executed separately in `scripts/tests.sh` file.

Issue #448, opened 2023-01-23
Support for unit type


Erisli:
When using ToSchema for structs that contains the following type for a member, Utoipa thinks this is a tuple type that was not supported even it is a unit type.

```
Option<Result<(), CustomError>>
```

biwecka:
Hi, I guess a first attempt in this direction was made by supporting unit structs (#392) but unfortunately this does not include the unit type `()`. I looked into the code and as far as I can tell support for the unit type cannot be implemented the same way support for unit structs was implemented because the `ToTokens` trait is imported from another library and can therefore not be implemented for the unit type.
But it would be possible to implement the `ToSchema` trait for `()` (similar to the suggestion here #332).
It would look something like this:
```rust
// utoipa/src/lib.rs

/// Implement [`ToSchema`] trait for unit type.
impl ToSchema for () {
    fn schema() -> openapi::RefOr<openapi::schema::Schema> {
        openapi::schema::ObjectBuilder::new()
            .nullable(true)
            .default(Some(serde_json::Value::Null))
            .example(Some(serde_json::Value::Null))
            .into()
    }
}
```
The only thing I'm unsure about is if the `.nullable(true)` is correct. Maybe someone can help with that.
@juhaku Would you accept a pull request containing this feature?

juhaku:
Sure PRs are welcome. We could implement `ToSchema` for `()` but I believe first we need to see how serde is actually going to serialize the unit `()` type. `.nullable(true)` allows value be `null` OpenAPI spec which is crusial in cases if  the json serialized value is expected to be null. Depending on if `()` would not serialize to `null` then `.nullable(true)` is not needed either. 

Another thing to confirm is how the `TypeTree` is currently behaving with the unit `()` type. And whether there is something that can be done to distinct it from other types.

biwecka:
I got the following code to compile with the changes made in my pull request draft.
```rust
#[derive(Serialize, Deserialize, Debug, Default, ToSchema)]
pub struct Data {
    pub unit_type   : (),
}

#[utoipa::path(
    get,
    path = "/root",
    responses (
        ( status = 200, description = "", body = Data)
    )
)]
pub fn root() -> () {}

#[derive(OpenApi)]
#[openapi(paths(self::root), components(schemas(Data)))]
struct ApiDoc;

fn main() {
    println!("{}", ApiDoc::openapi().to_pretty_json().unwrap());
}
```

The "components" part of the output looks like this:
```json
"components": {
[...]

PR #464, opened 2023-01-27
Add support for unit type `()`

biwecka:
Pull request for #448.

### DIFF
--- a/utoipa-gen/src/component/schema.rs
+++ b/utoipa-gen/src/component/schema.rs
@@ -1597,14 +1597,14 @@ impl ToTokens for SchemaProperty<'_> {
                     // TODO support for tuple types
                     ValueType::Tuple => {
                         // Detect unit type ()
-                        if type_tree.children == None {
-                            tokens.extend(quote!{
+                        if type_tree.children.is_none() {
+                            tokens.extend(quote! {
                                 utoipa::openapi::ObjectBuilder::new()
                                     .nullable(true)
                                     .default(Some(serde_json::Value::Null))
                             })
                         };
-                    },
+                    }
                 }
             }
         }
--- a/utoipa-gen/src/component/schema.rs
+++ b/utoipa-gen/src/component/schema.rs
@@ -1595,7 +1595,19 @@ impl ToTokens for SchemaProperty<'_> {
                         }
                     }
                     // TODO support for tuple types
-                    ValueType::Tuple => (),
+                    ValueType::Tuple => {
+                        // Detect unit type ()
+                        if type_tree.children == None {
+                            tokens.extend(quote!{
+                                utoipa::openapi::ObjectBuilder::new()
+                                    .nullable(true)
+                                    .default(Some(serde_json::Value::Null))
+                                    .example(Some(serde_json::Value::Null))
+                            })
+                        }
+
+                        ()
+                    },
                 }
             }
         }
--- a/utoipa-gen/src/component/schema.rs
+++ b/utoipa-gen/src/component/schema.rs
@@ -1603,9 +1603,7 @@ impl ToTokens for SchemaProperty<'_> {
                                     .nullable(true)
                                     .default(Some(serde_json::Value::Null))
                             })
-                        }
-
-                        ()
+                        };
                     },
                 }
             }
--- a/utoipa-gen/src/component/schema.rs
+++ b/utoipa-gen/src/component/schema.rs
@@ -1602,7 +1602,6 @@ impl ToTokens for SchemaProperty<'_> {
                                 utoipa::openapi::ObjectBuilder::new()
                                     .nullable(true)
                                     .default(Some(serde_json::Value::Null))
-                                    .example(Some(serde_json::Value::Null))
                             })
                         }
 
--- a/utoipa-gen/tests/request_body_derive_test.rs
+++ b/utoipa-gen/tests/request_body_derive_test.rs
@@ -433,6 +433,42 @@ fn derive_request_body_ref_path_success() {
 }
 
 #[test]
+fn unit_type_request_body() {
+    type UT = ();
+
+    #[utoipa::path(
+        post,
+        path = "/unit_type_test",
+        request_body = UT
+    )]
+    #[allow(unused)]
+    fn unit_type_test() {}
+
+    #[derive(OpenApi)]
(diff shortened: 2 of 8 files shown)


utoipa-gen/src/component/schema.rs
```diff
@@ -1595,7 +1595,18 @@ impl ToTokens for SchemaProperty<'_> {
...
+                        // Detect unit type ()
+                        if type_tree.children == None {
+                            tokens.extend(quote!{
+                                utoipa::openapi::ObjectBuilder::new()
+                                    .nullable(true)
+                                    .default(Some(serde_json::Value::Null))
+                            })
+                        }
+
+                        ()
```
juhaku, line 1608: I believe it is enough to add semicolon here at the end of the if.
```suggestion
                        };
```

juhaku (commented):
Pretty good. Then there is 2 places where the `ValueType::Tuple` is being used. You might want to update them as well. 

* into_params.rs
* media_type.rs

Also add a test cases for these in utoipa-gen/test/schema_derive_test.rs and for IntoParams you can add test somewhere where IntoParams tests have been added before.

Run clippy and cargo fmt for the project as well. :)

biwecka:
What exactly do you mean with `IntoParams` tests? Should I simply use `#[derive(ToSchema)]` on a `struct`?

juhaku:
> What exactly do you mean with `IntoParams` tests? Should I simply use `#[derive(ToSchema)]` on a `struct`?

For `ToSchema` you can find tests here: https://github.com/juhaku/utoipa/blob/master/utoipa-gen/tests/schema_derive_test.rs

But for `IntoParams` you can find tests for example here: https://github.com/juhaku/utoipa/blob/master/utoipa-gen/tests/path_derive.rs

IntoParams is another struct what is used to parse path parameters for the operation.

biwecka:
Thank you for your help! I applied your code suggestion, updated the other two places where `ValueType::Tuple` is used, implemented tests and made sure clippy and fmt don't complain about my code. Unfortunately I'm still not too sure about the tests. So please review them and tell me if that's not what you expected.

utoipa/src/lib.rs
```diff
@@ -384,6 +384,18 @@ pub trait ToSchema<'__s> {
...
+        (
+            "UnitType",
+            openapi::schema::ObjectBuilder::new()
+                .nullable(true)
+                .default(Some(serde_json::Value::Null))
+                .into(),
+        )
+    }
+}
+
```
juhaku, line 398: I belive this is unnecessary since the implementation is handled in utoipa-gen where the tuple type with zero arguments is being tokenized.

utoipa-gen/tests/schema_derive_test.rs
```diff
@@ -3533,6 +3533,30 @@ fn derive_schema_with_custom_field_with_schema() {
...
+                "unit_type": {
+                    "type": "object",
+                    "default": null,
+                    "nullable": true
+                }
+            }
+        })
+    )
+}
+
```
juhaku, line 3559: This test is okay :+1: This is exactly what should be tested. Here we create `ToSchema` derived struct with unit type and see whether it creates correct json when tokenized with ToTokens in `component/schema.rs` file. 

And when introduced breaking changes these tests then will make sure that the implementation always stays correct, because after all the changes are done, the tests need to pass the build.

utoipa-gen/tests/request_body_derive_test.rs
```diff
@@ -432,6 +432,42 @@ fn derive_request_body_ref_path_success() {
...
 }
 
+#[test]
+fn unit_type_request_body() {
+    type UT = ();
+
+    #[utoipa::path(
+        post,
+        path = "/unit_type_test",
+        request_body = UT
```
juhaku, line 442: Just wondering whether you are able now to directly use the `request_body = ()` because the change you made to `media_type_schema.rs` file for the tuple

juhaku:
Here is a example test for `IntoParams` usage. `IntoParams` will derive from struct argument path operation parameters.
You could create another test like this but to use `IntoParams` as follows. It should work because the change you did in the `to_tokens` in `into_params.rs` and  to the `TypeTree`
```rust
#[derive(IntoParams)]
struct Params {
    value: (),
}
```
https://github.com/juhaku/utoipa/blob/96acebf090a49bb78283478d3effbdb1d9e208a8/utoipa-gen/tests/path_derive.rs#L1186-L1225

utoipa/src/lib.rs
biwecka: Adding this originates in me trying to use `()` in a generic like this:
```rust
type UT = ();
#[derive(ToSchema)]
#[aliases(StrData = Data<UT>)]
pub struct Data<T> {
    #[allow(unused)]
    value: T
}

#[utoipa::path(
    get,
    path = "/root",
    responses (
        ( status = 200, description = "", body = StrData )
    )
)]
pub fn root() -> () {}
```

It's not possible to use `()` in aliases like this `#[aliases(StrData = Data<()>)]`. Therefore I created the type alias `type UT = ();` and used it in `#[aliases(...)]`. With this change the `"value"` property is a reference with `"$ref": "#/components/schemas/UT"`. And for this reference to resolve I need to add `UT` to the schemas `components(schemas(StrData, UT))`.

utoipa-gen/tests/request_body_derive_test.rs
biwecka: I just checked and it's not possible. I get the following error message: "ValueType::Primitive must have path".

biwecka:
Thank's again for your help! It's still hard for me to understand how everything works together in utoipa. As you suggested I created a test very similar to the one you mentioned and it passes as expected 👍

utoipa/src/lib.rs
juhaku: Okay

utoipa-gen/tests/request_body_derive_test.rs
juhaku: Okay. Hm.. then there is something else missing as well.. Though I can check this later as well. It should not be primitive type but `ValueType::Tuple`

juhaku:
Sorry to keep you await, merged it finally

biwecka:
No problem, thank you very much 👍

PR #470, opened 2023-01-31
render enum variant examples

shelbyd:
My project needs to demonstrate examples for enum variants. The docs seem to indicate that this is supported, but it is not working for us. https://docs.rs/utoipa/latest/utoipa/derive.ToSchema.html

This is a branch against 2.4.2 that serves our purposes, but I figure you can use this as a base to incorporate into master.


juhaku:
Yes, thank you :rocket: I'll check this out and continue from this to get in on master eventually.

juhaku:
Closing due #538 that implements the same logic but rebased over master.

Issue #471, opened 2023-02-02
Allow const static str's to be used as operation_id


djrenren:
**TL;DR** It would be useful if the `operation_id` option on the `#[path]` macro could be a const variable instead of just a static string like so:

```rust
const MY_ID: &str = "some_id";

#[utoipa::path(
    ...
    operation_id = MY_ID
)]
fn foo(...) { ... }
```

**Why**
I'm currently writing an API that runs on an AWS Lambda behind API Gateway. There's this really nice feature of API Gateways where you give it an openapi spec and when it invokes your lambda, it tells you the operation id that corresponds to this request. So you can easily write a router like so:

```rust

match request_context.operation_name.as_deref() {
   route::MY_ID => foo(...)
}
```

Currently, I have to keep the string id in sync between my const definition and the `#[path]` macro invocation. It would be nice to just have a single source of truth.

Is this something you'd be open to a PR for?

juhaku:
Sure any contributions are welcome :slightly_smiling_face:

PR #472, opened 2023-02-02
Support arbitrary exprs in operation_id

djrenren:
Fixes https://github.com/juhaku/utoipa/issues/471

### DIFF
--- a/utoipa-gen/src/component/serde.rs
+++ b/utoipa-gen/src/component/serde.rs
@@ -37,7 +37,9 @@ impl SerdeValue {
             let mut rest = *cursor;
             while let Some((tt, next)) = rest.token_tree() {
                 match tt {
-                    TokenTree::Ident(ident) if ident == "skip" || ident == "skip_serializing" => value.skip = true,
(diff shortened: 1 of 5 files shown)


juhaku (changes requested):
Seems good, one thing though is to add a test for a path operation to test whether it really uses value of the const as operation id. 

You can see some help for creating  test here https://github.com/juhaku/utoipa/blob/96acebf090a49bb78283478d3effbdb1d9e208a8/utoipa-gen/tests/path_derive.rs#L1186-L1225

You can basically create a similar test but ignore all the parameter releated things etc and only test the path attribute macro and it's generated output including the operation id.

`dbg!(..)` macro is handy in examining path operation content.

djrenren:
Added one! Rather than doing a simple const identifier I went for a full expression since I figured it demonstrates the full power more clearly

juhaku:
Awesome! :+1:  One more thing, let's update the docs for the `operation_id`.
 https://github.com/juhaku/utoipa/blob/96acebf090a49bb78283478d3effbdb1d9e208a8/utoipa-gen/src/lib.rs#L621

juhaku (approved):
Perfect :100:

PR #488, opened 2023-02-18
Update readme with Graphul example

SamuelBonilla:
https://github.com/graphul-rs/graphul/tree/main/examples/utoipa-swagger-ui


juhaku (approved):
Very nice :+1:

Issue #527, opened 2023-03-17
SwaggerUI is not working in Docker, but works locally. Only the SwaggerUI throws a `404 Not Found`


kerkmann:
Okay, I know this sounds weird ... but: The SwaggerUI is not working inside of Docker. ^^"

Let me explain what I did. When I write down everything like in the example for `axum` and I start the application with `cargo run`, I can see the `swagger-ui` and also see the `/docs/openapi.json` file.
Now the interesting part, when I compile the code inside docker and provide it inside docker only the `swagger-ui` endpoint (`/docs` in this example) is returning a `404 Not Found`. But the interesting thing is, all other endpoints are working... Including the `/docs/opanapi.json` endpoint.
So, only the `swagger-ui` isn't working inside Docker. ^^"

I thought "maybe it's a compiled problem", so I copied the locally build binary (which worked on the machine and showed the swagger-ui) into the docker container, started it and "boom", no swagger-ui. ^^"
It's so weird that every endpoint is working, also the `openapi.json` endpoint (which shows the full generated openapi json), but only the `swagger-ui` HTML-Page is not working. There is really no difference inside the Docker container and my local machine (like I said, I copied the binary to prove my assumption).
I also tried bigger OS containers like `ubuntu` and smaller ones like `alpine`, in all containers every endpoint is working, except the `swagger-ui`. :(

Does someone have an idea what I could test/think of? I'm getting out of ideas. ^^"

kerkmann:
Okay, that explains a lot .... you are totally right, it's working with a release build. :)

I created the build in debug mode, and it seems like that the `debug-embed` flag is not enabled inside Docker. ^^"

Maybe it makes sense that we pinpoint that "issue" out in the Readme, thank you very much! <3

kerkmann:
I've created a small documentation change to pinpoint it better in the README. Maybe it will help out. :3

juhaku:
> Maybe it makes sense that we pinpoint that "issue" out in the Readme, thank you very much! <3

Yeah, actually there is slight hint towards this in utoipa-swagger-ui README.md but it is not obivous to the eye.

kerkmann:
> Yeah, actually there is slight hint towards this in utoipa-swagger-ui README.md but it is not obivous to the eye.

Ah yeah, now I see what you mean. It wasn't quite obvious to me and two other people. So spending hours investigating that issue could be solved by pinpointing it out. :)

juhaku:
[...]

PR #528, opened 2023-03-17
Add General Pitfalls with Docker into README

kerkmann:
This is related to the issue #527 :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,6 +280,12 @@ This would produce api doc something similar to:
 - More about OpenAPI security in [security documentation](https://docs.rs/utoipa/latest/utoipa/openapi/security/index.html).
 - Dump generated API doc to file at build time. See [issue 214 comment](https://github.com/juhaku/utoipa/issues/214#issuecomment-1179589373).
 
+## General Pitfalls
+
+### SwaggerUI not working inside Docker
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -280,6 +280,12 @@ This would produce api doc something similar to:
 - More about OpenAPI security in [security documentation](https://docs.rs/utoipa/latest/utoipa/openapi/security/index.html).
 - Dump generated API doc to file at build time. See [issue 214 comment](https://github.com/juhaku/utoipa/issues/214#issuecomment-1179589373).
 
+## General Pitfalls
+
+### SwaggerUI not working inside Docker
+
+When running your Rust application with SwaggerUI inside a Docker container, it's only possible to use a release build. This is because the [debug-embed](https://github.com/pyrossh/rust-embed#features) flag is not enabled in debug builds by default, which can cause issues with embedding necessary files into the binary. Therefore, make sure to use the `--release` flag when building your Rust application for use inside Docker.
+
```
juhaku, line 288: Actually it should not be completely necesary to build it with `--release` flag, but alternatively you can add a cargo `debug-embed` feature flag to the `utoipa-swagger-ui` dependency seen here: https://github.com/juhaku/utoipa/tree/master/utoipa-swagger-ui#crate-features. That will enable the RustEmbed's `debug-embed` feature which allows debug builds to embed files as well.

juhaku: I think there could also be a link to the `utoipa-swagger-ui`'s REAMDE.md as well

kerkmann: Hmm ... maybe you are right. ^^"

I would at least describe it more, especially with the wording Docker. Because I think some other people will fall into the same kind of problem. ^^"

PR #533, opened 2023-03-20
Add support for auto-populating field default values

### LINKED ISSUE #519: Auto-populate default values in `ToSchema` derive macro
It would be great to be able to automatically populate the default values for fields of a struct that implements `Default`.
So instead of:

```rust
#[derive(ToSchema)]
struct Foo {
    #[schema(default = 12)]
    a: u64,
    #[schema(default = "Hello")]
    b: String,
}
```

We could use:

```rust
#[derive(ToSchema)]
struct Foo {
    a: u64,
    b: String,
}

impl Default for Foo {
    fn default() -> Self {
        Self { a: 12, b: "Hello" }
    }
}
```
(Maybe with the addition of  a`#[schema(default)]` attribute or similar on the struct.)

Which would add `.default(Some(::serde_json::to_value(#struct_name::default().#field_name).unwrap()))` on every property of the generated schema.

I am using utoipa to generate schemas for structs with many default values set in `impl Default`, so this would be very useful to prevent repetition.

cwatson-blackrock:
Closes #519

Thanks again for your feedback on the issue! Here is an attempt at solving it.

## Summary
A named struct with `#[schema(default="")]` or `#[serde(default)]` attribute will use `<struct_name>::default().<field_name>` as the default value for all of its fields.
Any existing `#[default]` attribute on a field takes precedence.

## Caveats / to do
- I couldn't figure out how to allow `#[schema(default)]`, as the attribute parser expects an equals sign and a value. Any ideas would be welcome!
- I added a variant to `AnyValue` that represents this way of setting a default value -- I am not sure if this is very clean / consistent with how `AnyValue` is used currently. Please let me know if there is a better approach (same goes for naming and organization of course).

### DIFF
--- a/utoipa-gen/src/component/schema.rs
+++ b/utoipa-gen/src/component/schema.rs
@@ -298,10 +298,7 @@ impl NamedStructSchema<'_> {
                 let field_ident = field.ident.as_ref().unwrap().to_owned();
                 let struct_ident = format_ident!("{}", &self.struct_name);
                 features_inner.push(Feature::Default(
-                    crate::features::Default::new_default_trait(
-                        struct_ident,
-                        field_ident,
-                    ),
+                    crate::features::Default::new_default_trait(struct_ident, field_ident.into()),
                 ));
             }
         }
@@ -530,6 +527,20 @@ impl ToTokens for UnnamedStructSchema<'_> {
                     .as_ref()
                     .map(|override_type| matches!(override_type.value_type, ValueType::Object))
                     .unwrap_or_default();
+            }
+
+            if fields_len == 1 {
+                if let Some(ref mut features) = unnamed_struct_features {
+                    if pop_feature!(features => Feature::Default(crate::features::Default(None)))
+                        .is_some()
+                    {
+                        let struct_ident = format_ident!("{}", &self.struct_name);
+                        let index: syn::Index = 0.into();
+                        features.push(Feature::Default(
+                            crate::features::Default::new_default_trait(struct_ident, index.into()),
+                        ));
+                    }
+                }
             }
 
             tokens.extend(
--- a/utoipa-gen/src/component/schema.rs
+++ b/utoipa-gen/src/component/schema.rs
@@ -290,17 +290,19 @@ impl NamedStructSchema<'_> {
             .unwrap_or(false);
 
         if schema_default || serde_default {
-            let field_ident = field.ident.as_ref().unwrap().to_owned();
-            let struct_ident = format_ident!("{}", &self.struct_name);
-            let default_feature = Feature::Default(
-                crate::component::features::Default::new_default_trait(struct_ident, field_ident),
-            );
-            let features_inner = field_features.get_or_insert(vec![default_feature.clone()]);
+            let features_inner = field_features.get_or_insert(vec![]);
             if !features_inner
                 .iter()
                 .any(|f| matches!(f, Feature::Default(_)))
             {
-                features_inner.push(default_feature);
+                let field_ident = field.ident.as_ref().unwrap().to_owned();
+                let struct_ident = format_ident!("{}", &self.struct_name);
+                features_inner.push(Feature::Default(
+                    crate::features::Default::new_default_trait(
+                        struct_ident,
+                        field_ident,
+                    ),
+                ));
             }
         }
 
--- a/utoipa-gen/src/component/schema.rs
+++ b/utoipa-gen/src/component/schema.rs
@@ -2,7 +2,7 @@ use std::borrow::Cow;
 
 use proc_macro2::{Ident, Span, TokenStream};
 use proc_macro_error::{abort, ResultExt};
-use quote::{quote, ToTokens};
+use quote::{format_ident, quote, ToTokens};
 use syn::{
     parse::Parse, punctuated::Punctuated, token::Comma, Attribute, Data, Field, Fields,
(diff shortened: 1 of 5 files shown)


utoipa-gen/src/component/schema.rs
```diff
@@ -278,6 +278,30 @@ impl NamedStructSchema<'_> {
...
             .into_inner();
 
+        let schema_default = self
+            .features
+            .as_ref()
+            .map(|features| features.iter().any(|f| matches!(f, Feature::Default(_))))
+            .unwrap_or(false);
+        let serde_default = serde::parse_container(self.attributes)
+            .map(|rules| rules.default)
+            .unwrap_or(false);
```
juhaku, line 288: We probably could pass the container args to the function so we do not need to parse it again here. From this line: https://github.com/juhaku/utoipa/pull/533/files#diff-f44986f2c073c8469bf9086f7d719014eacc7ae5dee63186c941cec5c8dd5e22L340

juhaku (commented):
This looks pretty okay, :trophy: 

> I couldn't figure out how to allow `#[schema(default)]`, as the attribute parser expects an equals sign and a value. Any ideas would be welcome!

To be able to parse the `#[schema(default)]` without equals token we need to change how the parsing is being done for the `Default` feature. Instead of directly parsing the next token like here: https://github.com/juhaku/utoipa/pull/533/files#diff-ae9861d2ee34546b6f79f8278e17358b5345b19bf73da445e56eab12180068c4L421 it could be changed like this:
```rust
if input.peek(Token![=]) {
  parse_utils::parse_next(input, || AnyValue::parse_any(input)).map(|v| Some(Self(v)))
} else {
  Self(None)
}
```
And we could change the `Default` to accept `None` like `struct Default(Option<AnyValue>);` Or we could change any value to accept `None`.

utoipa-gen/src/component/schema.rs
```diff
@@ -278,6 +278,30 @@ impl NamedStructSchema<'_> {
...
+            let default_feature = Feature::Default(
+                crate::component::features::Default::new_default_trait(struct_ident, field_ident),
+            );
+            if let Some(ref mut features) = field_features {
+                if !features.iter().any(|f| matches!(f, Feature::Default(_))) {
+                    features.push(default_feature);
+                }
+            } else {
+                field_features = Some(vec![default_feature])
+            }
```
juhaku, line 302: I believe this can be a little bit cleaner with something like this:
```suggestion
            let features_inner = field_features.get_or_insert(vec![default_feature]);
            if !features_inner.iter().any(|f| matches!(f, Feature::Default(_))) {
                f.push(default_feature);
            }
```

juhaku: Might actually need a clone in this case because `default_feature` would be moved.

cwatson-blackrock:
Excellent suggestions, thank you. I've cleaned things up and it seems to work as intended now.
Also I forgot to ask before: can / should this also be implemented for tuple structs? I'm not sure if this would play well with the representation as arrays in the schema.

utoipa-gen/src/component/schema.rs
```diff
@@ -283,22 +284,23 @@ impl NamedStructSchema<'_> {
...
             .map(|rules| rules.default)
             .unwrap_or(false);
 
         if schema_default || serde_default {
             let field_ident = field.ident.as_ref().unwrap().to_owned();
-            let struct_ident: Ident = syn::parse_str(&self.struct_name).unwrap();
+            let struct_ident = format_ident!("{}", &self.struct_name);
             let default_feature = Feature::Default(
                 crate::component::features::Default::new_default_trait(struct_ident, field_ident),
             );
```
juhaku: Also one thing came to my mind as well. If we make this `default_feature` a closure then we can get rid of the annoying `clone()` below.
```rust
let default_feature = || Feature::Default(...)
```

juhaku:
> can / should this also be implemented for tuple structs? I'm not sure if this would play well with the representation as arrays in the schema.
I belive it is a good idea to implement them as well. Yeah with arrays it might not work very well, but with new type pattern it should be doable. The new type pattern would create schema for the inner type.
```rust
struct MyWrapper(Foobar);

impl Default for MyWrapper {
 ...
}
```
I think we can only are this scenario here within `UnnamedFieldStruct`
https://github.com/juhaku/utoipa/blob/323b155039a87f1c63a9c1b221508b1c56e5b475/utoipa-gen/src/component/schema.rs#L489-L514

utoipa-gen/src/component/schema.rs
cwatson-blackrock: That actually also causes a move ("closure cannot be invoked more than once because it moves the variable `struct_ident` out of its environment"). But to avoid the clone I just realized I can also create an empty vector since the feature would get added right after:
```rust
let features_inner = field_features.get_or_insert(vec![]);
if !features_inner
    .iter()
    .any(|f| matches!(f, Feature::Default(_)))
{
     features_inner.push(default_feature);
}
```
I also cleaned things up a bit since `default_feature` no longer needs to be used twice.

juhaku (approved):
Super :1st_place_medal: Let's get the ball rolling :slightly_smiling_face:

juhaku:
> Thanks. I implemented the same behavior as for named structs, for the case of a tuple struct with one field. Hopefully that covers all use cases of this feature

I belive that will be good for now, until someone wants more usecases supported :laughing:

PR #536, opened 2023-03-22
bugfix: use `map()` instead of `unwrap()`

### LINKED ISSUE #535: unable to use `rustfmt::skip` attribute with `ToSchema`
I have an enum called `Country` ( specifically all the country codes ) which derives `ToSchema`. Since this enum can have many variants, I would like to skip the default formatting of rustfmt. But the `ToSchema` derive macro panics on an `unwrap` that is used internally somewhere. 
```rust
use utoipa::ToSchema;

#[derive(ToSchema)]
#[rustfmt::skip]
pub enum Country {
    AL, DZ, AS, AO, AG, AR, AU, AT, AZ, BH, BY, BE, BR, BG, CA, CL, CO, HR, CZ, DK, DO, EG,
    EE, FI, FR, DE, GR, HK, HU, IN, ID, IE, IL, IT, JP, JO, KZ, KE, KW, LV, LB, LT, LU, MY,
    MX, NL, NZ, NO, OM, PK, PA, PE, PH, PL, PT, QA, RO, RU, SA, SG, SK, ZA, ES, LK, SE, CH,
    TW, TH, TR, UA, AE, GB, UY, VN, CN, MO, AM, CY, FO, GE, GL, GG, IS, IM, JE, LI, MT,
    MD, MC, ME, SM, RS, SI, CR, PS, UM,
    #[default]
    US
}
```

The error being generated is
```rust
error: proc-macro derive panicked
 --> src/main.rs:3:10
  |
3 | #[derive(ToSchema)]
  |          ^^^^^^^^
  |
  = help: message: called `Option::unwrap()` on a `None` value
```

Narayanbhat166:
Closes #535 

This PR replaces few of the `unwraps()` used in the codebase to `map()` and then call `unwrap_or()` on those fields to provide a default value. This change is required because when using `rustfmt::skip` `get_ident()` function returns `None`, calling `unwrap()` on this value results in `panic`.

### DIFF
--- a/utoipa-gen/src/component.rs
+++ b/utoipa-gen/src/component.rs
@@ -39,7 +39,7 @@ fn is_default(container_rules: &Option<&SerdeContainer>, field_rule: &Option<&Se
 /// or field attributes of struct.
 fn get_deprecated(attributes: &[Attribute]) -> Option<Deprecated> {
     attributes.iter().find_map(|attribute| {
-        if *attribute.path.get_ident().unwrap() == "deprecated" {
+        if attribute.path.get_ident().map(|ident| *ident == "deprecated").unwrap_or(false) {
             Some(Deprecated::True)
         } else {
             None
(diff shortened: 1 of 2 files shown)


juhaku (approved):
Awesome, thanks for the patch :tada:

PR #568, opened 2023-04-10
Allow value_type serde_json::Value

jayvdb:
c.f. https://github.com/juhaku/utoipa/discussions/567

### DIFF
--- a/examples/raw-json-actix/Cargo.toml
+++ b/examples/raw-json-actix/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "raw-json-actix"
+description = "Simple actix-web using raw JSON with utoipa and Swagger"
+version = "0.1.0"
+edition = "2021"
+license = "MIT"
+authors = [
+  "Example <example@example.com>"
+]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+actix-web = "4"
+env_logger = "0.10.0"
+serde_json = "1.0"
+utoipa = { path = "../../utoipa", features = ["actix_extras"] }
+utoipa-swagger-ui = { path = "../../utoipa-swagger-ui", features = ["actix-web"] }
+
+[workspace]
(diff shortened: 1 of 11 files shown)


juhaku:
I am wondering how does the any type is different to the virtual `Object` type? Moreover from the perspective of the client side code generation?

Originally the `Object` type standed for the purpose of rendering any json content that does not have a type. But sure the any type support could also be added.

Seems like the array representation of `type:` field is not valid OpenAPI specification.
image

I believe the any type should rather be rendered as presented here: https://swagger.io/docs/specification/data-models/data-types/. Look for `Any Type` bottom of the page.

Just wondering about the name `Value` that is it too generic? While it would match well with serde's `Value` it simultaneously may match with many other `Value` types that would be unexpected behavior. Just wondering whether it should be something more specific / unique? Like bluntly `AnyValue` or `AnyType` or `Anything`.

jayvdb:
ok, yes, `type: [<all of the valid tyoes>]` is equivalent to not having any `type` at all.
But that is not what your image is showing.  it is saying it isnt happy with "null" being in the list, because `nullable: true" is another way of saying the same thing.
I am ambivalent with how we do `null`.  I can exclude it from the list, no problem.
*but* `object` is very different from `number`.  they are different, which is why `Value` is an enum with `Map` (which is `Object`) and `Number` as separate variants.

juhaku:
image

juhaku:
Aah, the `type: [string,number,integer,object,array,boolean,null]` syntax is from OpenAPI 3.1. But not valid 3.0. 

Utoipa does not support yet 3.1 spec but the support should be added sometime but not without breaking the 3.0 support. This would need a feature flag to enable 3.1 spec style.

jayvdb:
ping @juhaku .. is this now an acceptable approach?

juhaku (approved):
Super, I think this will do for now :1st_place_medal: Need to explore the OpenAPI 3.1 support later on. It for sure needs changes for `utoipa-gen` and `utoipa` types itself.

PR #575, opened 2023-04-14
Allow additional integer types

jayvdb:
Make use of formats added in https://github.com/juhaku/utoipa/pull/571

These are not recognised according to https://swagger.io/specification/#data-types , but are very common.

Maybe we need a feature flag to enable this?

### DIFF
--- a/utoipa/src/openapi/schema.rs
+++ b/utoipa/src/openapi/schema.rs
@@ -1178,20 +1178,32 @@ pub enum SchemaFormat {
 #[serde(rename_all = "lowercase")]
 pub enum KnownFormat {
     /// 8 bit integer.
+    #[cfg(feature = "non_strict_integers")]
+    #[cfg_attr(doc_cfg, doc(cfg(feature = "non_strict_integers")))]
     Int8,
     /// 16 bit integer.
+    #[cfg(feature = "non_strict_integers")]
+    #[cfg_attr(doc_cfg, doc(cfg(feature = "non_strict_integers")))]
     Int16,
     /// 32 bit integer.
     Int32,
     /// 64 bit integer.
     Int64,
     /// 8 bit unsigned integer.
+    #[cfg(feature = "non_strict_integers")]
+    #[cfg_attr(doc_cfg, doc(cfg(feature = "non_strict_integers")))]
     UInt8,
     /// 16 bit unsigned integer.
+    #[cfg(feature = "non_strict_integers")]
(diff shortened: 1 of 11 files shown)


juhaku:
Yeah, we could add a feature flag to enable extra formats. :+1:

juhaku:
```
error: test failed, to rerun pass `-p utoipa-gen --test path_derive_actix`
```
Test fails, you can run locally all tests for `utoipa-gen` with `./scripts/test.sh utoipa-gen`

jayvdb:
Ok, I'll feature flag this , then finish updating the tests.

(name suggestions? `unofficial-integer`? `non-strict-integers` ..?)

juhaku:
> (name suggestions? unofficial-integer? non-strict-integers ..?)

Hmmm.. not really :smile: , but your suggestions are pretty good ones. It is good if it is short, punchy and descriptive as much as possible :slightly_smiling_face: 

Perhaps the second one? Maybe written with underscores? `non_strict_integers`?

utoipa/src/openapi/schema.rs
```diff
@@ -1178,20 +1178,26 @@ pub enum SchemaFormat {
...
+    #[cfg(feature = "non_strict_integers")]
     UInt8,
     /// 16 bit unsigned integer.
+    #[cfg(feature = "non_strict_integers")]
     UInt16,
     /// 32 bit unsigned integer.
+    #[cfg(feature = "non_strict_integers")]
     UInt32,
     /// 64 bit unsigned integer.
+    #[cfg(feature = "non_strict_integers")]
```
juhaku, line 1200: Add the `doc_cfg` attribute as well above the variants. e.g.
```rust
    #[cfg(feature = "non_strict_integers")]
    #[cfg_attr(doc_cfg, doc(cfg(feature = "non_strict_integers")))]
   UInt64
```

This will add the `non_strict_integers` feature visible to the docs as well.

juhaku (commented):
Also add the `non_strict_integers` feature to the end of docs features as well at `utoipa/Cargo.toml` https://github.com/juhaku/utoipa/blob/3e6852db7a6275e7c23000a3a80af591106d55b8/utoipa/Cargo.toml#L49

juhaku:
Add the same feature description to `src/lib.rs` as well https://github.com/juhaku/utoipa/blob/3e6852db7a6275e7c23000a3a80af591106d55b8/utoipa/src/lib.rs#L80-L82

juhaku (approved):
Awesome :1st_place_medal:

PR #578, opened 2023-04-15
feat: Support deserializing other versions in 3.x

siketyan:
I am wondering it would be great if I can use deserialise implementations in this crate for parsing OpenAPI specifications. To do that, we need to set .version to '3.0.3', as specified in rename attribute of the serde derive. I think this crate can support parsing other versions of OpenAPI specification, such as 3.0.0 or 3.1, if the major version is '3'.

This pull request proposes to add deserialisation support of `OpenApiVersion` enum from other versions; `3.0.0`, `3.1`, or `3`.

### DIFF
--- a/utoipa/src/openapi.rs
+++ b/utoipa/src/openapi.rs
@@ -1,6 +1,7 @@
 //! Rust implementation of Openapi Spec V3.
 
-use serde::{de::Visitor, Deserialize, Serialize, Serializer};
+use serde::{de::Error, de::Visitor, Deserialize, Deserializer, Serialize, Serializer};
+use std::fmt::Formatter;
 
 pub use self::{
     content::{Content, ContentBuilder},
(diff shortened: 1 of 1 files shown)


juhaku (commented):
While this is good, but there is one change need to be made. We need to make distinction between `3.0.x` and `3.1.x` series. `3.1.x` has some changes that are different to the `3.0.x` and would require special support from `utoipa` create. 

It is being planned to add support for `3.1.x` spec to `utoipa` but before it is fully implemented I am afraid that we should not allow users to serialize OpenAPI 3.1 specs. If allowed, the resulting spec might end up being partial.

siketyan:
@juhaku I thought OpenAPI 3.1 does not have breaking changes from 3.0, so deserialising 3.1 spec as 3.0 is fine. However it is not right. The migration instruction says "v3.1 does have some small breaking changes".

I added denying 3.1 versions and now only accepts 3.0.x.

juhaku (approved):
Great :trophy:

PR #579, opened 2023-04-15
feat: Allow default value on Content::examples

siketyan:
`examples` field of Media Type object is not required in the spec, so it should have `default` attribute.

https://swagger.io/specification/#media-type-object


juhaku (approved):

juhaku:
True, we can have it there.

PR #585, opened 2023-04-17
Use swagger-ui v4.18.2

jayvdb:


jayvdb:
v5 is coming soon, so we should get the latest v4 in and stablised before the v5 arrives, as it will be a bit unstable for a while.

I did a bit of testing and this version looks good.
Also read through the relnotes at https://github.com/swagger-api/swagger-ui/releases

juhaku (approved):
Nice, :+1: that's a good idea. We can consider upgrading to v5 when it is stable since it might need some changes to get it integrated.

PR #586, opened 2023-04-17
Correct `with_schema` to `schema_with` in docs

jayvdb:
Also consistently use "schema with" in code, renaming `Property::WithSchema` to `Property::SchemaWith`.


juhaku (approved):
Although, this is good, but this enforces us to make a major version release because of breaking change cased by attribute name change from `with_schema` to `schema_with`.

PR #588, opened 2023-04-17
Add generics example

jayvdb:
I expect this may be able to be improved, in which case I will learn a bit along the way.

### DIFF
--- a/examples/generics-actix/src/main.rs
+++ b/examples/generics-actix/src/main.rs
@@ -1,0 +1,120 @@
+use std::{error::Error, net::Ipv4Addr};
+
+use actix_web::{
+    get,
+    middleware::Logger,
+    web::{Json, Query},
+    App, HttpServer, Responder, Result,
+};
+use serde::{Deserialize, Serialize};
+use utoipa::{
+    openapi::schema::{Object, ObjectBuilder},
+    IntoParams, OpenApi, PartialSchema, ToSchema,
+};
+use utoipa_swagger_ui::SwaggerUi;
+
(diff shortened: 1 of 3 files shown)


examples/generics-actix/src/main.rs
```diff
@@ -0,0 +1,117 @@
+use std::{error::Error, net::Ipv4Addr};
+
+use actix_web::{
+    get, middleware::Logger, web::Json, web::Query, App, HttpServer, Responder, Result,
+};
+use serde::{Deserialize, Serialize};
+use utoipa::{
+    openapi::{schema::Object, schema::ObjectBuilder},
```
jayvdb, line 8: can merge these two

examples/generics-actix/src/main.rs
```diff
@@ -0,0 +1,117 @@
+use std::{error::Error, net::Ipv4Addr};
+
+use actix_web::{
+    get, middleware::Logger, web::Json, web::Query, App, HttpServer, Responder, Result,
```
jayvdb, line 4: and merge the two `web::`

examples/generics-actix/src/main.rs
juhaku: Do yo mean, that the `web::Json` and `web::Query` could be provided via `utoipa` as `utoipa::Json` and `utoipa::Query`?

jayvdb: no. I was just noting the `use` syntax was not optimal.

juhaku: Ahh, you meant the imports. Sure they can.

PR #612, opened 2023-05-05
add preserve_path_order feature

### LINKED ISSUE #602: [Feature Request] Serialize paths in order of openapi paths definition
In some use-cases, the order in which paths are defined in the OpenAPI specification is significant, especially when certain paths must be included first due to their higher value over other paths. However, the current implementation of serializing OpenAPI paths alphabetically may not always be the optimal choice.

To address this, I propose adding an option to serialize OpenAPI paths in the order defined in the specification, rather than alphabetically. This would allow users to better control the order of paths and ensure they are sent in the appropriate order for their specific use-case. Issue is similar to [#428](https://github.com/juhaku/utoipa/issues/428).

### **Feature Description** 
I would like to suggest an improvement in the OpenAPI paths field where the paths are documented in the same order as they appear in the specification. 
I would appreciate your feedback on this suggestion. If you believe this change would be beneficial, I have already made the necessary changes and tested it, I would be happy to create a pull request. Please let me know your thoughts on this matter.

Chethan-rao:
As discussed in [602](https://github.com/juhaku/utoipa/issues/602) this PR introduces a cargo feature named preserve_path_order to retain path order when serializing OpenAPI paths.

Closes #602

### DIFF
--- a/utoipa/src/openapi/path.rs
+++ b/utoipa/src/openapi/path.rs
@@ -670,16 +670,19 @@ mod tests {
             )
             .build();
 
-        let mut actual_value = Vec::new();
-
-        for path in paths_list.paths.keys() {
-            let path_item = paths_list.paths.get(path);
-            if let Some(path_item) = path_item {
-                for method in path_item.operations.keys() {
-                    actual_value.push((path.as_str(), method));
-                }
-            }
-        }
+        let actual_value = paths_list
+            .paths
+            .iter()
+            .flat_map(|(path, path_item)| {
+                path_item.operations.iter().fold(
+                    Vec::<(&str, &PathItemType)>::with_capacity(paths_list.paths.len()),
+                    |mut acc, (method, _)| {
+                        acc.push((path.as_str(), method));
+                        acc
+                    },
+                )
+            })
+            .collect::<Vec<_>>();
 
         let get = PathItemType::Get;
         let post = PathItemType::Post;
--- a/utoipa/src/openapi/path.rs
+++ b/utoipa/src/openapi/path.rs
@@ -4,7 +4,7 @@
 use std::iter;
 
 #[cfg(not(feature = "preserve_path_order"))]
-use std::{collections::BTreeMap};
+use std::collections::BTreeMap;
 
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
@@ -91,7 +91,9 @@ impl PathsBuilder {
     pub fn path<I: Into<String>>(mut self, path: I, item: PathItem) -> Self {
         let path_string = path.into();
         if let Some(existing_item) = self.paths.get_mut(&path_string) {
-            existing_item.operations.extend(&mut item.operations.into_iter());
+            existing_item
+                .operations
+                .extend(&mut item.operations.into_iter());
         } else {
             self.paths.insert(path_string, item);
         }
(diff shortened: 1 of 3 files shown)


Chethan-rao:
Let me know if you want me to club this feature with already existing preserve_order feature or is it fine to introduce a new feature like I did here.

juhaku:
> Let me know if you want me to club this feature with already existing preserve_order feature or is it fine to introduce a new feature like I did here.

I think it is good to have it as a seprate cargo feature flag. This allows users to apply functionality in more granular way. Also later on there could be another feature flag that combines these feature flags all together but that is not a concern of this PR.

Just wondering whether you could write a simple unit test to test the order preservation.  There already are some tests that test structs fields order preservation. 
```rust
#[test]
#[cfg(feature = "preserver_path_order")]
fn test_foo() {
    // ...
}
```

Chethan-rao:
> > Let me know if you want me to club this feature with already existing preserve_order feature or is it fine to introduce a new feature like I did here.
> 
> I think it is good to have it as a seprate cargo feature flag. This allows users to apply functionality in more granular way. Also later on there could be another feature flag that combines these feature flags all together but that is not a concern of this PR.
> 
> Just wondering whether you could write a simple unit test to test the order preservation. There already are some tests that test structs fields order preservation.
> 
> ```rust
> #[test]
> #[cfg(feature = "preserver_path_order")]
> fn test_foo() {
>     // ...
> }
> ```

Thank you for sharing your thoughts on having a separate cargo feature flag. Regarding the simple unit test to test the order preservation, I would be happy to add it.

utoipa/src/openapi/path.rs
```diff
@@ -627,7 +637,81 @@ pub enum ParameterStyle {
...
+        let mut actual_value = Vec::new();
+
+        for path in paths_list.paths.keys() {
+            let path_item = paths_list.paths.get(path);
+            if let Some(path_item) = path_item {
+                for method in path_item.operations.keys() {
+                    actual_value.push((path.as_str(), method));
+                }
+            }
+        }
```
juhaku, line 682: Nested loops are not fun and can be hard to read and follow the code progress. I'd recommed to use the Rust's iterator API as much as possible as it provides fantastic functional interface to operate on collections of data and even more. 
```suggestion
        let actual_value = paths_list
            .paths
            .iter()
            .flat_map(|(path, path_item)| {
                path_item.operations.iter().fold(
                    Vec::<(&str, &PathItemType)>::with_capacity(paths_list.paths.len()),
                    |mut acc, (method, _)| {
                        acc.push((path.as_str(), method));
                        acc
                    },
                )
            })
            .collect::<Vec<_>>();
```

juhaku (changes requested):

utoipa/src/openapi/path.rs
Chethan-rao: I'm new to Rust programming and am still in the learning phase. I appreciate your inputs. I'll make the necessary changes and thanks for your suggestion.

juhaku: That's fine, :slightly_smiling_face: we all have been there at some point not matter the language. The willingness to learn and constantly improve is the key :+1:

juhaku (approved):
:100: super

PR #613, opened 2023-05-06
Fix typo in README

Ciubix8513:
`it could also be` sounds more natural


juhaku (approved):
While both of presentations are correct, I am find with any.

PR #618, opened 2023-05-15
Omit decimal zeros when serializing minimum/maximum/multiple

rbuch703:
This way, when these values are integers their serialized form is also a valid integer.


juhaku (commented):
This is awesome, could you write a test to `utoipa-gen/tests/schema_test.rs` to test this out.

juhaku (approved):

PR #619, opened 2023-05-16
Fix tests for feature non_strict_integers

jayvdb:


utoipa-gen/tests/schema_derive_test.rs
```diff
@@ -3864,11 +3864,6 @@ fn derive_struct_with_validation_fields() {
                     "format": "uint32",
                     "minimum": 2.0,
                 }
-                "unsigned_value": {
```
jayvdb, line 3867: this is a json syntax error in my last PR. ;-(

utoipa/src/lib.rs
```diff
@@ -1020,22 +1020,22 @@ mod tests {
             (
                 "u8",
                 u8::schema(),
-                json!({"type": "integer", "format": "uint8", "minimum": 0.0}),
+                json!({"type": "integer", "format": "uint8", "minimum": 0}),
```
jayvdb, line 1023: This appears to be caused by https://github.com/juhaku/utoipa/pull/618 - which is great.
but I am confused why the `test_partial_schema_strict_integers` test directly above this one doesnt also need the `.0` removed..?

juhaku: It was forgotten from previous commit to update this test. The thing seemingly is because the number comparison is done differently. This from the test above.
```rust
let config = Config::new(CompareMode::Strict).numeric_mode(NumericMode::AssumeFloat);
assert_json_matches!(schema, value, config);
```

juhaku: Anyhow I think we can live with that difference, and merge this to the master :smile:

juhaku (approved):

PR #627, opened 2023-05-26
exclude const generic arguments from generic_types

### LINKED ISSUE #626: procmacro incompatible with const generics
I have an Axum server that uses const generics to determine the required Role for the route, but when adding `utoipa::path` procmacro it gives an error (you can see it at the end)

this is a simplified version of the issue:

```rust

struct CoolExtractor<const USER_ROLE: u32>;

#[repr(u32)]
enum UserRole {
    Admin,
    User
}

#[utoipa::path(get, path = "/")]
async fn root(
   very_cool: CoolExtractor<{ UserRole::User as u32 }>,
) {}

#[tokio::main]
async fn main() {
    let app = Router::new()
        .route("/", get(root));

    let addr = SocketAddr::from((Ipv4Addr::UNSPECIFIED, 6060));

    axum::Server::bind(&addr)
        .serve(app.into_make_service())
        .await
        .expect("start server");
}
```

```
error: expected generic argument type or generic argument lifetime
   --> /cool/path
    |
460 |     very_cool: CoolExtractor<{ UserRole::User as u32 }>,
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^
```

it fails because of this `abort!` invocation because it doesn't handle the `Const` variant:
```rust
...
                        angle_bracketed_args
                            .args
                            .iter()
                            .filter(|arg| !matches!(arg, GenericArgument::Lifetime(_)))
                            .map(|arg| match arg {
                                GenericArgument::Type(arg) => arg,
                                _ => abort!(
                                    arg,
                                    "expected generic argument type or generic argument lifetime"
                                ),
                            }),
...
```

the code should probably be adjusted to check that branch. unfortunately im not too good with proc-macro dev, otherwise id give fixing it a try

BKSalman:
resolves #626 

nvm got help from @aumetra and solved it :)

### DIFF
--- a/utoipa-gen/src/component.rs
+++ b/utoipa-gen/src/component.rs
@@ -208,18 +208,24 @@ impl<'t> TypeTree<'t> {
         let mut generic_types = match &last_segment.arguments {
             PathArguments::AngleBracketed(angle_bracketed_args) => {
                 // if all type arguments are lifetimes we ignore the generic type
-                if angle_bracketed_args
-                    .args
-                    .iter()
-                    .all(|arg| matches!(arg, GenericArgument::Lifetime(_)))
-                {
+                if angle_bracketed_args.args.iter().all(|arg| {
+                    matches!(
+                        arg,
+                        GenericArgument::Lifetime(_) | GenericArgument::Const(_)
+                    )
+                }) {
                     None
                 } else {
                     Some(
                         angle_bracketed_args
                             .args
                             .iter()
-                            .filter(|arg| !matches!(arg, GenericArgument::Lifetime(_)))
+                            .filter(|arg| {
+                                !matches!(
+                                    arg,
+                                    GenericArgument::Lifetime(_) | GenericArgument::Const(_)
+                                )
+                            })
                             .map(|arg| match arg {
                                 GenericArgument::Type(arg) => arg,
                                 _ => abort!(


juhaku (approved):
Just wondering whether it actually should work with const as well if so, then the fix would be more evolved. Yet at the moment it is not working with consts so I guess it is just fine to ignore them though. But most likely users with a struct having generic const variable it would still fail but with a different error than now.

However that can be another PR later on. I'll create an issue to the kanban board to track this if I ever get to test the cons support.

BKSalman:
> Just wondering whether it actually should work with const as well if so, then the fix would be more evolved. Yet at the moment it is not working with consts so I guess it is just fine to ignore them though. But most likely users with a struct having generic const variable it would still fail but with a different error than now.
> 
> However that can be another PR later on. I'll create an issue to the kanban board to track this if I ever get to test the cons support.

sounds good :)

I was actually trying to somehow change the code to make it also support consts (not just filter it out), but I didn't even know what the idea of it would be, or where it can be used in the generated docs

aumetra:
> or where it can be used in the generated docs

Yeah, whatever can be done with const-generics (at the moment) seems somewhat irrelevant to OpenAPI docs (I could also just not be aware of use-cases).

> But most likely users with a struct having generic const variable it would still fail but with a different error than now.

Do you mean with a constant associated type? Something like `Foo<BAR = 2>`? These will receive the same errors presumably, since they fall under the branch `GenericArgument::AssocConst`, which isn't covered by `utoipa`.  

If you mean something like `Foo<2>`, it would fall under the same branch (`GenericArgument::Const`) and would be ignored.

juhaku:
> Do you mean with a constant associated type? Something like `Foo<BAR = 2>`? These will receive the same errors presumably, since they fall under the branch `GenericArgument::AssocConst`, which is covered by `utoipa` to fail with the same error.
>
> If you mean something like `Foo<2>`, it would fall under the same branch (`GenericArgument::Const`) that was patched with this PR and would be ignored.

Either of those, frankly I haven't looked in to the details how or what could be done with a `const` generic type. For `utoipa` OpenAPI docs it indeed does not make much sense and even a generic `ToSchema` type should be used with `#[aliases]` that tell the actual types of the generic arg anyway.

PR #641, opened 2023-06-07
Add support for chrono::NaiveTime

beyarkay:
Since the OpenAPI spec does not explicitly support times without attached dates, I have implemented this in the same way that support for `duration` was done, but formatting them as strings.

### DIFF
--- a/utoipa-gen/src/schema_type.rs
+++ b/utoipa-gen/src/schema_type.rs
@@ -174,26 +174,28 @@ impl ToTokens for SchemaType<'_> {
             "String" | "str" | "char" => {
                 tokens.extend(quote! {utoipa::openapi::SchemaType::String})
             }
+
             "bool" => tokens.extend(quote! { utoipa::openapi::SchemaType::Boolean }),
+
             "i8" | "i16" | "i32" | "i64" | "i128" | "isize" | "u8" | "u16" | "u32" | "u64"
             | "u128" | "usize" => tokens.extend(quote! { utoipa::openapi::SchemaType::Integer }),
             "f32" | "f64" => tokens.extend(quote! { utoipa::openapi::SchemaType::Number }),
+
             #[cfg(feature = "chrono")]
-            "DateTime" => tokens.extend(quote! { utoipa::openapi::SchemaType::String }),
-            #[cfg(feature = "chrono")]
-            "NaiveDateTime" => tokens.extend(quote! { utoipa::openapi::SchemaType::String }),
-            #[cfg(feature = "chrono")]
-            "NaiveDate" => tokens.extend(quote!(utoipa::openapi::SchemaType::String)),
-            #[cfg(feature = "chrono")]
-            "NaiveTime" => tokens.extend(quote!(utoipa::openapi::SchemaType::String)),
+            "DateTime" | "NaiveDateTime" | "NaiveDate" | "NaiveTime" => tokens.extend(quote! { utoipa::openapi::SchemaType::String }),
+
             #[cfg(any(feature = "chrono", feature = "time"))]
             "Date" | "Duration" => tokens.extend(quote! { utoipa::openapi::SchemaType::String }),
+
             #[cfg(feature = "decimal")]
             "Decimal" => tokens.extend(quote! { utoipa::openapi::SchemaType::String }),
+
             #[cfg(feature = "rocket_extras")]
             "PathBuf" => tokens.extend(quote! { utoipa::openapi::SchemaType::String }),
+
             #[cfg(feature = "uuid")]
             "Uuid" => tokens.extend(quote! { utoipa::openapi::SchemaType::String }),
+
             #[cfg(feature = "time")]
             "PrimitiveDateTime" | "OffsetDateTime" => {
                 tokens.extend(quote! { utoipa::openapi::SchemaType::String })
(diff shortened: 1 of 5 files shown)


beyarkay:
I am getting some errors when I try to run the tests, but I'm not sure if it's because I have the wrong feature flags set up or because I'm on an M1 mac and that's doing funky things. 

<details>
<summary>Errors</summary>

```
$ cargo test -F debug,chrono
   Compiling utoipa-gen v3.3.0 (/Users/brk/projects/open_source/utoipa/utoipa-gen)
   Compiling utoipa-swagger-ui v3.1.3 (/Users/brk/projects/open_source/utoipa/utoipa-swagger-ui)
error[E0061]: this enum variant takes 1 argument but 0 arguments were supplied
    --> utoipa-gen/tests/schema_derive_test.rs:14:22
     |
14   |               #[derive(ToSchema)]
     |                        ^^^^^^^^ an argument of type `SchemaFormat` is missing
...
2699 |       let post = api_doc! {
     |  ________________-
2700 | |         struct Post {
2701 | |             id: i32,
2702 | |             value: String,
...    |
2709 | |         }
2710 | |     };
     | |_____- in this macro invocation
     |
note: tuple variant defined here
    --> /Users/brk/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/option.rs:572:5
     |
572  |     Some(#[stable(feature = "rust1", since = "1.0.0")] T),
     |     ^^^^
     = note: this error originates in the derive macro `ToSchema` which comes from the expansion of the macro `api_doc` (in Nightly builds, run with -Z macro-backtrace for more info)
help: provide the argument
     |
14   |             #[derive(ToSchema(/* SchemaFormat */))]
     |                              ++++++++++++++++++++

For more information about this error, try `rustc --explain E0061`.
error: could not compile `utoipa-gen` (test "schema_derive_test") due to previous error
warning: build failed, waiting for other jobs to finish...
```

</details>

The `note` in particular seems specific to `stable-aarch64-apple-darwin`.

```
note: tuple variant defined here
    --> /Users/brk/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/option.rs:572:5
```

Do you have GitHub actions setup to run the tests? Or alternatively, could you point me in the right direction for how I could resolve this issue?

Thanks for the great crate!

juhaku:
> Do you have GitHub actions setup to run the tests? Or alternatively, could you point me in the right direction for how I could resolve this issue?

Yes there is Github Actions, also you can use the shell script `./scripts/test.sh` to run the tests in crate basis `./scripts/test.sh <crate>` e.g. `./scripts/test.sh utoipa-gen`.

juhaku (approved):
Awesome, looks good :tada: Thanks for the patch and great that you like the crate.

beyarkay:
Okay thanks, it's still [failing](https://github.com/juhaku/utoipa/actions/runs/5204205509/jobs/9389224481?pr=641#step:7:395) I think because I haven't implemented `NaiveTime` properly. I've tried looking at the implementation of `Duration` (since it's also just implemented as a string) and `NaiveDate` (since it's also from `chrono`) but I can't see what I'm missing.

Any chance you know what needs to be added?

juhaku:
After short debuggin the issue in `utoipa-gen/src/schema_type.rs` I'll try to fix the issue soon. There seems to be some overlapping feature not just consering the chrono itself but others as well one of them being uuid. These overlapping features causes the test to behave weirdly what manifests as not correctly creating the schema. e.g. in this case the generated schema is missing the `SchemaType`. 

You can also test this behavior by runnig this test with different flags:
```
cargo test -p utoipa-gen --features actix_extras,time,time,repr,uuid

cargo test -p utoipa-gen --features actix_extras,time,time,repr // <-- this passes

cargo test -p utoipa-gen --features actix_extras,time,time,repr,chrono
```

utoipa-gen/src/schema_type.rs
```diff
@@ -266,7 +268,10 @@ impl Type<'_> {
 
             #[cfg(feature = "chrono")]
             if !known_format {
-                known_format = matches!(name, "DateTime" | "Date" | "NaiveDate" | "NaiveDateTime");
+                known_format = matches!(
+                    name,
+                    "DateTime" | "Date" | "NaiveDate" | "NaiveTime" | "NaiveDateTime"
```
juhaku, line 273: Now I know what is going on here. The issue is that you have defined `NaiveTime` as a known format, but it does not have a format, but only a type which should be string. Now with that known format the naive_time will generate the spec as:
```rust
                .property(
                    "naive_time",
                        .schema_type(utoipa::openapi::SchemaType::String)
                        .format(Some()), // <-- note missing format!
                )
```
Note the missing format, it renders empty format because you haven't defined what should be the format of the `NaiveTime`. 
https://github.com/juhaku/utoipa/blob/f8c6d07ef699d4fd6d3eeddfb55d6955d89af0f6/utoipa-gen/src/schema_type.rs#L295-L340

Though in this case I believe the fix would be to remove the `NaiveTime` from  known format function since there is no format specifier for the time like the duration does not have.

juhaku: And then the format in the test should be `null` as in below:
```rust
        "properties.naive_time.format" = r#"null"#, "Post time format"
```

beyarkay:
Thanks so much!

Issue #647, opened 2023-06-12
openapi::schema::empty is confusing or wrong


jayvdb:
https://docs.rs/utoipa/latest/utoipa/openapi/schema/fn.empty.html says

> Create an empty [Schema](https://docs.rs/utoipa/latest/utoipa/openapi/schema/enum.Schema.html) that serializes to null.
> Can be used in places where an item can be serialized as null. This is used with unit type enum variants and tuple unit types.

It becomes a jsonschema
```yaml
type: object
default: ~
nullable: true
```

Because it has `type: object` in there, it can not be used to describe an empty string, int, or any other type.  It can only be used to describe an empty *object*.

It should be changed to use the new `Value` support added in https://github.com/juhaku/utoipa/pull/568

Issue #576, opened 2023-04-15
Accept HashMap<(), ()> as object


leon3s:
Hey i'm generating structures from the docker openapi spec and for some field they use `HashMap<(), ()>` for example:
```rust
volumes: HashMap<String, HashMap<(), ()>>
```
To define an empty object [see](https://docs.docker.com/engine/api/v1.42/#tag/Container/operation/ContainerCreate) for volumes as example.

But if i want to convert it back using utoipa i get the following error:
```console
error[E0277]: the trait bound `AdditionalProperties<Schema>: From<Schema>` is not satisfied
    --> codegen/swagger/src/models.rs:6637:39
     |
6637 | #[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
     |                                       ^^^^^^^^^^^^^^^^ the trait `From<Schema>` is not implemented for `AdditionalProperties<Schema>`
     |
     = help: the following other types implement trait `From<T>`:
               <AdditionalProperties<Schema> as From<()>>
               <AdditionalProperties<Schema> as From<ObjectBuilder>>
               <AdditionalProperties<Schema> as From<utoipa::openapi::Ref>>
     = note: required for `Schema` to implement `Into<AdditionalProperties<Schema>>`
note: required by a bound in `ObjectBuilder::additional_properties`
```

I tried to do it, but i'm not familiar enough with macro

juhaku:
In docker OpenAPI spec the `Volumes` field is defined as follows:
```yaml
Volumes:
  description: |
    An object mapping mount point paths inside the container to empty
    objects.
  type: "object"
  additionalProperties:
    type: "object"
    enum:
      - {}
    default: {}
```

But `utoipa` will render the volumes differently becuase of the unit type `()`. Unit type is by default rendered as `null` with serde and `utoipa` will create schemas for both types of the `HashMap` by the way of using `additionalProperties` resulting strictier interpretation of OpenAPI schema than what is defined in the docker OpenAPI schema.
```json
{
    "properties": {
        "volumes": {
            "additionalProperties": {
                "additionalProperties": {
                    "type": "object",
                    "default": null,
                    "nullable": true,
                },
                "type": "object"
            },
            "type": "object"
        },
    },
    "required": [
        "volumes"
    ],
    "type": "object"
}
```

[...]

PR #648, opened 2023-06-12
Remove `type: object` restriction in empty()

jayvdb:
Fixes https://github.com/juhaku/utoipa/issues/647

### DIFF
--- a/utoipa-gen/tests/schema_derive_test.rs
+++ b/utoipa-gen/tests/schema_derive_test.rs
@@ -4038,7 +4038,6 @@ fn derive_unit_type() {
             "required": [ "unit_type" ],
             "properties": {
                 "unit_type": {
-                    "type": "object",
                     "default": null,
                     "nullable": true
                 }
@@ -4056,7 +4055,6 @@ fn derive_unit_struct_schema() {
     assert_json_eq!(
         value,
         json!({
-            "type": "object",
             "nullable": true,
             "default": null,
         })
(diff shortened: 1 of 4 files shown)


utoipa-gen/tests/schema_derive_test.rs
```diff
@@ -4477,7 +4474,6 @@ fn derive_schema_with_unit_hashmap() {
                 "volumes": {
                     "additionalProperties": {
                         "additionalProperties": {
-                            "type": "object",
```
jayvdb, line 4480: I am not sure about this one.  To begin with I dont get `HashMap<(), ()>` because in JSON keys must be strings. https://stackoverflow.com/questions/9304528/why-json-allows-only-string-to-be-a-key

juhaku: That was to provide some sort of compatibility for https://github.com/juhaku/utoipa/issues/576 to mitigate a case where `additionalProperties` key and value is both defined as empty object `{}` as seen in docker's OpenAPI specification.

juhaku (approved):
Looks good

utoipa-gen/tests/schema_derive_test.rs
juhaku: But for compatibility it does not matter whether there actually is `type: object` or not because it is enough that `utoipa` parses and compiles the schema.

Issue #591, opened 2023-04-18
Automatic type recognition for path operation handlers


juhaku:
## Description

To streamline further the `utoipa` library and to provide deeper integeration for frameworks like `actix_web`, `axum`, and `rocket` it would be great to be able to automatically recognize necessary types from operation handler and add them to the path schema of OpenAPI doc generated with `utoipa`. This will provide more automation and saves the keystrokes the user is ought to take to define `#[utoipa::path(...)]` macro content manually.

For example, from following handler function we would need to be able to recognize _`path`_, _`query`_, _`body`_ and _`response`_. These recognized values, if successful, should be added automatically to the generated OpenAPI spec.
```rust
#[utoipa::path]
async fn post_item(path: Path<PathParams>, query: Query<QueryParams>, item: Json<ItemBody>) -> Response {
    Response::Success(Item { value: "super".to_string() })
}
```

This functionality will enforce users to be more strict on what is being returned from the handler functions and what are the args given to the handler functions. Benefit of this is the users are able to declare the OpenAPI spec mostly via code and get compile errors directly to the code of invalid spec.

## Tasks

- [ ] Implement general response body recognition - All responses need to implement `IntoResponses` trait.
- [ ] Implement response body recognition for actix_web
- [ ] Implement response body recognition for axum
- [ ] Implement response body recognition for rocket
- [ ] Implement request body recognition for actix_web
- [ ] Implement request body recognition for axum
- [ ] Implement request body recognition for rocket
- [ ] Enhance current path and query parameter support in actix_web to support auto recognition
- [ ] Enhance current path and query parameter support in axum to support auto recognition
- [ ] Enhance current path and query parameter support in rocket to support auto recognition

TheDan64:
I think related; it was really surprising (in axum) that `Path<(X, Y)>` would auto generate the path params but `Path<X>` does not

juhaku:
Yeah, I can take a look at this at the same time.

juhaku:
[...]

PR #649, opened 2023-06-12
Wip experiemental actix_web auto-repsonses

juhaku:
Experiemental actix_web auto-responses from handler return type.

There are noticeable drawbacks with the automatic response resolution. And the biggest one is question "How to override response a type that is resolved automatically? Or even ignore one?"

Currently I have no idea on how to proceed with this one. And as the implementation is now, is too limiting, meaning it does not cooperate well with the tuple style responses.

Any ideas are welcome.

#591


DenuxPlays:
Any updates?

PR #669, opened 2023-07-10
Fix broken links

TaKO8Ki:
`openapi::path::ParameterIn` is broken in a doc.

ref: 

- https://docs.rs/utoipa/latest/utoipa/attr.path.html#tuples
- https://docs.rs/utoipa/latest/utoipa/derive.IntoParams.html#intoparams-container-attributes-for-into_params


juhaku (approved):
Thanks for noticing and fixing :+1:

Issue #685, opened 2023-07-17
Support for field deprecation without using #[deprecated]


nullsauce:
Hi there,

Just a remark: It would be handy if it was possible to to mark a field `deprecated` in the spec, without having to use `#[deprecated]`, as it can generate spurious warnings.

Maybe a ` #[schema(deprecated)]` like below ?

```
#[derive(ToSchema)]
struct Foo {
  #[schema(deprecated)]
  bar: String
}
``

juhaku:
So you would like to mark some field deprecated in OpenAPI spec, but not actually in code? Isn't that a bit counter intuitive? As what I consider very useful in Rust is that if you mark fields or generally anything `#[deprecated]` you get the warnings directly to your editor. This works as constant reminder that something should be migrated to another implementation? With that, then `utoipa` just follows the code base deprecation rules. Because if there is no warnings anywhere then at then end the parameters just get buried under the ever changing API. Though if the migration loop is long, it might become irritating to be constantly warned about things that are not actionable at the moment.

Nonetheless implementing this support is relatively small effort. And PR:s are welcome. My timeline is so congested that I am not able to provide any timeline for adding support for this unfortunately.

juhaku:
> ...Hope it makes sense. 

Yeah, makes sense..

Great that you found the place where to make changes.

PR #688, opened 2023-07-18
Adds support for deprecation using schema attribute

nullsauce:
Follow-up to #685 

### Summary 

Attempt at adding a new `schema(deprecated)` attribute that allows to mark a field, enum or struct as deprecated in the API spec. Has the same effect on the generated API spec as using `#[deprecated]`.

### DIFF
--- a/utoipa-gen/src/lib.rs
+++ b/utoipa-gen/src/lib.rs
@@ -1732,10 +1732,6 @@ pub fn openapi(input: TokenStream) -> TokenStream {
 ///   Free form type enables use of arbitrary types within map values.
 ///   Supports formats _`additional_properties`_ and _`additional_properties = true`_.
 ///
-/// * `deprecated` Can be used to mark the field as deprecated in the generated OpenAPI spec but
-///   not in the code. If you'd like to mark the field as deprecated in the code as well use
-///   Rust's own `#[deprecated]` attribute instead.
-///
 /// #### Field nullability and required rules
 ///
 /// Same rules for nullability and required status apply for _`IntoParams`_ field attributes as for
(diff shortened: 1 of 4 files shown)


utoipa-gen/src/lib.rs
```diff
@@ -180,6 +180,9 @@ use self::{
 ///   [`HashMap`](std::collections::HashMap) and [`BTreeMap`](std::collections::BTreeMap).
 ///   Free form type enables use of arbitrary types within map values.
 ///   Supports formats _`additional_properties`_ and _`additional_properties = true`_.
+/// * `deprecated` Can be used to mark field as deprecated in the generated OpenAPI spec but
+///   not in the code. If you'd like to mark field as deprecated in the code as well use
+///   Rust's own `#[deprecated]` attribute instead.
```
juhaku, line 185: This is good addition. Though this doc should be repeated to the all the types the `deprected` is supported. e.g. for the unnamed structs and enums as well. Also it's worth to check whether `IntoParams` docs are in up-to-date as well.

You can generate the doc with `./scripts/doc.sh` and then just open the generated files with your favorite browser. Files can be found from `target` folder.

juhaku: Furthermore these changes contain the `deprecated` support for `ToSchema` derived types. But query parameters that are derived from `IntoParams` are not affected by this change. They use the same features though and the code can be found from `into_params.rs`. If you find this useful for them as well it might be worthwhile to make some changes there as well. 

All the docs are still in the same place `lib.rs` but just above the `IntoParams` derive trait function.

nullsauce: Makes sens. Updated the docs where it was missing, I think.. Let me know

juhaku:
Are you going to make any changes to `into_params.rs`? If not I can merge this one.

nullsauce:
Please go ahead. I most likely won’t have time to dive in into_params very
soon.

PR #694, opened 2023-07-20
Allow settings titles on all OpenApi Schema types and allow descriptions to propagate for UnnamedStructSchema

lithiumFlower:


juhaku:
You could add few test cases to `schema_derive_test.rs` file to test the description on those edge cases.

lithiumFlower:
I'll add test cases to the suggested file. If you have any other pointers or contribution guidelines please let me know!

juhaku:
> d test cases to the suggested file. If you have any other pointers or contribution guidelines please let me know!

Sure, usually docs are also good to update but this change does not concern anything that would make public api changes. Thus with this PR there is nothing to update to the docs.

utoipa-gen/src/component/schema.rs
```diff
@@ -588,21 +580,15 @@ impl ToTokens for UnnamedStructSchema<'_> {
...
                 tokens.extend(attrs.to_token_stream())
             }
-        };
-
-        let description = CommentAttributes::from_attributes(self.attributes).as_formatted_string();
-        if !description.is_empty() && !is_object {
-            tokens.extend(quote! {
-                .description(Some(#description))
-            })
         }
```
juhaku, line 583: The reason why there is `!is_object` check is that description cannot be set on `$ref` type.

lithiumFlower: Yes :) But when we convert to an array we no longer have a ref. Without this change 2 of the tests I modified will not produce descriptions. 

You're right that it will error on the Ref type if we try to add a description via token extension to any unnamed struct parse. The enum reference tests catch that case and error (and do not error in this case because we're converting to array).

TLDR Ref types don't error with snippet and the snippet enables descriptions on unnamed structs with more than one field 

How would you like me to modify it?

juhaku: I'd make sure that this scenario passes:
```rust
#[derive(ToSchema)]
struct Bar {
  value: String
}

/// Description now will be tried to pass on $ref that is created for inner type Bar
#[derive(ToSchema)]
struct Foo(Bar);
```

Though this will be possible in openapi 3.1. But that needs other changes as well and is another topic for later PRs.

lithiumFlower: Thank you. I added the above test, which doesn't fail and doesn't add a description to the ref if I understand the testing correctly.

juhaku (approved):
:medal_sports: super

utoipa-gen/src/component/schema.rs
juhaku: Great

lithiumFlower:
This is ready to merge so feel free to whenever

PR #706, opened 2023-07-25
Adding "AnyOf" branch for Schema so that utoipa can deserialize schem…

zajko:
…as produced by other libraries


juhaku (approved):
Nice :+1:

PR #708, opened 2023-07-25
allow and ignore #[doc(...)] tags in ToSchema derive

### LINKED ISSUE #704: ToSchema Derive fails when #[doc(...)] is used
`ToSchema` generate a compiler error when trying this code:
```rust
use utoipa::ToSchema;

#[derive(ToSchema)]
#[doc(alias = "Animal")]
struct Pet {
   id: u64,
   name: String,
   age: Option<i32>,
}
```
The error:
```
error: Expected only Meta::NameValue type
 --> src/main.rs:3:10
  |
3 | #[derive(ToSchema)]
  |          ^^^^^^^^
  |
  = note: this error originates in the derive macro `ToSchema` (in Nightly builds, run with -Z macro-backtrace for more info)
```

This happens because utoipa wants only pub enum `doc` attribute which are `Meta:NameValue` (i.e. doc comment), but `#[doc(something)]` is a `Meta::List` and gets rejected instead of ignored. Note that the order of `derive`/`doc` doesn't seem to matter.

trinity-1686a:
fix #704

### DIFF
--- a/utoipa-gen/src/doc_comment.rs
+++ b/utoipa-gen/src/doc_comment.rs
@@ -1,7 +1,6 @@
 use std::ops::Deref;
 
 use proc_macro2::Ident;
-use proc_macro_error::abort_call_site;
 use syn::{Attribute, Expr, Lit, Meta};
 
 const DOC_ATTRIBUTE_TYPE: &str = "doc";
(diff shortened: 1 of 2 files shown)


juhaku (approved):

juhaku:
I'd create one small test to `schema_derive_test.rs` file that tests this case.

juhaku (approved):
Great

PR #709, opened 2023-07-27
Fix typos in doc

ben6reul:
Fixing some minor typos I found while reading the docs


juhaku (approved):
:100: Awesome.. thanks for proof reading.. The "copy ala pasta" has definitely taken it's toll when writing the docs.

Issue #681, opened 2023-07-14
ToSchema on an enum with  docstring-prefaced tuple variant containing an Option as the first element of the tuple does not produce a description


lithiumFlower:
Creating an enum with a tuple variant that contains an Option as the first element does not produce a description for that variant despite being prefaced with a docstring and deriving from ToSchema.

Here are 2 examples and their corresponding openapi json output to reproduce the issue:
```rust
/// top level doc for My enum
#[derive(ToSchema)]
pub enum MyEnum {
            /// doc for tuple variant with Option as first element - I do not produce a description
            TupleVariantWithOptionFirst(Option<String>),

            /// doc for tuple variant without Option as first element - I produce a description
            TupleVariantWithNoOption(String),
}
```
produces:
```json
      "MyEnum": {
        "oneOf": [
          {
            "type": "object",
            "required": [
              "TupleVariantWithOptionFirst"
            ],
            "properties": {
              "TupleVariantWithOptionFirst": {
                "type": "string",
                "nullable": true
              }
            }
          },
          {
            "type": "object",
            "required": [
              "TupleVariantWithNoOption"
            ],
            "properties": {
              "TupleVariantWithNoOption": {
                "type": "string",
                "description": "doc for tuple variant without Option as first element - I produce a description"
              }
            }
          }
        ],
        "description": "top level doc for My enum"
      },
```

```rust
/// top level doc for My enum
#[derive(ToSchema)]
pub enum MyEnum {
            /// doc for tuple variant with Option as first element - I do not produce a description
            TupleVariantWithOptionFirst(Option<String>, String),

            /// doc for tuple variant without Option as first element - I produce a description
            TupleVariantWithOptionSecond(String, Option<String>),
}
```
produces:
```json
      "MyEnum": {
        "oneOf": [
          {
            "type": "object",
            "required": [
              "TupleVariantWithOptionFirst"
            ],
            "properties": {
              "TupleVariantWithOptionFirst": {
                "type": "array",
                "items": {
                  "type": "object"
                },
                "maxItems": 2,
                "minItems": 2
              }
            }
          },
          {
            "type": "object",
            "required": [
              "TupleVariantWithOptionSecond"
            ],
            "properties": {
              "TupleVariantWithOptionSecond": {
                "type": "array",
                "items": {
                  "type": "object",
                  "description": "doc for tuple variant without Option as first element - I produce a description"
                },
                "maxItems": 2,
                "minItems": 2
              }
            }
          }
        ],
[...]

Issue #679, opened 2023-07-14
ToSchema on an enum with a documented unit variant does not produce a description


lithiumFlower:
If I create an enum with unit variants that derives from ToSchema I'm unable to use docstrings to generate descriptions for the unit variants. For example:

```rust
/// top level doc for My enum
#[derive(ToSchema)]
pub enum MyEnum {
            /// unit variant doc
            UnitVariant,

            /// non unit variant doc
            NonUnitVariant(String),
}
```

generates the following openapi json:
```json
      "MyEnum": {
        "oneOf": [
          {
            "type": "string",
            "enum": [
              "UnitVariant"
            ]
          },
          {
            "type": "object",
            "required": [
              "NonUnitVariant"
            ],
            "properties": {
              "NonUnitVariant": {
                "type": "string",
                "description": "non unit variant doc"
              }
            }
          }
        ],
        "description": "top level doc for My enum"
      },
```

juhaku:
True that, just checked and in fact that support has not been implemented. Seems like it has been forgotten somehow :smile: 

Need investigate in future what kind of effort would it be to get this support done.

juhaku:
I boldly guess this is related to the #681.

juhaku:
@lithiumFlower  Is this now something  that can be closed?

lithiumFlower:
Yes, this is complete :)

arifd:
If the enum contains only unit variants, no documentation is created.

this code:
```rs
use std::{error::Error, net::Ipv4Addr};

use actix_web::{App, HttpServer};
use utoipa::{ToSchema, OpenApi};
use utoipa_swagger_ui::{SwaggerUi, Url};

#[derive(ToSchema)]
pub enum MyEnum {
            /// unit variant doc
            UnitVariant,

            /// non unit variant doc
            NonUnitVariant(String),
}

#[derive(ToSchema)]
pub enum AllUnits {
        /// unit variant doc
        UnitA,

        /// another unit variant doc
        UnitB
}

#[actix_web::main]
async fn main() -> Result<(), impl Error> {
    #[derive(OpenApi)]
    #[openapi(components(schemas(MyEnum, AllUnits)))]
    struct ApiDoc;

    HttpServer::new(move || {
        App::new()
            .service(SwaggerUi::new("/swagger-ui/{_:.*}").urls(vec![
                (
                    Url::new("api", "/api-docs/openapi.json"),
                    ApiDoc::openapi(),
                ),
            ]))
    })
[...]

PR #714, opened 2023-07-30
Add descriptions to 2 variants of complex enums 

lithiumFlower:
Add descriptions to 2 variants of complex enums:
1 unit variants
2. tuple variants with an `Option` as the first element

Resolves:
https://github.com/juhaku/utoipa/issues/681
https://github.com/juhaku/utoipa/issues/679

### DIFF
--- a/utoipa-gen/src/component/schema.rs
+++ b/utoipa-gen/src/component/schema.rs
@@ -10,7 +10,7 @@ use syn::{
 };
 
 use crate::{
-    component::features::{Example, Rename},
+    component::features::{Description, Example, Rename},
     doc_comment::CommentAttributes,
     Array, Deprecated, ResultExt,
 };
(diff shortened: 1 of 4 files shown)


utoipa-gen/src/component/schema.rs
```diff
@@ -1007,7 +1007,12 @@ impl ComplexEnum<'_> {
                     rename_all,
                 );
 
-                let example = pop_feature!(unit_features => Feature::Example(_));
+                let example: Option<Feature> = pop_feature!(unit_features => Feature::Example(_));
+
+                let description =
+                    CommentAttributes::from_attributes(&variant.attrs).as_formatted_string();
+                let description = (!description.is_empty())
+                    .then(|| Feature::Description(Description(description)));
```
lithiumFlower, line 1015: If desired I can extend the enum token stream with the correct tokenization instead of wrapping with `Feature::Description` (which needs the pub identifier `features.rs`). I like leveraging the existing source of truth for description token generation but can also see that maybe we don't want to use a `Feature` for a non-feature.

juhaku: Yeah less sources of truth there are the better. There are few other cases where a feature is being manually constructed from a value. I think this is fine but though better would be to implement `From` trait for the description feature. This way there is no need to make the value public.
```rust
impl From<String> for Description {
  fn from(value: String) -> Self {
    Self(value)
  }
}
```

lithiumFlower:
Changed to use the From trait for String -> Description, nice suggestion.

Ready to merge :)

PR #733, opened 2023-08-12
Support ULID

pbzweihander:
This PR adds support for [ulid](https://github.com/dylanhart/ulid-rs).

### DIFF
--- a/utoipa-gen/src/schema_type.rs
+++ b/utoipa-gen/src/schema_type.rs
@@ -428,7 +428,7 @@ impl Parse for Variant {
                 }
                 #[cfg(all(not(feature = "uuid"), not(feature = "ulid")))]
                 {
-                    _format != &"Ulid" && _format != &"Ulid"
+                    _format != &"Uuid" && _format != &"Ulid"
                 }
             })
             .collect::<Vec<_>>();
--- a/utoipa-gen/src/schema_type.rs
+++ b/utoipa-gen/src/schema_type.rs
@@ -34,6 +34,7 @@ impl SchemaType<'_> {
             feature = "decimal",
             feature = "rocket_extras",
             feature = "uuid",
+            feature = "ulid",
             feature = "time",
         )))]
         {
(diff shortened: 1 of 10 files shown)


utoipa-gen/src/schema_type.rs
```diff
@@ -372,26 +400,36 @@ pub enum Variant {
...
                 }
+                #[cfg(all(feature = "uuid", not(feature = "ulid")))]
+                {
+                    _format != &"Ulid"
+                }
+                #[cfg(all(not(feature = "uuid"), not(feature = "ulid")))]
+                {
+                    _format != &"Ulid" && _format != &"Ulid"
+                }
             })
```
juhaku, line 433: This seems a bit fishy

juhaku (changes requested):
Seems otherwise good, just a test missing from `schema_derive_test.rs` for the ulid case.

Also perhaps the known format thingy could be bit refactored.
```rust
#[cfg(not(any(feature = "uuid", feature = "ulid")))]
{
  true
}
#[cfg(all(feature = "uuid", not(feature = "ulid")))]
{
  format != "Ulid"
}
#[cfg(all(feature = "ulid", not(feature = "uuid")))]
{
  format != "Uuid"
}
```

utoipa-gen/src/schema_type.rs
pbzweihander: Can we have a better code for this? I cannot come up with it.

pbzweihander:
Can you help with the test? I'm not sure how to write one

utoipa-gen/src/schema_type.rs
juhaku: Perhaps we can have this as is, and change it in future.

juhaku:
> Can you help with the test? I'm not sure how to write one

https://github.com/juhaku/utoipa/blob/56b6326e28f0df2a88f5b0572c415d6bd9b8fb1a/utoipa-gen/tests/schema_derive_test.rs#L4707-L4730

Just like in above one but it it should test a type that has the `Ulid` type.

```rust
#[test]
#[cfg(feature = "ullid")]
fn test {
    #[derive(ToSchema)]
    struct Foo {
        value: Ulid,
    }
    
    // ... and so on the assertion
}
```

pbzweihander:
I added test. Thanks!

juhaku (approved):
Great :trophy:

PR #743, opened 2023-08-20
Add support for serde skip in `IntoParams` derive

### LINKED ISSUE #732: `#[serde(skip)]` support for `#[derive(IntoParams]`
What it says on the tin, thanks. Surprised at the omission actually, maybe there was a good reason?

juhaku:
Add support for serde's `skip` attribute in `IntoParams` derive macro. This allows users to use the serde's `skip`, `skip_serializing` or `skip_deserializing` attribute to ignore the field being added as a parameter to a OpenAPI documentation.
```rust
 #[derive(IntoParams, Serialize)]
 #[allow(unused)]
 struct Params {
     name: String,
     name2: Option<String>,
     #[serde(skip)]
     name3: Option<String>,
 }
```

Resolves #732

### DIFF
--- a/utoipa-gen/src/component/into_params.rs
+++ b/utoipa-gen/src/component/into_params.rs
@@ -27,7 +27,7 @@ use super::{
         impl_into_inner, impl_merge, parse_features, pop_feature, pop_feature_as_inner, Feature,
         FeaturesExt, IntoInner, Merge, ToTokensExt,
     },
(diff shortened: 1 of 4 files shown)


VirusBLITZ:
any option to keep a field, for example when I'm only using `#[serde(skip_deserializing)]` but still want the field to show up? (I'm never deserializing it)

PR #747, opened 2023-08-31
Feat url type

rlebran-netwo:
This PR adds support for [url](https://github.com/servo/rust-url).

`Url` type will be presented as `String` with  format `uri` in OpenAPI spec.


juhaku (approved):
Super :tada:  Thanks for the contribution

PR #748, opened 2023-08-31
Feat std::collections::LinkedList as a known field type for schema

vinc3nzo:
One might find `std::collections::LinkedList` more preferable than `Vec`, when defining the return type of an API responder. Hence, it would be great to be able to use `LinkedList` in schema declaration.

This PR registeres `LinkedList` as another `GenericType` and makes aforesaid possible.

I am open to any comments and suggestions.

### DIFF
--- a/utoipa-gen/tests/schema_derive_test.rs
+++ b/utoipa-gen/tests/schema_derive_test.rs
@@ -3571,9 +3571,7 @@ fn derive_component_with_linked_list() {
     use std::collections::LinkedList;
 
     let example_schema = api_doc! {
-        struct ExampleSchema<'a> {
-            id: u64,
-            name: &'a str,
+        struct ExampleSchema {
             values: LinkedList<f64>
         }
     };
@@ -3582,20 +3580,14 @@ fn derive_component_with_linked_list() {
         example_schema,
         json!({
             "properties": {
-                "id": {
-                    "type": "uint64"
-                },
-                "name": {
-                    "type": "string"
-                },
                 "values": {
                     "items": {
-                        "type": "string"
+                        "type": "number"
                     },
                     "type": "array"
                 }
             },
-            "required": ["links"],
+            "required": ["values"],
             "type": "object"
         })
     )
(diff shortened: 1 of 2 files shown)


juhaku (approved):
Awesome, just add one test case to `schema_derive_test.rs` file that tests this `LinkedList`. There are plenty of test's using `Vec` so they can work as an example.

vinc3nzo:
@juhaku, here it is. Take a look.

Also, a bit off-topic, while I was running `cargo test utoipa-gen`, it complained about enum `PathItemType` not having `#[derive(Debug)]` (`utoipa/src/openapi/path.rs` line `699`). Further investigation showed that `PathItemType` derives `Debug` if, and only if, feature `debug` is enabled.

The solution might be to always derive `Debug` for this enum , or to mark the test case that causes the issue with `#[cfg(feature = "debug")]`. The latter is not as groundbreaking.

juhaku:
> Also, a bit off-topic, while I was running `cargo test utoipa-gen`, it complained about enum `PathItemType` not having `#[derive(Debug)]` (`utoipa/src/openapi/path.rs` line `699`). Further investigation showed that `PathItemType` derives `Debug` if, and only if, feature `debug` is enabled.
> 
> The solution might be to always derive `Debug` for this enum , or to mark the test case that causes the issue with `#[cfg(feature = "debug")]`. The latter is not as groundbreaking

Yeah, there is the `./scripts/test.sh` that is better for running the tests with necessary feature flags. Just bluntly running `cargo test -p utoia-gen` will not run all the tests. And using `--all-features` will result errors, because of conflicting implementation for `axum`, `rocket` and `actix_web`. You could run tests individually with custom set of features enabled for your ide as well which is something I usually do. Also just adding necessary features to the `cargo` command is okay e.g. `cargo test --features debug,...x,....y -p utoipa-gen`.

utoipa-gen/tests/schema_derive_test.rs
```diff
@@ -3566,6 +3566,41 @@ fn derive_component_with_raw_identifier() {
...
+            "properties": {
+                "id": {
+                    "type": "uint64"
+                },
+                "name": {
+                    "type": "string"
+                },
+                "values": {
+                    "items": {
+                        "type": "string"
```
juhaku, line 3593: should this type be: `number`?

juhaku: Great

Issue #738, opened 2023-08-16
Manually specify Discriminator Mapping


1lutz:
I need to define a schema which contains a one of with a discriminator. The discrimantor must have a custom mapping.

As I saw [in the docs](https://docs.rs/utoipa/latest/utoipa/openapi/schema/struct.Discriminator.html) the discriminator struct does not have a field for the mapping. But when I let utoipa to automatically generate a schema for a struct which has a serde tag, a mapping will be output. So code for mappings must be there.

Can it be made possible to manually specify a mapping?

jayvdb:
Could you explain more what you are wanting?  Ideally giving a code example.
https://github.com/juhaku/utoipa/blob/b1ce2d0/utoipa-gen/tests/schema_derive_test.rs#L1138 gives an example of how to build the Rust structs to construct a `oneOf`.

If you search that file for `discriminator` you will find about another 8 examples.

jayvdb:
Does utoipa support external schemas? i.e. is this part already not supported?

```yaml
$ref: 'https://gigantic-server.com/schemas/Monster/schema.json'
```

I don't see test cases which show this in action.  Did I miss some?

Now looking at the other subset

```yaml
MyResponseType:
      oneOf:
      - $ref: '#/components/schemas/Cat'
      - $ref: '#/components/schemas/Dog'
      discriminator:
        propertyName: petType
        mapping:
          dog: '#/components/schemas/Dog'
```

What is the benefit here?  Does that mean that `Dog` doesnt need the field `petType` inside it?

Or does it just mean that `petType` can be `dog` instead of `Dog`?

If it is only that the latter, this could be described in Rust by using serde aliases. c.f. https://serde.rs/variant-attrs.html

i.e.
```rust
#[derive(....)]
enum PetType {
    Cat,
    #[serde(alias = "dog")]
    Dog,
}
```

Then utoipa would pick up those aliases and dynamically create the `mapping`.

juhaku:
> Does utoipa support external schemas? i.e. is this part already not supported?

Not yet fully, only in certain case, there was an external ref support for responses if I remember right.

> Or does it just mean that petType can be dog instead of Dog?

Yup.

> If it is only that the latter, this could be described in Rust by using serde aliases....

True, that would work for the general cases.

juhaku:
> I am trying to create OneOfs with a discriminator and a mapping using code. I need that something like this is possible:
> 
> ```rust
> let mut discriminator = Discriminator::new("type");
[...]

PR #752, opened 2023-09-04
Add Discriminator mapping

1lutz:
I need to set the mapping property for the discriminator of oneOfs.

Resolves https://github.com/juhaku/utoipa/issues/738

### DIFF
--- a/utoipa/src/openapi/schema.rs
+++ b/utoipa/src/openapi/schema.rs
@@ -285,7 +285,9 @@ pub struct Discriminator {
     /// objects.
     pub property_name: String,
 
-    // An object to hold mappings between payload values and schema names or references.
+    /// An object to hold mappings between payload values and schema names or references.
+    /// This field can only be populated manually. There is no macro support and no
+    /// validation.
     #[serde(skip_serializing_if = "BTreeMap::is_empty", default)]
     pub mapping: BTreeMap<String, String>,
 }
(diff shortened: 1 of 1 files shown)


jayvdb (commented):
this isnt where the mapping exists.  The `Discriminator` only holds the name of the property which will be used to check whether the alternatives have compatible structures.
I suggest closing this PR and explaining what you need more on the issue.

utoipa/src/openapi/schema.rs
```diff
@@ -284,6 +284,10 @@ pub struct Discriminator {
     /// Defines a discriminator property name which must be found within all composite
     /// objects.
     pub property_name: String,
+
+    // An object to hold mappings between payload values and schema names or references.
```
jayvdb, line 288: Can you change this to triple slash and expand the comment so it is more suitable for public documentation?
It needs to state that it can only be populated manually - i.e. there is no macro support, and there is no validation of the mapping.

jayvdb:
Given the work you are doing at https://github.com/1lutz/geoengine/blob/5e17832833fefc3a0915df4de5ee68bb3c8cee5d/services/src/util/apidoc.rs#L168 , I can see the value of getting this in, even if utoipa doesnt do anything with this mapping field.

And you could grab a subset of that code and convert it to be a unit test quite easily...?

juhaku:
Yeah I agree as I also described  here https://github.com/juhaku/utoipa/issues/738#issuecomment-1720490552. That is unfortunate at the moment that there is a need for such manual processing. 

I think we can let this in without further additions but the doc comment could be improved as @jayvdb mentioned above.

utoipa/src/openapi/schema.rs
1lutz: I edited the comment and added a test here: [c2834c5](https://github.com/juhaku/utoipa/pull/752/commits/c2834c575716219ba4226fc3ac6ed0c1ae0bd01e)

jayvdb (approved):

juhaku (approved):
Super :tada:

PR #757, opened 2023-09-08
Fix: panic on missing trailing / in rocket environment (#645)

### LINKED ISSUE #645: [utopia-swagger-ui] missing trailing / causes thread to panic in rocket environment
Hello, 
I'm currently running into a problem where a thread is panicing, when accessing the swagger-ui path

```
> curl http://localhost:6913/documentation/

<!-- HTML for static distribution bundle build -->
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8">
    <title>Swagger UI</title>
    <link rel="stylesheet" type="text/css" href="./swagger-ui.css" />
    <link rel="stylesheet" type="text/css" href="index.css" />
    <link rel="icon" type="image/png" href="./favicon-32x32.png" sizes="32x32" />
    <link rel="icon" type="image/png" href="./favicon-16x16.png" sizes="16x16" />
  </head>

  <body>
    <div id="swagger-ui"></div>
    <script src="./swagger-ui-bundle.js" charset="UTF-8"> </script>
    <script src="./swagger-ui-standalone-preset.js" charset="UTF-8"> </script>
    <script src="./swagger-initializer.js" charset="UTF-8"> </script>
  </body>
</html>
```  
works, while

```
> curl http://localhost:6913/documentation
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="utf-8">
    <title>500 Internal Server Error</title>
</head>
<body align="center">
    <div role="main" align="center">
        <h1>500: Internal Server Error</h1>
        <p>The server encountered an internal error while processing this request.</p>
        <hr />
    </div>
    <div role="contentinfo" align="center">
        <small>Rocket</small>
    </div>
</body>
</html>
```
crashes with 
```
GET /documentation:
   >> Matched: GET /documentation/<_..>
thread 'rocket-worker-thread' panicked at 'byte index 15 is out of bounds of `/documentation`', /home/rustdev/.cargo/registry/src/github.com-1ecc6299db9ec823/utoipa-swagger-ui-3.1.3/src/rocket.rs:87:29
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
   >> A handler panicked.
   >> This is an application bug.
   >> A panic in Rust must be treated as an exceptional event.
   >> Panicking is not a suitable error handling mechanism.
   >> Unwinding, the result of a panic, is an expensive operation.
   >> Panics will degrade application performance.
   >> Instead of panicking, return `Option` and/or `Result`.
   >> Values of either type can be returned directly from handlers.
   >> A panic is treated as an internal server error.
   >> Outcome: Failure
   >> No 500 catcher registered. Using Rocket default.
   >> Response succeeded.
```

I'm currently using following version of:
rocket = { version = "0.5.0-rc.3", features = ["secrets", "json"]}
utoipa = { version = "3", features = ["debug", "rocket_extras", "yaml"]}
utoipa-swagger-ui = { version = "3.1.3", features = ["rocket"]}

Do you have any what might be causing this behaviour ?
And if so, what could be a possible workaround ?

Thanks.

reflix:
The ServeSwagger handler extracts the base path from the route by taking the beginning of the route until it finds the opening '<' and uses the extracted &str as base path to resolve incoming requests.

Thus, it removes this base path from the request URI and proceeds serving with the corresponding response.

Rocket also routes '/url' (note the missing trailing slash) to the handler which then panics as the base URL is shorter than request URI.

This patch adds a test for the expected base path prefix and response with a redirection if missing resulting in a successful page load.

fixes #645

### DIFF
--- a/utoipa-swagger-ui/src/rocket.rs
+++ b/utoipa-swagger-ui/src/rocket.rs
@@ -2,8 +2,6 @@
 
 use std::{borrow::Cow, io::Cursor, sync::Arc};
 
-use rocket::response::status::Custom;
-use rocket::response::Responder;
 use rocket::{
     http::{Header, Status},
     response::{status::NotFound, Responder as RocketResponder},
@@ -94,7 +92,7 @@ impl Handler for ServeSwagger {
                 .unwrap_or_else(|| Outcome::from(request, NotFound("Swagger UI file not found"))),
             Err(error) => Outcome::from(
                 request,
-                Custom(Status::InternalServerError, error.to_string()),
+                rocket::response::status::Custom(Status::InternalServerError, error.to_string()),
             ),
         }
     }
@@ -111,7 +109,7 @@ impl<'r, 'o: 'r> RocketResponder<'r, 'o> for SwaggerFile<'o> {
 }
 
 struct RedirectResponder(String);
-impl<'r, 'a: 'r> Responder<'r, 'a> for RedirectResponder {
+impl<'r, 'a: 'r> RocketResponder<'r, 'a> for RedirectResponder {
     fn respond_to(self, _request: &'r Request<'_>) -> rocket::response::Result<'a> {
         Response::build()
             .status(Status::Found)
--- a/utoipa-swagger-ui/src/rocket.rs
+++ b/utoipa-swagger-ui/src/rocket.rs
@@ -114,7 +114,7 @@ struct RedirectResponder(String);
 impl<'r, 'a: 'r> Responder<'r, 'a> for RedirectResponder {
     fn respond_to(self, _request: &'r Request<'_>) -> rocket::response::Result<'a> {
         Response::build()
-            .status(Status::MovedPermanently)
+            .status(Status::Found)
             .raw_header("Location", self.0)
             .ok()
     }
--- a/utoipa-swagger-ui/src/rocket.rs
+++ b/utoipa-swagger-ui/src/rocket.rs
@@ -2,12 +2,11 @@
 
 use std::{borrow::Cow, io::Cursor, sync::Arc};
 
+use rocket::response::status::Custom;
+use rocket::response::Responder;
 use rocket::{
     http::{Header, Status},
-    response::{
-        status::{self, NotFound},
-        Responder as RocketResponder,
-    },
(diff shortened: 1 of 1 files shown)


utoipa-swagger-ui/src/rocket.rs
```diff
@@ -79,31 +78,44 @@ struct ServeSwagger(Cow<'static, str>, Arc<Config<'static>>);
...
+            .status(Status::Ok)
+            .finalize())
+    }
+}
+
+struct RedirectResponder(String);
+impl<'r, 'a: 'r> Responder<'r, 'a> for RedirectResponder {
+    fn respond_to(self, _request: &'r Request<'_>) -> rocket::response::Result<'a> {
+        Response::build()
+            .status(Status::MovedPermanently)
```
juhaku, line 117: I'd instead use `Status::Found`. This way it does not get stored to the users browsers that might cause conflicts with other similar endpoints or services. For example if you develop multiple services at your localhost and one creates a permanent redirect but other similar endpoint does not want it to be redirected to the given location but something else instead. 

I believe that the permanent should be used only when some address will nearly forever be the same, but usually these cases the redirect should be "temporary" thus the `Found`.

juhaku (approved):
Looks good, just change the `MovedPermanently` to `Found` :+1:

reflix:
Awesome! Happy to contribute! :)

utoipa-swagger-ui/src/rocket.rs
```diff
@@ -2,12 +2,11 @@
 
 use std::{borrow::Cow, io::Cursor, sync::Arc};
 
+use rocket::response::status::Custom;
+use rocket::response::Responder;
```
jayvdb, line 6: this is imported twice, here as `Responder` and below as `RocketResponder`.  That is confusing.  IMO keep the existing `RocketResponder` and remove this import, and also remove `use rocket::response::status::Custom;` as these are unnecessary changes which change the blame of multiple lines that did not actually change.

reflix: I applied the suggested changes!

jayvdb (approved):

Issue #693, opened 2023-07-20
Looks like format option is ignoring


shandak:
```rust
#[utoipa::path(
    get,
    path = "/visitors/v1/{start_date}/{end_date}",
   ...
    params(
        ("start_date", Path, description = "Start date filter", format = DateTime),
        ("end_date", Path, description = "End date filter", format = Date)
    )
)]
``` 

The description is showing, but the format is always the string
![TinyTake20-07-2023-11-55-54](https://github.com/juhaku/utoipa/assets/3428345/174bc7ad-0f6b-48b9-bb7b-e8d4bc18b5b0)

shandak:
piece of json
```
[
  {
    "name": "start_date",
    "in": "path",
    "description": "Start date filter",
    "required": true,
    "schema": {
      "type": "string"
    }
  },
  {
    "name": "end_date",
    "in": "path",
    "description": "End date filter",
    "required": true,
    "schema": {
      "type": "string"
    }
  }
]
```

shandak:
The same issue if I'm trying to do it via struct
```
#[derive(Deserialize, IntoParams, utoipa::ToSchema)]
pub struct StartEndDate {
    /// Start date filter
    #[schema(value_type = String, format = Date)]
    pub start_date: String,
    /// End date filter
    #[schema(value_type = String, format = Date)]
    pub end_date: String,
}
```
the `format`  is just ignoring in the `path` part, whereas `schema` looks correct

![TinyTake20-07-2023-12-03-45](https://github.com/juhaku/utoipa/assets/3428345/518213e7-eac3-4d2b-bf83-e204e9f7905f)

juhaku:
Interesting, there are some known issues with the `IntoParams` but didn't expect this to be one of them. :eyes: Need to check this same time with other `IntoParams` improvements. 

What framework you are using? Axum? Actix_web? Rocket?

shandak:
actix_web

jayvdb:
Hi @shandak , I added a test in https://github.com/juhaku/utoipa/pull/758 which shows one way of getting these data type formats working in actix-web.

juhaku:
@jayvdb Yeah it seemingly works with the `Query` type of parameters, but not with the `Path` :facepalm: But there is a fix finally for this one in here #928.

PR #758, opened 2023-09-10
Add test for date types in actix params

jayvdb:
Relates to https://github.com/juhaku/utoipa/issues/693

### DIFF
--- a/utoipa-gen/tests/path_parameter_derive_actix.rs
+++ b/utoipa-gen/tests/path_parameter_derive_actix.rs
@@ -164,3 +164,112 @@ fn derive_params_from_method_args_actix_success() {
         "[1].schema.format" = r#"null"#, "Parameter schema format"
     };
 }
+
+#[test]
+fn derive_path_with_date_params_implicit() {
+    mod mod_derive_path_with_date_params {
+        use actix_web::{get, web, HttpResponse, Responder};
+        use chrono::{DateTime, Utc};
+        use serde_json::json;
+        use time::Date;
+
+        #[utoipa::path(
+            responses(
+                (status = 200, description = "success response")
+            ),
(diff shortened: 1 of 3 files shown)


utoipa-gen/tests/path_parameter_derive_actix.rs
```diff
@@ -164,3 +164,113 @@ fn derive_params_from_method_args_actix_success() {
...
+        "[0].schema.type" = r#""string""#, "Parameter schema type"
+        "[0].schema.format" = r#""date""#, "Parameter schema format"
+
+        "[1].in" = r#""path""#, "Parameter in"
+        "[1].name" = r#""end_date""#, "Parameter name"
+        "[1].description" = r#""End date filter""#, "Parameter description"
+        "[1].required" = r#"true"#, "Parameter required"
+        "[1].deprecated" = r#"null"#, "Parameter deprecated"
+        "[1].schema.type" = r#""string""#, "Parameter schema type"
+        "[1].schema.format" = r#"null"#, "Parameter schema format"
```
jayvdb, line 274: note this is probably the problem described in https://github.com/juhaku/utoipa/issues/693 - `format = DateTime` does not override the rust arg being of type `String`.

I am not sure if this is expected / desirable behaviour.

juhaku: Right, I believe this is at least one of the inconsistencies between the without `actix_web` or other framework and with.

I don't remember what was the original thought when implementing this but now I could consider this as undesired behavior. I reckon the order of recognizing parameters should first come from fn arguments if framework is supported and later they should be overridable with manual `params(...)` definition.

juhaku (approved):
:+1: Great work

PR #759, opened 2023-09-13
Path impl_for override. PathBuilder::path_from

wyatt-herkamp:
I personally do like the derive OpenAPI macro

So I tend to implement [OpenAPI](https://docs.rs/utoipa/3.5.0/utoipa/trait.OpenApi.html) by hand.

However it generates a pretty ugly to look at type `__path_{function_name}` I use Actix so Actix already generates a type for my route. So I just want the path macro to implement for that type.

## Example

```rust
#[utoipa::path(get,  
    impl_for = me,  
    path = "/api/me",  
    responses(  
        (status = 200, description = "You are Logged In", body = User),  
        (status = 401, description = "You are not logged in")  
    ),  
    security(  
        ("api_key" = [])  
    )  
)]  
#[get("")]  
pub async fn me(auth: Authentication) -> JsonResponse<User> {  
    JsonResponse::from(Into::<User>::into(auth))  
}  
```

Then I added a new function to PathBuilder path\_from. Just for a cleaner way of appending a path.

### DIFF
--- a/utoipa-gen/src/lib.rs
+++ b/utoipa-gen/src/lib.rs
@@ -673,6 +673,10 @@ pub fn derive_to_schema(input: TokenStream) -> TokenStream {
 ///
 /// * `path = "..."` Must be OpenAPI format compatible str with arguments within curly braces. E.g _`{id}`_
 ///
+///
+/// * `impl_for = ...` Optional type to implement the [`Path`][path] trait. By default a new type
+///   is used for the implementation.
+///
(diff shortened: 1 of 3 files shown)


juhaku (approved):

juhaku:
> However it generates a pretty ugly to look at type `__path_{function_name}`

It generates ugly to look so that it does not name clash even possibly with any other type that might be generated. However as it is a generated type it does not really matter what it looks like though. But I can see why somebody would save a few bytes without generating an "extra" type. Yet this actually could open a door for path support in traits as well.

PR #763, opened 2023-09-18
add openapi extensions "x-tokenName"

lucasgranberg:
Setting x-tokenname makes it possible to choose which token to use. This is useful for OICD and non-standard oauth servers. ex. if you want to use id_token from google instead of auth_token.

https://github.com/swagger-api/swagger-ui/issues/7561

For some reason the Debug derive was missing from PathItemType which made the build fail

### DIFF
--- a/utoipa/src/openapi/info.rs
+++ b/utoipa/src/openapi/info.rs
@@ -6,6 +6,8 @@
 //! [info]: <https://spec.openapis.org/oas/latest.html#info-object>
 //! [openapi_trait]: ../../trait.OpenApi.html
 //! [derive]: ../../derive.OpenApi.html
+use std::collections::HashMap;
+
 use serde::{Deserialize, Serialize};
 
 use super::{builder, set_value};
@@ -66,6 +68,10 @@ builder! {
 
         /// Document version typically the API version.
         pub version: String,
+
+        /// Optional extensions "x-something"
+        #[serde(skip_serializing_if = "Option::is_none", flatten)]
+        pub extensions: Option<HashMap<String, serde_json::Value>>,
     }
 }
 
@@ -119,6 +125,11 @@ impl InfoBuilder {
     /// Add license of the API.
     pub fn license(mut self, license: Option<License>) -> Self {
         set_value!(self license license)
+    }
+
+    /// Add openapi extensions (x-something) of the API.
+    pub fn extensions(mut self, extensions: Option<HashMap<String, serde_json::Value>>) -> Self {
+        set_value!(self extensions extensions)
     }
 }
 
--- a/utoipa/src/openapi/path.rs
+++ b/utoipa/src/openapi/path.rs
@@ -1,7 +1,7 @@
 //! Implements [OpenAPI Path Object][paths] types.
 //!
 //! [paths]: https://spec.openapis.org/oas/latest.html#paths-object
-use std::iter;
+use std::{collections::HashMap, iter};
 
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
@@ -34,6 +34,10 @@ builder! {
         /// Map of relative paths with [`PathItem`]s holding [`Operation`]s matching
         /// api endpoints.
         pub paths: PathsMap<String, PathItem>,
+
+        /// Optional extensions "x-something"
+        #[serde(skip_serializing_if = "Option::is_none", flatten)]
+        pub extensions: Option<HashMap<String, serde_json::Value>>,
     }
 }
 
@@ -97,6 +101,11 @@ impl PathsBuilder {
 
         self
     }
+
+    /// Add extensions to the paths section.
+    pub fn extensions(mut self, extensions: Option<HashMap<String, serde_json::Value>>) -> Self {
+        set_value!(self extensions extensions)
+    }
 }
 
 builder! {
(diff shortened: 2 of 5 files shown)


lucasgranberg:
This is probably handled better by serde_with and prefix or some other way
https://docs.rs/serde_with/latest/serde_with/macro.with_prefix.html

This way utoipa could support any openapi extension
https://swagger.io/docs/specification/openapi-extensions/

lucasgranberg:
I changed it to use serde Value instead. If we want to force the x- prefix we would have to use something like serde_with but that adds additional dependencies.

The new way can be used like
```rs
impl Modify for SecurityAddon {
    fn modify(&self, openapi: &mut utoipa::openapi::OpenApi) {
        if let Some(components) = openapi.components.as_mut() {
            let mut oauth = OAuth2::new([Flow::AuthorizationCode(AuthorizationCode::new(
                "https://accounts.google.com/o/oauth2/auth",
                "https://oauth2.googleapis.com/token",
                Scopes::from_iter([("email", "use your email address"), ("openid", "openid")]),
            ))]);
            oauth.set_extensions(Some(Value::Object(Map::from_iter([(
                "x-tokenName".to_string(),
                Value::String("id_token".to_string()),
            )]))));
            components.add_security_scheme("token", SecurityScheme::OAuth2(oauth));
        }
    }
}
            ```

**Comment by lucasgranberg** [comment]
>For some reason the Debug derive was missing from PathItemType which made the build fail
I missed it was behind a feature flag

**Comment by juhaku** [comment]
> > For some reason the Debug derive was missing from PathItemType which made the build fail
> 
> I missed it was behind a feature flag

Yeah, thats fine :slightly_smiling_face:

**Comment by juhaku** [comment]
I'd create a test for this case according to what you suggested as a way to use the field. You can find references for how to test this from `schema.rs` and `openapi.rs` files at `utoipa` crate. 

> I changed it to use serde Value instead. If we want to force the x- prefix we would have to use something like serde_with but that adds additional dependencies.
> 
> The new way can be used like
> 
> ```rust
> impl Modify for SecurityAddon {
>     fn modify(&self, openapi: &mut utoipa::openapi::OpenApi) {
>         if let Some(components) = openapi.components.as_mut() {
>             let mut oauth = OAuth2::new([Flow::AuthorizationCode(AuthorizationCode::new(
>                 "https://accounts.google.com/o/oauth2/auth",
>                 "https://oauth2.googleapis.com/token",
>                 Scopes::from_iter([("email", "use your email address"), ("openid", "openid")]),
>             ))]);
>             oauth.set_extensions(Some(Value::Object(Map::from_iter([(
>                 "x-tokenName".to_string(),
>                 Value::String("id_token".to_string()),
>             )]))));
>             components.add_security_scheme("token", SecurityScheme::OAuth2(oauth));
>         }
>     }
> }
> ```

**Comment by juhaku** [comment]
Also this PR states that it adds support for extensions, but perhaps this PR should point out that it only does it for security schema or this could also add the same support for the other places as well which are bound to support extensions.

**Comment by lucasgranberg** [comment]
> we could use HashMap<String, serde_json::Value> or just boldly HashMap<String, String>

I think `HashMap<String, serde_json::Value>` could work. Even the openapi specs example talks about a nested structure for extensions
https://swagger.io/docs/specification/openapi-extensions/

> Also this PR states that it adds support for extensions, but perhaps this PR should point out that it only does it for security schema or this could also add the same support for the other places as well which are bound to support extensions.

I agree it should be supported everywhere. Ill have to see how that could be done. Maybe some kind of extendable trait. Per the specs

>Extensions are supported on the root level of the API spec and in the following places:
>
>    info section
>    paths section, individual paths and operations
>    operation parameters
>    responses
>    tags
>    security schemes

>I'd create a test for this case according to what you suggested as a way to use the field.

That makes sense to me, Ill do that last when we have agreed on an approach. I should have made this a draft PR

**Comment by lucasgranberg** [comment]
Allright! Added the extensions in all places required by the spec. I went with HashMap<String, serde_json::Value>. I also removed the setter from security as I see you use pub instead and methods in the builders.

 I still have to figure out how to add this to the docs.

**Comment by juhaku** [comment]
> Allright! Added the extensions in all places required by the spec. I went with HashMap<String, serde_json::Value>. I also removed the setter from security as I see you use pub instead and methods in the builders.
> 
> I still have to figure out how to add this to the docs.

Nice,

You can build the docs just with `./script/doc.sh` script and then open the docs in your browser. The generated files will be located in `target` directory. There are some html files namely for browsing the docs. You need to have nightly target installed for the docs to build. 

Docs are formed from the `///` doc comments within the `utoipa` crate.

**Comment by lucasgranberg** [comment]
The /// docs I created are on par with what existed before for the builders. I don't know where to put an example for how to use extensions but maybe it is easy enough if people read the code.

This is how I use it

```rs
#[derive(Clone)]
struct SecurityAddon;

impl Modify for SecurityAddon {
    fn modify(&self, openapi: &mut utoipa::openapi::OpenApi) {
        if let Some(components) = openapi.components.as_mut() {
            let mut oauth = OAuth2::new([Flow::AuthorizationCode(AuthorizationCode::new(
                "https://accounts.google.com/o/oauth2/auth",
                "https://oauth2.googleapis.com/token",
                Scopes::from_iter([("email", "use your email address"), ("openid", "openid")]),
            ))]);
            oauth.extensions = Some(HashMap::from_iter([(
                "x-tokenName".to_string(),
                Value::String("id_token".to_string()),
            )]));
            components.add_security_scheme("token", SecurityScheme::OAuth2(oauth));
        }
    }
}
```

lucasgranberg:
@juhaku ready to merge imo

juhaku:
> @juhaku ready to merge imo

Oh, yeah, sorry for the latency, Sure let's merge this one, the possible example can be figured out later.

PR #799, opened 2023-11-13
Support `#[serde(flatten)]` for maps.

blp:
`utoipa` already supported `#[serde(flatten)]` on fields with structure type, by putting the fields inside those structures into the parent type. This is commonly used for factoring out frequently used keys, as documented at <https://serde.rs/attr-flatten.html#factor-out-frequently-grouped-keys>.

`#[serde(flatten)]` has another use that utoipa does not support: to capture additional unnamed fields within a structure, as documented at <https://serde.rs/attr-flatten.html#capture-additional-fields>.  This commit adds support for that functionality as well as a pair of tests. It only makes sense to have one such field per structure, so this commit reports an error if there is more than one.


blp:
Thank you!

PR #828, opened 2023-12-29
Generate embed code instead of using interpolation

helaan:
Because the output path of the embedded files changes per machine, the interpolate-folder-path feature of rust-embed was used to set the correct embed path. This feature however creates a rather large dependency tree (shell_expand -> dirs -> dirs-sys -> option-ext)

By generating a small file with the correct folder already filled in, then including it in lib.rs we can remove this feature and reduce the amount of dependencies that are used.

Implementation was inspired by how the "built" crate recommends people to use it.


juhaku (approved):
Pretty neat :+1:

PR #858, opened 2024-02-07
Add example: simple-axum

vi:
There is already "todo-axum" example, but it is rather long and includes UI and other advanced things.

I think an additional, short and simple example should useful.


juhaku (approved):
Sure why not

PR #867, opened 2024-02-14
Refactor `RapiDoc` to take `Cow<'static, str>` instead of borrowed `str`

### LINKED ISSUE #805: How to dynamically set `spec_url` in RapiDoc?
I'm wondering how I can dynamically set the `spec_url` for RapiDoc:

This works fine: 

```rust
use actix_web::{HttpServer, App}; 
use utoipa_rapidoc::Rapidoc;

HttpServer::new(move || {
    App::new()
        .service(RapidDoc::new("/api-docs/openapi.json"))
})
...
```

But what if I want to do the following?: 

```rust
use actix_web::App; 
use utoipa_rapidoc::Rapidoc;

let base_url = "/some/path";
let spec_url = format!("{}/api-docs/openapi.json", base_url);

HttpServer::new(move || {
    App::new()
        .service(RapiDoc::new(&spec_url));
})
```

The fact that I have to pass a `&str` to `RapiDoc::new` makes this impossible!

simongoricar:
> This PR fixes #805

# Problem
As far as I can tell, it was previously impossible to dynamically construct URLs or custom HTML in certain cases. This was due to `RapiDoc`'s `path`, `spec_url` and `html` fields being `&'_ str`.

Not bad in principle of course, but a problem arose when trying to use this in combination with `actix_web` (possibly other integrations, but I haven't tried). `actix_web`'s [`HttpServer::new`](https://docs.rs/actix-web/latest/actix_web/struct.HttpServer.html#method.new) requires that you pass a factory closure with the following conditions: `Fn() -> I + Send + Clone + 'static`, and inside of that closure we need to initialize `RapiDoc`. As such, the issue is that we can't dynamically construct a `'static` string in these conditions ([excluding quite terrible practices](https://users.rust-lang.org/t/how-to-make-a-static-str-from-a-variable/53718/3)).

For example, this would not compile beforehand (in this case due to `temporary value dropped while borrowed`):
```rust
use actix_web::{App, HttpServer};
use utoipa_rapidoc::RapiDoc;

fn main() {
    let base_url = "/some/path";
    let spec_url = format!("{}/api-docs/openapi.json", base_url);

    HttpServer::new(move || App::new().service(RapiDoc::new(spec_url.clone().as_str())));
    // ...
}
```

# Solution
Instead of having `RapiDoc`'s methods take `&'_ str`, I modified them to take `Into<Cow<'static, str>>` and modified the `path`, `spec_url` and `html` fields to be `Cow<'static, str>`. This is also how this is handled in the adjacent `utoipa_swagger_ui` crate.

This example with a dynamic URL now compiles:
```rust
use actix_web::{App, HttpServer};
use utoipa_rapidoc::RapiDoc;

fn main() {
    let base_url = "/some/path";
    let spec_url = format!("{}/api-docs/openapi.json", base_url);

    HttpServer::new(move || App::new().service(RapiDoc::new(spec_url.clone())));
    // ...
}
```

---

## Breaking changes
As far as the methods go, I don't think this has any breaking changes due to the `Into<Cow<'static, str>>` generic. **However**, I think the removal of the three lifetime parameters on `RapiDoc` can break some code.

### DIFF
--- a/utoipa-rapidoc/src/lib.rs
+++ b/utoipa-rapidoc/src/lib.rs
@@ -133,6 +133,8 @@
 //! [examples]: <https://github.com/juhaku/utoipa/tree/master/examples>
 //! [rapidoc_quickstart]: <https://rapidocweb.com/quickstart.html>
 
+use std::borrow::Cow;
+
 const DEFAULT_HTML: &str = include_str!("../res/rapidoc.html");
 
 /// Is [RapiDoc][rapidoc] UI.
@@ -144,16 +146,16 @@ const DEFAULT_HTML: &str = include_str!("../res/rapidoc.html");
 /// [rapidoc]: <https://rapidocweb.com>
 /// [standalone]: index.html#using-standalone
 #[non_exhaustive]
-pub struct RapiDoc<'u, 's, 'h> {
+pub struct RapiDoc {
     #[allow(unused)]
-    path: &'u str,
-    spec_url: &'s str,
-    html: &'h str,
+    path: Cow<'static, str>,
+    spec_url: Cow<'static, str>,
+    html: Cow<'static, str>,
     #[cfg(any(feature = "actix-web", feature = "rocket", feature = "axum"))]
     openapi: Option<utoipa::openapi::OpenApi>,
 }
 
-impl<'u, 's, 'h> RapiDoc<'u, 's, 'h> {
+impl RapiDoc {
     /// Construct a new [`RapiDoc`] that points to given `spec_url`. Spec url must be valid URL and
     /// available for RapiDoc to consume.
     ///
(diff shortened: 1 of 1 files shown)


simongoricar:
Now that I think about it, I think it's possible to use `Cow<'a, str>`'s instead of `'static`s. This would eliminate breaking changes as the three lifetimes would still be present on `RapiDoc` (one for each field, as before) *and* we could still pass in owned strings.

Let me know if you'd prefer manual lifetimes instead of `'static`s. It might make sense to refactor other UI crates to the same idea if we go that route, though.

juhaku (approved):
Nice update, probably something that should be done for the Redoc if not already there. :+1: But that is another thing for another time. Thanks :slightly_smiling_face:

juhaku:
I think this is just fine with the `'static` unless there is someone needing the manual lifetimes. Manual lifetimes might sometimes be quite annoying.

simongoricar:
> Nice update, probably something that should be done for the Redoc if not already there. 👍 But that is another thing for another time. Thanks 🙂

Sure, why not. I'll submit another PR :)

PR #869, opened 2024-02-19
Refactor `ReDoc` to take `Cow<'static, str>` instead of borrowed `str`

simongoricar:
Same thing as #867, except I modified `ReDoc` this time. 

### Breaking changes
Just like #867: I think the removal of the two lifetime parameters on `ReDoc` can break some code (e.g. when implementing custom traits for `ReDoc`). As far as the method parameters are concerned, I don't think there are any breaking changes due to the `Into<Cow<'static, str>` generic.

### DIFF
--- a/utoipa-rapidoc/src/lib.rs
+++ b/utoipa-rapidoc/src/lib.rs
@@ -133,6 +133,8 @@
 //! [examples]: <https://github.com/juhaku/utoipa/tree/master/examples>
 //! [rapidoc_quickstart]: <https://rapidocweb.com/quickstart.html>
 
+use std::borrow::Cow;
+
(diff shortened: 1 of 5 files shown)


juhaku (approved):
Great, :+1:

juhaku:
> Just like https://github.com/juhaku/utoipa/pull/867: I think the removal of the two lifetime parameters on `ReDoc` can break some code (e.g. when implementing custom traits for `ReDoc`). As far as the method parameters are concerned, I don't think there are any breaking changes due to the `Into<Cow<'static, str>` generic.

That is true, but the changes are low, but sure when the release is done this is going to be a major release. :slightly_smiling_face:  Just to be explisit about possibility of stuff breaking.

PR #870, opened 2024-02-19
Add new GitHub Actions job to check that examples compile and are properly formatted

simongoricar:
This PR adds a new GitHub Actions job called `test (examples)` to the existing `Utoipa build` workflow. The job runs alongside the existing `test (...)` jobs, which means it will run on push or on a PR to master.

It validates all examples in the `examples` directory by checking that they are properly formatted (with `cargo fmt --check`) and that they can be compiled (with `cargo clippy --all-features --all-targets --workspace --quiet`). 

I've modified the caching key for the new job so that it will be separate from other jobs, since this cache is likely to be quite large due to several full-fledged web frameworks being included, and I assume downloading it each time would slow down other tests that don't need it.


simongoricar:
Oops, looks like one example requires nightly and the script fails if nightly isn't already installed.

simongoricar:
Alright, the nightly problem is fixed. Fingers crossed that the caching works as well.

juhaku (approved):
:1st_place_medal: :trophy: Great addition, this really useful and been wanting to have but it has just been in the back burner.

PR #876, opened 2024-02-26
feat: allow for multiple req body content_type

### LINKED ISSUE #896: Is it possible to accept multiple content-types for requests?
OpenAPI 3 allows describing multiple accepted content-types for a request ([first example on this link](https://swagger.io/docs/specification/describing-request-body/)).

Is it possible using utoipa?

I'd like to make an endpoint accepting both `application/json` and `application/x-www-form-urlencoded`. I implemented it in Actix using two handlers and guards, but I don't see how to document it in utoipa.

TroyKomodo:
This change adds the ability to specify multiple accepted request body content-types.

Resolves #896

### DIFF
--- a/utoipa-gen/src/lib.rs
+++ b/utoipa-gen/src/lib.rs
@@ -362,7 +362,7 @@ use self::{
 /// where super type declares common code for type aliases.
 ///
 /// In this example we have common `Status` type which accepts one generic type. It is then defined
-/// with `#[aliases(...)]` that it is going to be used with [`String`](std::string::String) and [`i32`] values.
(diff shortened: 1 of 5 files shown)


TroyKomodo:
@juhaku any updates on this PR?

juhaku:
Seems like a format is not correct, you should first run `cargo fmt` for the project.

juhaku (changes requested):
Looks good, but two things need to be addressed. :slightly_smiling_face: 
1. Fix format.
2. The docs need to be updated with the changed functionality of the content type support for the request body. The correct place to change the functionality is in `utoipa-gen/src/lib.rs` and you can find it under the `utoipa::path` macro request body heading.

Issue #439, opened 2023-01-16
Custom path description


PierreNeumann:
Hi, 

I have some code documentation for my handler that I don't want to appear in the path description field.
I see that the path description field is automatically filled with everything having the doc attribute. 

How can I set a custom description for the API endpoint so the user don't see all the code documentation ?

juhaku:
Hey @PierreNeumann ,

Unfortunately this is not possible at the moment via `#[utoipa::path]` macro. Only way to do this now is manually edit the generated OpenApi. This is not very elegant though especially if there is many such endpoints. One way to do this is though implement path manually for the OpenApi and with that creating a custom macro rules to help with boilerplate. 

What kind of functionality you had in mind? E.g. it would be quite easy to implement functionality for users to define custom `description = ...` attribute in utoipa path macro as seen below. But would it be sufficient? 
````rust
#[utoipa::path(
    description = "This is description override",
)]
````

**Comment by PierreNeumann** [comment]
Hi, 

I found a workaround. 
I created another function that contain all my handler logic and the "private" documentation. 
Then the handler call this function and is documented with the public utoipa documentation.

Example : 
`/// Do some work on a JSON Message
/// 
/// The public documentation I want to show on the swagger ui.
#[utoipa::path(
    post,
    path = "/work",
    responses(
        (status = 200, description = "Work done !", body = String)
    ),
    request_body = Message,
)]
pub async fn handler(
    item: web::Json<Message>,
    ) -> impl Responder {
    do_work(item)
}

/// Work on Json Message
///
/// All private documentation I don't want to show on the swagger ui.
/// 
/// # Arguments
/// * `item` - The JSON Message
///
/// # Example
/// ```bash
/// curl -X POST http://localhost:8080/work -d "Hello World"
/// ```
pub fn do_work(
    item: web::Json<Message>,
) -> HttpResponse {

    /*  */
    let res = work(item);

    match res {
        Ok(_) => HttpResponse::Ok().json("Work done !".to_string()),
        Err(_e) => {
            HttpResponse::InternalServerError().body("Error working on content".to_string())
        }
    }
}`

**Comment by juhaku** [comment]
This is nice, and probalby most feasible solution.

**Comment by nacairns1** [comment]
Hey I wanted to bump this. I think having a custom Summary and Description via the macro attributes would be very nice and consistent with the rest of the API. That would look something like this:
```rust 
#[utoipa::path(
    description = "This is description override",
    summary = "This is summary override",
)]
```

[...]

PR #881, opened 2024-03-03
Skip first doc comment line in path macro description expansion 

edward-shen:
Previously, the entire doc comment was used as the description in addition to the summary using the first line of the doc comment, resulting in duplicated text.

This commit changes the path macro doc comment extraction logic to split the first line from the rest of the body, and skip all lines until the first non-whitespace line. This results in the description no longer containing the summary.

In degenerate cases, such as a doc comment consisting of new lines, this will invoke a full linear search over all lines in the doc comment. This should be relatively acceptable as most reasonable forms of doc comments shouldn't have more than a couple of newlines between the first line and subsequent body.

For testing and verification, I've tested this in a private project using the `patch` override section in `Cargo.toml`, and tests needed to be changed to account for this change.

This change was driven by my own preferences in documentation and in addition to [this comment](https://github.com/juhaku/utoipa/issues/439#issuecomment-1504081544) by @chrsteer:

> I was also looking for a way to put the summary and description inside the path macro, because it bothered me that the first line was included in both fields.

### DIFF
--- a/utoipa-gen/src/lib.rs
+++ b/utoipa-gen/src/lib.rs
@@ -657,12 +657,12 @@ pub fn derive_to_schema(input: TokenStream) -> TokenStream {
 /// `#[deprecated  = "There is better way to do this"]` the reason would not render in OpenAPI spec.
 ///
 /// Doc comment at decorated function will be used for _`description`_ and _`summary`_ of the path.
-/// First line of the doc comment will be used as the _`summary`_ and the whole doc comment will be
+/// First line of the doc comment will be used as the _`summary`_ while the remaining lines will be
 /// used as _`description`_.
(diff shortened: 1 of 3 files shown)


edward-shen:
Looks like CI failure is related to not providing a feature flag? I don't believe that it's related to these changes.

junlarsen:
Yeah, seems like utoipa-gen requires features query+json to compile (https://github.com/juhaku/utoipa/pull/893)

juhaku:
Yeah, probably some upstream crate feature changes that hasn't been propagated here has caused the error.

juhaku (approved):
Great, :tada: Thanks for the contribution.

PR #889, opened 2024-04-10
Replace `zip` with `zip_next`

Pr0methean:
The `zip` crate hasn't been updated in almost a year, and has a number of bugs that can cause panics when trying to process an invalid zip file. This PR replaces it with `zip_next`, a fork I actively maintain.


juhaku:
Sure

PR #893, opened 2024-04-15
Implement include_str! for tags

### LINKED ISSUE #877: Use include_str! in the openapi's tag definition
Would it be possible to use the `include_str!` macro inside `openapi` derive macro `tag` attribute, like it is possible with the `info` attribute?

```rust
#[derive(OpenApi)]
#[openapi(
    info(description=include_str!("../documentation/description.md")),    // <--- This works
    paths(get_channels, create_channel),
    components(schemas(crate::routes::Channel, crate::routes::CreateChannelRequest)),
    tags(
        (name="My Tag", description=include_str!("../documentation/channels.md"))   // <--- This fails
    ),
)]
pub struct ApiDoc;
```
This is the error I got:
```
error: expected string literal
 --> src/openapi.rs:9:39
  |
9 |         (name="Channels", description=include_str!("../documentation/channels.md"))
  |
```

Looking quickly at the sources, it seems that Info Description uses a certain Type that allow both `String` and `include_str!`. Would it be possible to reuse this Type here?

junlarsen:
Fixes #877 

Moves the Str enum into parse_utils and wraps with parse_next_literal_str_or_include_str().

Also includes a test-case to test the new functionality

### DIFF
--- a/utoipa-gen/Cargo.toml
+++ b/utoipa-gen/Cargo.toml
@@ -27,7 +27,7 @@ utoipa = { path = "../utoipa", features = ["debug", "uuid"], default-features = 
 serde_json = "1"
 serde = "1"
 actix-web = { version = "4", features = ["macros"], default-features = false }
-axum = { version = "0.7", default-features = false }
+axum = { version = "0.7", default-features = false, features = ["json", "query"] }
 paste = "1"
 rocket = { version = "0.5", features = ["json"] }
 smallvec = { version = "1.10", features = ["serde"] }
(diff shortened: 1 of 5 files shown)


utoipa-gen/Cargo.toml
```diff
@@ -27,7 +27,7 @@ utoipa = { path = "../utoipa", features = ["debug", "uuid"], default-features =
 serde_json = "1"
 serde = "1"
 actix-web = { version = "4", features = ["macros"], default-features = false }
-axum = { version = "0.7", default-features = false }
+axum = { version = "0.7", default-features = false, features = ["json", "query"] }
```
junlarsen, line 30: Doctests failed to compile without these

juhaku: Nice, it might be that there is some upstream changes that hasn't been updated here previously.

juhaku (approved):
Solid work :+1:

PR #910, opened 2024-05-06
Have OpenApi::merge consider operations/methods as well as paths

### LINKED ISSUE #907: Shouldn't `merge` take Operations into account?
https://github.com/juhaku/utoipa/blob/19fb23ccb57dbb3fd9b103963464d6a4a6c2157a/utoipa/src/openapi.rs#L182

It seems like I should be able to merge a doc with the same path as long as the operation doesn't conflict. I should be able to have `/api/v1/foo` for both a POST and a GET, right?

jcrossley3:
Fixes #907

### DIFF
--- a/utoipa/src/openapi.rs
+++ b/utoipa/src/openapi.rs
@@ -176,11 +176,12 @@ impl OpenApi {
         }
 
         if !other.paths.paths.is_empty() {
-            other
-                .paths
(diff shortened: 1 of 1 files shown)


juhaku (approved):
Nice :+1:  Let's get this merged when the CI finishes

PR #917, opened 2024-05-08
Use `minreq` instead of `reqwest`

aumetra:
The latest version of `utoipa-swagger-ui` introduced a dependence on `reqwest` at build-time just to download a single file.  

That feels unnecessarily heavy, considering it's intended to be a one-off download thing, and `reqwest` pulls in the entirety of hyper, tokio, mio, and all supporting libraries of that.

This PR replaces it with `minreq` which has a smaller dependency tree.  
The difference here is that `minreq` buffers the entire body into memory, but I checked the size of the archive and 4MB seem like a fine amount to buffer in memory.

If it isn't, then we can switch to `ureq` which offers an `io::Read` based interface, so it streams the response (but it is *a little* heavier on dependencies than `minreq`, but not by much).

### DIFF
--- a/utoipa-swagger-ui/Cargo.toml
+++ b/utoipa-swagger-ui/Cargo.toml
@@ -32,6 +32,6 @@ features = ["actix-web", "axum", "rocket"]
 rustdoc-args = ["--cfg", "doc_cfg"]
 
 [build-dependencies]
-zip = { version = "1", default-features = false, features = ["deflate"] }
+minreq = { version = "2.11.2", features = ["https-rustls"] }
 regex = "1.7"
-reqwest = { version = "0.12", features = ["blocking", "rustls-tls"], default-features = false }
+zip = { version = "1", default-features = false, features = ["deflate"] }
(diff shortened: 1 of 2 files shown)


juhaku:
I am not fond to accept a crate as a dependency that is not licensed either with MIT or Apache-2.0 or both of them. Also as what comes to the functionalities it is unfortunate that using async rust based crate bring a lot of dependencies. However like the reqwest crate https://docs.rs/reqwest/latest/reqwest/#proxies, the library should support proxies, more over the system proxies that are set with `HTTP_PROXY` and `HTTPS_PROXY` environment variables. This is fundamental because some people might be working from within corporation network that is behind a VPN where all the internet traffic is routed through a custom HTTP(S) proxy. If the crate does not support these by default then those users will face frustration when they cannot build their crates.

aumetra:
That would then be ureq. Is MIT/Apache Dual licensed and can read proxies from the environment.

Just one sidenote: the fact that anything is downloaded here at all breaks building this crate via Nix since you can't just access the network while Nix builds a crate, since that would potentially break reproducibility.

juhaku:
> Just one sidenote: the fact that anything is downloaded here at all breaks building this crate via Nix since you can't just access the network while Nix builds a crate, since that would potentially break reproducibility.

Yeah related to that there is ongoing discussion  here https://github.com/juhaku/utoipa/pull/923.

PR #923, opened 2024-05-14
Make SWAGGER_UI_DOWNLOAD_URL support file:// urls

### LINKED ISSUE #922: This PR also breaks sandboxed builds that deny network access, sadly.
This PR also breaks sandboxed builds that deny network access, sadly.

[Nix](https://nixos.org/) builds typically have no external network access, instead letting Nix download files itself and injecting them into the build. Typically you'd do this with a tool like crate2nix, but that is only aware of Cargo dependencies, not external downloads like this.

IMO, the ideal would be to either:

1. Add a `vendored` feature (this could pull in a separate `utoipa-swagger-ui-vendored` crate to avoid bloating the main crate archive),
2. Make `SWAGGER_UI_DOWNLOAD_URL` support `file://` URLs as well, or
3. Add a separate `SWAGGER_UI_ZIP=/path/to/the/zip` environment variable

_Originally posted by @nightkr in https://github.com/juhaku/utoipa/issues/845#issuecomment-2107585774_

oscar6echo:
Cf. merged PR https://github.com/juhaku/utoipa/pull/845 for context.

Resolves #922

### DIFF
--- a/utoipa-swagger-ui/build.rs
+++ b/utoipa-swagger-ui/build.rs
@@ -11,7 +11,8 @@ use zip::{result::ZipError, ZipArchive};
 
 /// the following env variables control the build process:
 /// 1. SWAGGER_UI_DOWNLOAD_URL:
-/// + the url from where to download the swagger-ui zip file
+/// + the url from where to download the swagger-ui zip file if starts with http:// or https://
+/// + the file path from where to copy the swagger-ui zip file if starts with file://
 /// + default value is SWAGGER_UI_DOWNLOAD_URL_DEFAULT
 /// + for other versions, check https://github.com/swagger-api/swagger-ui/tags
 /// 2. SWAGGER_UI_OVERWRITE_FOLDER
@@ -32,10 +33,18 @@ fn main() {
     let zip_path = [&target_dir, &zip_filename].iter().collect::<PathBuf>();
 
     if !zip_path.exists() {
-        println!("start download to : {:?}", zip_path);
-        download_file(&url, zip_path.clone()).unwrap();
+        if url.starts_with("http://") || url.starts_with("https://") {
+            println!("start download to : {:?}", zip_path);
+            download_file(&url, zip_path.clone()).unwrap();
+        } else if url.starts_with("file://") {
+            let file_path = url.replace("file://", "");
+            println!("start copy to : {:?}", zip_path);
+            fs::copy(file_path, zip_path.clone()).unwrap();
+        } else {
+            panic!("invalid SWAGGER_UI_DOWNLOAD_URL: {} -> must start with http:// | https:// | file://", url);
+        }
     } else {
-        println!("already downloaded: {:?}", zip_path);
+        println!("already downloaded or copied: {:?}", zip_path);
     }
 
     println!("cargo:rerun-if-changed={:?}", zip_path.clone());
--- a/utoipa-swagger-ui/README.md
+++ b/utoipa-swagger-ui/README.md
@@ -55,13 +55,14 @@ The following configuration env variables are available at build time:
 
 * `SWAGGER_UI_DOWNLOAD_URL`:
 
-  * the url from where to download the swagger-ui zip file
+  * the url from where to download the swagger-ui zip file if starts with `http://` or `https://`
+  * the file path from where to copy the swagger-ui zip file if starts with `file://`
   * default value: <https://github.com/swagger-api/swagger-ui/archive/refs/tags/v5.17.3.zip>
-  * All versions: <https://github.com/swagger-api/swagger-ui/tags>
+    * all versions: <https://github.com/swagger-api/swagger-ui/tags>
 
 * `SWAGGER_UI_OVERWRITE_FOLDER`:
 
-  * absolute path to a folder containing files to overwrite the default swagger-ui files
+  * absolute path to a folder containing files to overwrite the swagger-ui files extracted from the `.zip` file
   * typically you might want to overwrite `index.html`
 
 ## Examples


juhaku (approved):

utoipa-swagger-ui/build.rs
```diff
@@ -32,10 +33,18 @@ fn main() {
...
     if !zip_path.exists() {
-        println!("start download to : {:?}", zip_path);
-        download_file(&url, zip_path.clone()).unwrap();
+        if url.starts_with("http://") || url.starts_with("https://") {
+            println!("start download to : {:?}", zip_path);
+            download_file(&url, zip_path.clone()).unwrap();
+        } else if url.starts_with("file://") {
+            let file_path = url.replace("file://", "");
+            println!("start copy to : {:?}", zip_path);
+            fs::copy(file_path, zip_path.clone()).unwrap();
```
nightkr, line 42: No need to copy the file here, right? Wouldn't it be enough to set `zip_path = file_path;`?

utoipa-swagger-ui/build.rs
```diff
@@ -32,10 +33,18 @@ fn main() {
     let zip_path = [&target_dir, &zip_filename].iter().collect::<PathBuf>();
 
     if !zip_path.exists() {
-        println!("start download to : {:?}", zip_path);
-        download_file(&url, zip_path.clone()).unwrap();
+        if url.starts_with("http://") || url.starts_with("https://") {
+            println!("start download to : {:?}", zip_path);
+            download_file(&url, zip_path.clone()).unwrap();
+        } else if url.starts_with("file://") {
+            let file_path = url.replace("file://", "");
```
nightkr, line 40: Arguably, file URLs should still be parsed as URLs. If it looks like a URL I'd usually assume urlencoding and such to apply as usual (on the other hand, building the path becomes easier when you just have to prepend file://... sigh).

utoipa-swagger-ui/build.rs
oscar6echo: In principle yes, but I prefer having minimal changes between the 2 cases. Here once the download/copy operation is performed, the next steps are exactly the same: Everthing happens in `target_dir`.

utoipa-swagger-ui/build.rs
oscar6echo: Except for space in a path, any other case where prepending `file://` is different from parsing as url ?

nightkr: Lots of special characters end up percentage-encoded. The most obvious case would be that `%` is used as the escape character, so it itself needs to be escaped as `%25`.

utoipa-swagger-ui/build.rs
nightkr: Fwiw, this also means that if the zip file changes then that won't take effect before you `cargo clean` (or remove the cached file manually).

juhaku: Hm. I would then guess it would be useful to make the file type of `SWAGGER_UI_DOWNLOAD_URL` to just rely on external Swagger UI. Then it is not tied to the _cargo lifecycle_ and can be externally updated.

oscar6echo: Note that if the file name changes (via env var update), then no need to `cargo clean`. And these `.zip` files are meant to be immutable, straight from official source.

utoipa-swagger-ui/build.rs
juhaku: Good point, though, not sure whether this actually comes to a problem, but if it does, I guess this could then be addressed. While I did not find a way to simply url encode string for the purpose with a quick look I assume as the `ZipArchive` works with rust `File` which itself uses `Path` to describe the location of a file system agnostic manner it should just simply fail if for one reason or another the rust cannot interpret the input as a valid `OsStr` that points to a real location.

That said e.g. Unix BASH does require strings to be double quoted to prevent the word globbing thus the variable could just be defined as `SWAGGER_UI_DOWNLOAD_URL="/path/to/my spaced swagger ui.zip"` if I am not mistaken.

utoipa-swagger-ui/build.rs
juhaku: Oh, yeah indeed because of this: `println!("cargo:rerun-if-changed={:?}", zip_path.clone());`

utoipa-swagger-ui/build.rs
juhaku: True.. Not a bad idea, I guess the `Url` is from the reqwest crate?

utoipa-swagger-ui/build.rs
oscar6echo: and just `if !zip_path.exists() {...}` tests if download/copy is necessary.

utoipa-swagger-ui/build.rs
oscar6echo: I was thinking of crate [url](https://crates.io/crates/url) and method [to_file_path](https://docs.rs/url/2.2.2/url/struct.Url.html#method.to_file_path).
What u think ?

juhaku: Seems okay, I checked the amount of dependencies and it only has few of them so think we can have this there as well. I believe it does not increase the compile time enormously to out weight the benefit of "validating" URL.

nightkr: I need to double check, but I think reqwest::Url is already just a reexport of url::Url.

utoipa-swagger-ui/build.rs
nightkr: The `rerun-if-changed` detects if the _cached zip_ changes, not the source zip fwiw.

utoipa-swagger-ui/build.rs
oscar6echo: Indeed `reqwest` can be used.  
So I committed to my fork: 
```rs
let file_path = Url::parse(&url).unwrap().to_file_path().unwrap();
```
Hopefully that should solve this edge case.

utoipa-swagger-ui/build.rs
oscar6echo: So to summarize:
+ one can change a `swagger-xxx.zip` file via env var `SWAGGER_UI_DOWNLOAD_URL` with `file://` 
+ these file, typically from [official releases](https://github.com/swagger-api/swagger-ui/tags), are uniquely identified by their name.
+ if for some reason somebody creates their own, they just have to change filename, change `SWAGGER_UI_DOWNLOAD_URL` correspondingly and `cargo build` - if they use a name the same name it will not be downloaded/copied again.
+ to modifiy the unzipped contents one can use env var `SWAGGER_UI_OVERWRITE_FOLDER` for selective overwrites - this will be applied in all cases.

This a fine grained config allows most if not all custom needs, I think.

utoipa-swagger-ui/build.rs
oscar6echo: Since the current PR is merged already, I created a new one with just this improvement: https://github.com/juhaku/utoipa/pull/925

juhaku: Merged that one as well

PR #926, opened 2024-05-15
fix syntax: change pub(super) enum to pub enum

cuiihaoo:
When I was using utoipa, I found a syntax error that prevented me from completing the cargo build. pub(super) enum str will cause the function parse_next_literal_str_or_include_str error on L.2913 in this file. As shown in the figure below:
image

Therefore, the deletion (super) is consistent with the `pub enum Value`  on L.2856 in this file.
image

### DIFF
--- a/utoipa-gen/src/lib.rs
+++ b/utoipa-gen/src/lib.rs
@@ -2984,7 +2984,7 @@ mod parse_utils {
 
     #[derive(Clone)]
     #[cfg_attr(feature = "debug", derive(Debug))]
-    pub(super) enum Str {
+    pub enum Str {
         String(String),
         IncludeStr(TokenStream),
     }


juhaku:
Thanks for fixing this, That is weird, I am not able to reproduce this issue, but better to change the syntax here. What Rust version you are using?

cuiihaoo:
> Thanks for fixing this, That is weird, I am not able to reproduce this issue, but better to change the syntax here. What Rust version you are using?

cargo 1.73.0 (9c4383fb5 2023-08-26)
rustc 1.73.0 (cc66ad468 2023-10-03)

juhaku:
Right, seems like an issue with the Rust level. The minimum will change in master to 1.75 from now on.

PR #930, opened 2024-05-17
Add nest `OpenApi` support

### LINKED ISSUE #445: Group and nest related paths/components
It seems to me that all the routes and components need to be defined in one spot when using `#[derive(OpenApi)]`. For a larger project, this can easily balloon in size. It would be nice if we could group related paths and components into smaller API definitions and include them all at once in a nested route.

```rs
mod foo {
    #[derive(SubApi)]
    #[subapi(
        paths(get_foo, set_foo, list_foos),
        components(schemas(Foo)),
    )]
    struct FooModule;

    // Routes and struct definitions...
    #[path(get, path = "/{id}", /* ... */)]
    async fn get_foo(Path(id): Path<u32>) { todo!() }
}

mod bar {
    #[derive(SubApi)]
    #[subapi(
        paths(get_bar, set_bar, list_bars),
        components(schemas(Bar)),
    )]
    struct BarModule;

    // Routes and struct definitions...
}

#[derive(OpenApi)]
#[openapi(
    nest(
        ("/foo", FooModule),
        ("/bar", BarModule),
    ),
)]
struct ApiDoc;
```

### LINKED ISSUE #872: Could utoipa support nesting of ApiDoc definitions?
axum support nesting for routers - https://docs.rs/axum/latest/axum/struct.Router.html#method.nest and it helps to structure large APIs a bit. But if we use utopia for the same API one needs to pass every path and nested type into #[derive(OpenApi)] in `#[openapi]`. Or is it intentional that API doc is describe as a whole?

juhaku:
This PR implements API nesting support to enable modular OpenAPI specification declaration. This functionality is implemented at OpenApi type itself as well as at OpenApi derive trait. Prior to this PR there was no simple way to modularize the OpenApi definition and one way to do this was to use `context_path` attribute on `#[utoipa::path]` macro but this functionality undoes the need for `context_path` in most cases.

Add `nest(...)` method to `OpenApi` to allow nesting of other `OpenApi` instances to the current `OpenApi` instance.
```rust
  let api = OpenApiBuider::new().build();
  let nested = api.nest("/nest/path", OpenApiBuilder::new().build());
```

Add `nest` attribute to `OpenApi` derive macro.
```rust
 #[derive(OpenApi)]
 #[openapi(
   paths(...),
   nest(
     ("path/to/nest", NestableApi)
   )
 )]
 struct RootApi;
```

Resolves #445 Resolves #872

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,6 @@
+[workspace.package]
+rust-version = "1.75"
+
 [workspace]
 resolver = "2"
 members = [
--- a/utoipa-gen/Cargo.toml
+++ b/utoipa-gen/Cargo.toml
@@ -8,6 +8,7 @@ readme = "README.md"
 keywords = ["openapi", "codegen", "proc-macro", "documentation", "compile-time"]
 repository = "https://github.com/juhaku/utoipa"
 authors = ["Juha Kukkonen <juha7kukkonen@gmail.com>"]
+rust-version.workspace = true
 
 [lib]
 proc-macro = true
--- a/utoipa-gen/src/component/serde.rs
+++ b/utoipa-gen/src/component/serde.rs
@@ -87,10 +87,11 @@ impl SerdeValue {
 
 /// The [Serde Enum representation](https://serde.rs/enum-representations.html) being used
 /// The default case (when no serde attributes are present) is `ExternallyTagged`.
-#[derive(Clone)]
+#[derive(Clone, Default)]
 #[cfg_attr(feature = "debug", derive(Debug))]
 #[cfg_attr(test, derive(PartialEq, Eq))]
 pub enum SerdeEnumRepr {
+    #[default]
     ExternallyTagged,
     InternallyTagged {
         tag: String,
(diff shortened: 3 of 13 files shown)


InAnYan:
@juhaku Hi! Is it available on cargo? Because compiler says `nest` is not expected field...

InAnYan:
UPDATE: yes, the feature is not released yet.

But I think I left my comment, in case anyone has this problem

wrongbyte:
@juhaku is there a predicted date when this function will be available? I just got here because in the [`todo-axum` example](https://github.com/juhaku/utoipa/blob/master/examples/todo-axum/src/main.rs), `nest` is used, but there's no disclaimer saying that this is not officially released.

juhaku:
The `master` branch is the working progress version. If you want to look for examples in certain version you should select a tag of your choice and then view the examples. 

> is there a predicted date when this function will be available?

There is no set date for it to come out officially, but should be soon enough nonetheless. Most of the functionality should be there while some work still remains.

Issue #934, opened 2024-05-21
Build script should respect the Cargo `http.cainfo`


ranger-ross:
When fetching Swagger UI in the [`build.rs`](https://github.com/juhaku/utoipa/blob/master/utoipa-swagger-ui/build.rs#L173) I think we should configure the reqwest client to respect the Cargo [`http.cainfo`](https://doc.rust-lang.org/cargo/reference/config.html#httpcainfo) setting.
This would be useful as sometimes various security software deployed by companies will use their own CA.

I think this should be a fairly small change and I would be happy to open a PR if you are up for it.

juhaku:
Yeah, this is probably a good idea. Though should be optional if possible.

PR #935, opened 2024-05-21
Use `CARGO_HTTP_CAINFO` CA file in build script if present

ranger-ross:
This PR adds support for using the [`CARGO_HTTP_CAINFO`](https://doc.rust-lang.org/cargo/reference/config.html#httpcainfo) cargo option if its present. 
See https://github.com/juhaku/utoipa/issues/934 for more details.

`CARGO_HTTP_CAINFO` is optional and the behavior will not change if it is not present. 
If it is present and we fail to read/parse the file, we just skip adding it and attempt to download the file without setting a custom CA.

### DIFF
--- a/utoipa-swagger-ui/build.rs
+++ b/utoipa-swagger-ui/build.rs
@@ -3,7 +3,7 @@ use std::{
     env,
     error::Error,
     fs::{self, File},
-    io,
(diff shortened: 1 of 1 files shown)


juhaku (approved):
Great, thanks for the PR

PR #936, opened 2024-05-21
Enhance utoipa-swagger-ui build

juhaku:
Use `curl` to download Swagger UI by default and allow using `reqwest` optionally with `reqwest` feature flag. Reqwest will be automatically enabled on Windows targets.

Change the re-build parameters as follows.
* For file protocol the re-build should happen when the upstream file changes
* For http protocol the re-build should happen when the environment variable holding the downloadd URL will change.

### DIFF
--- a/utoipa-swagger-ui/Cargo.toml
+++ b/utoipa-swagger-ui/Cargo.toml
@@ -12,15 +12,20 @@ authors = ["Juha Kukkonen <juha7kukkonen@gmail.com>"]
 rust-version.workspace = true
 
 [features]
+default = ["url"]
 debug = []
 debug-embed = ["rust-embed/debug-embed"]
+reqwest = ["dep:reqwest"]
+url = ["dep:url"]
 
 [dependencies]
 rust-embed = { version = "8" }
 mime_guess = { version = "2.0" }
 actix-web = { version = "4", optional = true, default-features = false }
 rocket = { version = "0.5", features = ["json"], optional = true }
-axum = { version = "0.7", default-features = false, features = ["json"], optional = true }
+axum = { version = "0.7", default-features = false, features = [
+    "json",
+], optional = true }
 utoipa = { version = "4", path = "../utoipa" }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0" }
@@ -35,4 +40,18 @@ rustdoc-args = ["--cfg", "doc_cfg"]
 [build-dependencies]
 zip = { version = "1", default-features = false, features = ["deflate"] }
 regex = "1.7"
-reqwest = { version = "0.12", features = ["blocking", "rustls-tls"], default-features = false }
+
+# enabled optionally to allow rust only build with expense of bigger dependency tree and platform
+# independant build. By default `curl` system package is tried for downloading the Swagger UI.
+reqwest = { version = "0.12", features = [
+    "blocking",
+    "rustls-tls",
+], default-features = false, optional = true }
+url = { version = "2.5", optional = true }
+
+# Windows is forced to use reqwest to download the Swagger UI.
+[target.'cfg(windows)'.build-dependencies]
+reqwest = { version = "0.12", features = [
+    "blocking",
+    "rustls-tls",
+], default-features = false }
--- a/utoipa-swagger-ui/build.rs
+++ b/utoipa-swagger-ui/build.rs
@@ -1,10 +1,10 @@
-use reqwest::Url;
 use std::{
     env,
     error::Error,
     fs::{self, File},
-    io::{self, Read},
-    path::PathBuf,
+    io,
+    path::{Path, PathBuf},
+    process::Command,
 };
 
 use regex::Regex;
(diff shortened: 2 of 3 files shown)


juhaku:
Follow up for #923 

@nightkr @oscar6echo 
I am making further enhancements for the build.

oscar6echo:
Thx for refining it further.

utoipa-swagger-ui/Cargo.toml
```diff
@@ -12,15 +12,20 @@ authors = ["Juha Kukkonen <juha7kukkonen@gmail.com>"]
 rust-version.workspace = true
 
 [features]
+default = ["url"]
 debug = []
 debug-embed = ["rust-embed/debug-embed"]
+reqwest = ["dep:reqwest"]
+url = ["dep:url"]
```
nightkr, line 19: I don't see anything feature-flagged on url, so it looks like it's just always required?

utoipa-swagger-ui/build.rs
```diff
@@ -22,34 +22,40 @@ use zip::{result::ZipError, ZipArchive};
...
-        }
+    if url.starts_with("file:") {
+        let mut file_path = url::Url::parse(&url).unwrap().to_file_path().unwrap();
+        file_path = fs::canonicalize(file_path).expect("swagger ui download path should exists");
+
+        // with file protocol utoipa swagger ui should compile when file changes
+        println!("cargo:rerun-if-changed={:?}", file_path);
+
+        println!("start copy to : {:?}", zip_path);
+        fs::copy(file_path, zip_path.clone()).unwrap();
```
nightkr, line 47: At this point this _really_ just seems equivalent to `zip_path = file_path;`.

utoipa-swagger-ui/build.rs
```diff
@@ -22,34 +22,40 @@ use zip::{result::ZipError, ZipArchive};
...
+const SWAGGER_UI_OVERWRITE_FOLDER: &str = "SWAGGER_UI_OVERWRITE_FOLDER";
+
 fn main() {
     let target_dir = env::var("OUT_DIR").unwrap();
-    println!("OUT_DIR: {}", target_dir);
+    println!("OUT_DIR: {target_dir}");
 
     let url =
-        env::var("SWAGGER_UI_DOWNLOAD_URL").unwrap_or(SWAGGER_UI_DOWNLOAD_URL_DEFAULT.to_string());
+        env::var(SWAGGER_UI_DOWNLOAD_URL).unwrap_or(SWAGGER_UI_DOWNLOAD_URL_DEFAULT.to_string());
```
nightkr, line 33: `url` is parsed into `url::Url` on every branch taken (explicitly for file://, explicitly for curl, and implicitly via `IntoUrl` for reqwest).

Doing it up here instead would also give access to `Url::schema` instead of approximating it with `str::starts_with`.

utoipa-swagger-ui/build.rs
```diff
@@ -22,34 +22,40 @@ use zip::{result::ZipError, ZipArchive};
...
 
-    println!("SWAGGER_UI_DOWNLOAD_URL: {}", url);
+    println!("{SWAGGER_UI_DOWNLOAD_URL}: {url}");
     let zip_filename = url.split('/').last().unwrap().to_string();
     let zip_path = [&target_dir, &zip_filename].iter().collect::<PathBuf>();
 
-    if !zip_path.exists() {
```
nightkr, line 36: I assume the cache is still desired for HTTP(S) downloads.

utoipa-swagger-ui/build.rs
oscar6echo: Indeed.

utoipa-swagger-ui/Cargo.toml
oscar6echo: I think there are some places like [here](https://github.com/juhaku/utoipa/blob/403d716513612cb52bef0525640b9a96cd1bb18f/utoipa-swagger-ui/build.rs#L155).  

But in principle at least, url is necessary to parse `SWAGGER_UI_DOWNLOAD_URL` in all cases except in the case of vendored swagger_ui.zip so maybe the vendored flag is enough ?

utoipa-swagger-ui/build.rs
oscar6echo: Yes: I understand that this is covered by `println!("cargo:rerun-if-env-changed={SWAGGER_UI_DOWNLOAD_URL}");
` [here](https://github.com/juhaku/utoipa/blob/403d716513612cb52bef0525640b9a96cd1bb18f/utoipa-swagger-ui/build.rs#L179), right ?

nightkr: Sort of, `rerun-if-env-changed` tells cargo about _more_ reasons to rebuild, while the old cache was trying to avoid redownloading when it _was_ rebuilt. (On the other hand, I'm not sure how useful that was anyway given that I'm not sure how Cargo decides on when to recreate out folders.)

nightkr: To be clear: I'm fine with the changed behaviour, just wanted to double check that it was intended.

juhaku: Yeah, I removed it just because it appears to be not constant on creating the folders thus even if there was the cache previously cargo might still create new set of folders when building thus invalidating the cache in a mean time as well. Perhaps just better of to let it download it when ever cargo should rebuild it. After all that is not going to be build unless one runs `cargo clean`.

utoipa-swagger-ui/build.rs
juhaku: True that. Maybe something that can be considered in future if it starts to bother someone :laughing: Not going to edit it for now though

utoipa-swagger-ui/Cargo.toml
juhaku: True, would be enough at this point, and would work without explicit `url` feature since the optional dependencies becomes implicit feature flags anyways. But it does not matter.

utoipa-swagger-ui/build.rs
juhaku: Yeah, while this is little bit older commit but in practice yes, the copy seems to be superfluous since it is anyway extracted to the `OUTPUT` dir. But perhaps we can live with that :laughing:

PR #951, opened 2024-05-26
utoipa-gen next version

ifree92:


juhaku:
This would actually need a new release but there is more work to be done before next release will be out.

ifree92:
Oh got it.
Should I then just close my PR for now? Or is there any way to add a patch version for now?
Thank you

juhaku:
> Oh got it. Should I then just close my PR for now?

Yes, there is no reason to merge this PR. Also the next version to be release is 5.0.0 so all the crates need to be released.

>  Or is there any way to add a patch version for now? Thank you

At the moment only way is to use the git reference in in the dependencies block of `Cargo.toml` file. Pointing it to the master. or perhaps the latest commit of the master to give some sort of stability so you would not get always the next changes from the master when merged. More from specifying git dependencies https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#choice-of-commit.

Perhaps I will release a alpha version of 5 in couple of days but I am not sure do I have enough time for it. That would be unstable version.

Issue #727, opened 2023-08-09
`#[schema(inline)]` for enums like for structs


smndtrl:
After digging through `utoipa-gen` tests, I was wondering, if there's a possibility for using sth like `#[schema(inline)]` for `enum`s so that definitions that were otherwise end up being referenced are inlined.

```rust
enum Auth {
    #[schema(inline)] 
    BasicAuth(BasicAuth),
    ClientCertAuth { certificate: Vec<u8> },
}
```

would end up being a schema without any `$ref`. Maybe it would make sense to move the `#[schema(inline)]` to the actual type like this `BasicAuth(#[schema(inline)] BasicAuth)`

We have another system that doesn't handle `$ref` too well. Having the `#[schema(inline)]` for structs, it would be nice for enums as well.

juhaku:
Pretty good idea. Surely this is something that could be explored in future and if possible added to the library as well. Also PRs are welcome if there is enough interest on implementing this feature.

JMLX42:
@juhaku if that's still on the table I'll try to do it.

JMLX42:
@juhaku @smndtrl PR: #963

juhaku:
@JMLX42 great, all the help is welcome :slightly_smiling_face:

juhaku:
Done in the PR @JMLX42 pointed above

PR #963, opened 2024-06-20
Add support for inlined enum variants

JMLX42:
Implementation for #727

The following is now supported:

```rust
    #[derive(ToSchema)]
    enum Number {
        One,
        Two,
        Three,
        Four,
        Five,
        Six,
        Seven,
        Height,
        Nine,
    }

    #[derive(ToSchema)]
    enum Color {
        Spade,
        Heart,
        Club,
        Diamond,
    }

    #[derive(ToSchema)]
    enum Card {
        Number(#[schema(inline)] Number),
        Color(#[schema(inline)] Color),
    }
```

and will produce the following schema:

```json
{
            "oneOf": [
                {
                    "properties": {
                        "Number": {
                            "enum": [
                                "One",
                                "Two",
                                "Three",
                                "Four",
                                "Five",
                                "Six",
                                "Seven",
                                "Height",
                                "Nine",
                            ],
                            "type": "string",
                        },
                    },
                    "required": [
                        "Number",
                    ],
                    "type": "object",
                },
                {
                    "properties": {
                        "Color": {
                            "enum": [
                                "Spade",
                                "Heart",
                                "Club",
                                "Diamond",
                            ],
                            "type": "string",
                        },
                    },
                    "required": [
                        "Color",
                    ],
                    "type": "object",
                },
            ],
        }
```

### DIFF
--- a/utoipa-gen/src/component/schema.rs
+++ b/utoipa-gen/src/component/schema.rs
@@ -619,6 +619,14 @@ impl ToTokensDiagnostics for UnnamedStructSchema<'_> {
 
             if fields_len == 1 {
                 if let Some(ref mut features) = unnamed_struct_features {
+                    let inline =
+                        features::parse_schema_features_with(&first_field.attrs, |input| {
+                            Ok(parse_features!(input as super::features::Inline))
+                        })?
+                        .unwrap_or_default();
+
+                    features.extend(inline);
+
                     if pop_feature!(features => Feature::Default(crate::features::Default(None)))
                         .is_some()
                     {
--- a/utoipa-gen/src/lib.rs
+++ b/utoipa-gen/src/lib.rs
@@ -133,6 +133,30 @@ use self::{
 /// _`rename`_ attribute. It behaves similarly to serde's _`rename`_ attribute. If both _serde_
 /// _`rename`_ and _schema_ _`rename`_ are defined __serde__ will take precedence.
 ///
+/// ## Enum Unnamed Variant Field Configuration Options
+///
+/// * `inline` If the type of this field implements [`ToSchema`][to_schema], then the schema definition
+///   will be inlined. **warning:** Don't use this for recursive data types!
+///
+///   _**Inline unnamed field variant schemas.**_
+///   ```rust
+///   # use utoipa::ToSchema;
+///   # #[derive(ToSchema)]
+///   # enum Number {
+///   #     One,
+///   # }
+///   #
+///   # #[derive(ToSchema)]
+///   # enum Color {
+///   #     Spade,
+///   # }
+///    #[derive(ToSchema)]
+///    enum Card {
+///        Number(#[schema(inline)] Number),
+///        Color(#[schema(inline)] Color),
+///    }
+///   ```
+///
 /// # Unnamed Field Struct Optional Configuration Options for `#[schema(...)]`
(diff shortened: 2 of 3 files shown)


utoipa-gen/src/component/schema.rs
```diff
@@ -619,6 +619,14 @@ impl ToTokensDiagnostics for UnnamedStructSchema<'_> {
 
             if fields_len == 1 {
                 if let Some(ref mut features) = unnamed_struct_features {
+                    let inline =
+                        features::parse_schema_features_with(&first_field.attrs, |input| {
```
JMLX42, line 623: @juhaku that looks kinda naive. Isn't there a way to properly specify which features are supported by this specific case?

JMLX42: What I mean is I'd rather have something like this: https://github.com/juhaku/utoipa/blob/d020f92a0446adc7e99999aa9e1be63e95728fbb/utoipa-gen/src/component/schema/features.rs#L150

But those features are for the variant themselves, so `Inline` would be:

```rust
    #[derive(ToSchema)]
    enum Card {
        #[schema(inline)]
        Number(Number),
        #[schema(inline)]
        Color(Color),
    }
```

Which IMO does not fit with the rest of the API and is counter-intuitive.

But I haven't found the features struct for the unnamed field.

juhaku: > @juhaku that looks kinda naive. Isn't there a way to properly specify which features are supported by this specific case?

No features have been originally defined for unnamed field enum variants itself thus there is no such struct. Though if there is just a single feature to be parsed the `parse_with` is just fine but if there would be multiple features or the set of features is known to be needed in multiple locations then a feature struct is preferred. 

> Which IMO does not fit with the rest of the API and is counter-intuitive.


Yeah, I agree the API would be counter intuitive and the features defined directly for the unnamed enum variant field is cleaner approach. 

> But I haven't found the features struct for the unnamed field.

Such struct does not exists by design but could be added if needed.

juhaku (changes requested):
This is good for me :tada: 

Though docs should be updated with the added feature at `ToSchema` macro by changing the doc comment here https://github.com/juhaku/utoipa/blob/d020f92a0446adc7e99999aa9e1be63e95728fbb/utoipa-gen/src/lib.rs#L58.

JMLX42:
> This is good for me 🎉
> 
> Though docs should be updated with the added feature at `ToSchema` macro by changing the doc comment here
> 
> https://github.com/juhaku/utoipa/blob/d020f92a0446adc7e99999aa9e1be63e95728fbb/utoipa-gen/src/lib.rs#L58
> .

@juhaku I guess [this section](https://github.com/juhaku/utoipa/blob/d020f92a0446adc7e99999aa9e1be63e95728fbb/utoipa-gen/src/lib.rs#L136) would be the right place to document the feature?

juhaku:
> @juhaku I guess [this section](https://github.com/juhaku/utoipa/blob/d020f92a0446adc7e99999aa9e1be63e95728fbb/utoipa-gen/src/lib.rs#L136) would be the right place to document the feature?

Yes that chapter would be the right place for the doc. It should describe that it is used directly for the variant field and not for the variant itself. And perhaps an example of this would be good to add as well under the examples section a little bit further down.

PR #982, opened 2024-07-29
fix(utoipa-gen): remove unnecessary allocation with to_string in expanded code

CBenoit:
~~Taking something that is a string and converting it to a string using `to_string()` kind of misses the point of why we are doing the conversion in the first place, and also fails to document this to the reader. `to_owned()` fully captures the reason that a conversion is required at a particular spot: borrowed (`&str`) to owned (`String`).~~

~~Some people enable this clippy lint in order to enforce this style:~~

~~- https://rust-lang.github.io/rust-clippy/master/index.html#/str_to_string~~

~~Code generated by utoipa macro is triggering this lint. To avoid that, this patch is replacing the `to_string()` usages in generated code by `to_owned()`.~~

~~This patch is not changing any other `to_string()` because the utoipa project itself does not need to adhere to this style. The goal is just to reduce friction for downstream consumers.~~

Here is how it shows up without this fix:
```
warning: `to_string()` called on a `&str`
   --> devolutions-gateway/src/api/webapp.rs:271:1
    |
271 | / #[utoipa::path(
272 | |     post,
273 | |     operation_id = "SignSessionToken",
274 | |     tag = "WebApp",
...   |
286 | |     ),
287 | | )]
    | |__^
    |
    = help: consider using `.to_owned()`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#str_to_string
    = note: this warning originates in the attribute macro `utoipa::path` (in Nightly builds, run with -Z macro-backtrace for more info)
```

~~There is no lint for the opposite ("use `to_string()` instead of `to_owned()`), so it seems safe to apply this fix.~~

### DIFF
--- a/utoipa-gen/src/path.rs
+++ b/utoipa-gen/src/path.rs
@@ -394,7 +394,7 @@ impl<'p> ToTokensDiagnostics for Path<'p> {
             })
             .unwrap_or_else(|| {
                 quote! {
-                    #path.to_owned().replace('"', "")
+                    #path.replace('"', "")
                 }
             });
 
--- a/utoipa-gen/src/path.rs
+++ b/utoipa-gen/src/path.rs
@@ -386,8 +386,8 @@ impl<'p> ToTokensDiagnostics for Path<'p> {
                 let context_path = context_path.to_token_stream();
                 let context_path_tokens = quote! {
                     format!("{}{}",
-                        #context_path.to_owned().replace('"', ""),
-                        #path.to_owned().replace('"', "")
+                        #context_path.replace('"', ""),
+                        #path.replace('"', "")
                     )
                 };
                 context_path_tokens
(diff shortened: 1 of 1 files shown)


juhaku:
Might be just useful to try not to actually use the `.to_string()` or `to_owned()` call in the first place. Mist likely it is relic from older times. But anyhow if it is necessary to have either one then I guess this change for the to_owned() is alright.

CBenoit:
To be fair, I don't think either one are really relic of the past.

ToString is a generic way of converting a value into a string representation (typically derived from Display), and ToOwned a generic way of obtaining an owned version of the value. to_string used to be less performant because it was going through the formatting machinery, but this was fixed in 2017 or so using specialization, and the behavior is now effectively identical to_owned, so as of today it's mostly a matter of taste. I think that ToOwned is the semantic we are after when converting a &str into String, but there is no reason for both &str and String to not implement ToString and it's notably useful in generic code, that's probably why it is the way it is.

That being said, this could be replaced by String::from  (infaillible type conversion), would you prefer that? I think it's equally good, both semantically and, most importantly here, in the fact that there is no warning against that.

CBenoit:
Sounds right! I removed the superfluous allocation.

juhaku (approved):
Great :100:

PR #983, opened 2024-07-31
Add extensions support for schemas

heat1q:
When deserializing existing OpenAPI files with `utoipa` I noticed that extensions on schemas are not included. Though as noted in the [spec](https://spec.openapis.org/oas/latest.html#schema-object) schema objects may contain extensions. I saw that in #763 extensions support was added but not for schemas. Was this intentional? 

This PR also adds support for arbitrary extensions on schema objects.

### DIFF
--- a/utoipa/src/openapi/schema.rs
+++ b/utoipa/src/openapi/schema.rs
@@ -2,7 +2,7 @@
 //! used to define field properties, enum values, array or object types.
 //!
 //! [schema]: https://spec.openapis.org/oas/latest.html#schema-object
-use std::collections::BTreeMap;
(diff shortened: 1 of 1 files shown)


juhaku:
> When deserializing existing OpenAPI files with utoipa I noticed that extensions on schemas are not included. Though as noted in the [spec](https://spec.openapis.org/oas/latest.html#schema-object) schema objects may contain extensions. I saw that in https://github.com/juhaku/utoipa/pull/763 extensions support was added but not for schemas. Was this intentional?

I don't remember anymore whether it was intentional but thanks for making a PR for adding it to the `Schema` type as well.

juhaku (approved):

PR #987, opened 2024-08-09
Fix openapi serialized twice when Rapidoc is served with Rocket

Rauks:
Hello,

When Rocket serves a Rapidoc built using `with_openapi`, the served `openapi.json` is the result of an `ApiDoc` serialized twice.

```rust
rocket::build()
    .mount(
        "/",
        RapiDoc::with_openapi("/api-docs/openapi.json", ApiDoc::openapi()).path("/rapidoc"),
    )
```

As a result, Rapidoc fails to load the spec:

image

Inspecting the request response for `/api-docs/openapi.json` shows a "string of a json":

image

This PR is a fix proposal that resolves the double serialization.
After the fix, the spec now loads correctly and the `/api-docs/openapi.json` request response inspection shows "just a json":

image

image

### DIFF
--- a/utoipa-rapidoc/src/lib.rs
+++ b/utoipa-rapidoc/src/lib.rs
@@ -361,7 +361,7 @@ mod rocket {
                 routes.push(Route::new(
                     Method::Get,
                     value.spec_url.as_ref(),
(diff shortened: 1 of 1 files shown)


juhaku (approved):
Thanks for the catch :+1:

PR #1003, opened 2024-08-25
Add axum bindings for utoipa

### LINKED ISSUE #991: Axum bindings for utoipa - better path registration for axum
## Description

Currently with axum one needs to do duplicate work when adding utoipa `#[utoipa::path]` macros to the `OpenApi` and then registering those handlers for `axum::Router`.

```rust
#[derive(OpenApi)]
#[openapi(
    nest(
        (path = "/api/v1/todos", api = todo::TodoApi)
    ),
    tags(
        (name = "todo", description = "Todo items management API")
    )
)]
struct ApiDoc;

[derive(OpenApi)]
#[openapi(
    paths(list_todos, search_todos, create_todo, mark_done, delete_todo,),
    components(schemas(Todo, TodoError))
)]
pub(super) struct TodoApi;

let app = Router::new()
    .nest("/api/v1/todos", todo::router());
mod todo {
    pub(super) fn router() -> Router {
        let store = Arc::new(Store::default());
        Router::new()
            .route("/", routing::get(list_todos).post(create_todo))
            .route("/search", routing::get(search_todos))
            .route("/:id", routing::put(mark_done).delete(delete_todo))
            .with_state(store)
    }
}
```

## Proposed Solution

Solution currently under experimenting makes huge improvement to the current one by allowing users directly use custom `OpenApiRouter` that handles `#[utoipa::path]` attribute macro path registration behind the scenes and registers the route handlers on the single go. ~This completely makes OpenApi derive macro redundant but can be used if needed together with the router. The benefit is that there will be no more duplicate work and handlers can be registered once.~ This completely makes `OpenApi` derive macro redundant in terms of path registration but schemas still need to be registered the old way or directly to the `OpenApi` type. 
```rust
let user_router: OpenApiRouter = OpenApiRouter::new()
    .routes(utoipa_axum::get(search_user))
    .routes(utoipa_axum::get(get_user).post_path(post_user).delete_path(delete_user));

let customer_router: OpenApiRouter = OpenApiRouter::new()
    .routes(
        utoipa_axum::get(get_customer)
            .post_path(post_customer)
            .delete_path(delete_customer),
    )
    .routes(utoipa_axum::get(search_customer));

let mut router = OpenApiRouter::new()
    .nest("/api/user", user_router)
    .nest("/api/customer", customer_router)
    .route("/", get(root));
let api = router.to_openapi();
let router: axum::Router = router.into();
```

Relates #624 Relates #662 Relates #537 Relates #394

juhaku:
This PR adds a new crate `utoipa-axum` which provides bindings between
`axum` and `utoipa`. It aims to blend as much as possible to the
existing philosophy of axum way of registering handlers.

This commit introduces new `OpenApiRouter` what wraps `OpenApi` and axum
`Router` which provides passthrough implementation for most of the axum
Router methods and collects and combines the `OpenApi` from registered
routes. Routes registred only via `routes!()` macro will get added to
the `OpenApi`.

Also this commit introduces `routes!()` macro which collects axum
handlers annotated with `#[utoipa::path()]` attribute macro to single
paths intance which is then provided to the `OpenApiRouter`.

Fixes #991


juhaku:
Done here #1004 with better branch name

PR #1004, opened 2024-08-26
Add utoipa axum bindings

### LINKED ISSUE #394: Use API spec for actual route definition
### Discussed in https://github.com/juhaku/utoipa/discussions/355

<div type='discussions-op-text'>

<sup>Originally posted by **TitanNano** November 19, 2022</sup>
We started to build our public API with `axum` and `utoipa` and are already facing the issue that we discover miss matches between the API spec and the actual implementation in axum.

Examples for this are, `route` path does not match `utoipa` path or method does not match `utoipa` path. For this reason, I started looking into what is being generated for the OpenAPI spec and how we could use that to avoid the drift between implementation and documentation. 
What I found is quite nice, we can access the entire APS spec and extract all paths and methods from it to build our router. I discovered only two issues that block us from doing this, though:

1. there is no way of getting from the API spec to the correct handler function for axum. There is the operation id which by default matches the handler function name and with some macro magic, I might be able to locate the correct function, but I do not want to expose such an implementation detail in our public API spec. We would be depending on the operation id always matching the function name. 
2. We are currently using route layer extensions to inject dependencies into our route handlers, and we will need some way of associating them with the path definition.</div>

### LINKED ISSUE #537: Derive URL path form handler to avoid declaring it twice (axum)
Hi!

I think it is quite error prone when the URL of an API endpoint has to be declared twice.

```rust
let app = Router::new()
    .route("/users", post(create_user))
```

```rust
/// Create new user
#[utoipa::path(
    post,
    path = "/users",
    request_body = CreateUser,
    responses(
        (status = 201, description = "Create user", body = [String])
    )
)]
pub async fn create_user(
    // this argument tells axum to parse the request body
    // as JSON into a `CreateUser` type
    Json(payload): Json<CreateUser>,
) -> (StatusCode, Json<User>) {
   // .. 
}
```

Could we add an option to make utoipa derive it from the API handler declaration? If that is possible I can also imagine that the Method (GET, POST) and even the output schema can be derived to avoid boilerplate. 

Thanks!

### LINKED ISSUE #991: Axum bindings for utoipa - better path registration for axum

juhaku:
This PR adds a new crate `utoipa-axum` which provides bindings between
`axum` and `utoipa`. It aims to blend as much as possible to the
existing philosophy of axum way of registering handlers.

This commit introduces new `OpenApiRouter` what wraps `OpenApi` and axum
`Router` which provides passthrough implementation for most of the axum
Router methods and collects and combines the `OpenApi` from registered
routes. Routes registred only via `routes!()` macro will get added to
the `OpenApi`.

Also this commit introduces `routes!()` macro which collects axum
handlers annotated with `#[utoipa::path()]` attribute macro to single
paths intance which is then provided to the `OpenApiRouter`.

Example of supported syntax.
```rust
let user_router: OpenApiRouter = OpenApiRouter::new()
    .routes(routes!(search_user))
    .routes(routes!(get_user, post_user, delete_user));
```

Fixes #991 Fixes #394 Fixes #537


Borderliner:
Can't wait for this feature! Please package and release it.

juhaku:
Will do that shortly, shouldn't be too long.

PR #1020, opened 2024-09-03
Wip make `as` attribute authoritative name

### LINKED ISSUE #993: `#[schema(as = ...)]` infects the code with `#[schema(value_type = ...)]` and `#[param(value_type = ...)]`
First of all, thank you for `utoipa`!

## Context

We use `utoipa` to generate the OpenAPI spec for a web service, which - among other things - proxies requests for a category of Providers of data.

For that reason, we have to expose third-party OpenAPI types in ours.

Since there are multiple Providers in the category and they all deal with the same domain, their API names overlap.

So `provider1::Thing` is not the same thing as `provider2::Thing`.

To discriminate them in the OpenAPI spec, we use paths like `provider1.Thing` and `provider2.Thing`.

`provider1/mod.rs`:

```rs
#[derive(ToSchema, ...)]
#[schema(as = provider1::Thing)]
pub struct Thing {
  // ...
}
```

and a similar thing in `provider2/mod.rs`.

## Problem

Unfortunately, if we try to use `Thing` from `provider1/mod.rs` in another type...

`provider1/mod.rs`:

```
#[derive(ToSchema, ... )]
pub struct Stuff {
  pub thing: Thing,
  // ...
}
```

**Expectation:** Everything works just fine.
**Reality:**: the type of field `thing` in the OpenAPI spec appears as `Thing` rather than `provider1.Thing` even though `provider1::Thing` implements `ToSchema` and has a `#[schema(as = ...)]` attribute, which can cause either collisions or unknown identifier errors.
**Workaround:** We have to write the code like this.

```
#[derive(ToSchema, ... )]
pub struct Stuff {
  #[schema(value_type = provider1::Thing)]
  pub thing: Thing,
  // ...
}
```

and for bigger structs, it nearly doubles the size of the struct, making the code less readable.

Not only that, in combination with `axum`, it forces me to define a custom `Path` type because if we use `Path` extractor from `axum`,

```
#[utoipa::path(
        ...
    )]
    pub async fn get_thing(
        State(ctx): State<Arc<Context>>,
        Path(path_params): Path<(Uuid, Uuid, ThingId)>,
    ) -> impl IntoResponse {
        // ...
    }
```

**Expectation:** Everything works fine.
**Reality**: The type `provider1::ThingId` appears in the OpenAPI spec as `ThingId` even though `ThingId` (trust me) implements `ToSchema` and has a `#[schema(as = ...)` attribute, which can cause either collisions or unknown identifier errors.
**Workaround:** We have to define a whole new type:

```
mod get_thing {
        use utoipa::IntoParams;
        use uuid::Uuid;

        use crate::provider1::ThingId;

        #[derive(IntoParams, ...)]
        pub struct PathParams {
            pub first_uuid: Uuid,
            pub second_uuid: Uuid,
            #[param(value_type = provider1::ThingId)]
            pub thing_id: ThingId,
        }
    }
```

so that we can use it

```
#[utoipa::path(
        ...
    )]
    pub async fn get_thing(
        State(ctx): State<Arc<Context>>,
        path_params: get_thing::PathParams,
    ) -> impl IntoResponse {
        let PathParams {
            first_uuid,
            second_uuid,
            thing_id,
        } = path_params;
        // ...
    }
```

and have the correct OpenAPI schema.

## Possible solutions

1. Just rely on the workarounds.

**Pros:** It's a no-op for the maintainers.
**Cons:** It induces pain for the `utoipa` users who have to use `#[schema(as = ...)]` types for struct fields.

2. Add an opt-in feature for the code that would infer the OpenAPI names of types from the Rust types.

**Pros:** It *considerably eases* the pain for the `utoipa` users who have to use `#[schema(as = ...)]` types for struct fields.
**Cons:** This adds a feature (=complexity). It also burdens the users who have to use `#[schema(as = ...)]` types for struct fields with the need to discover and then enable the feature.

3. Make a breaking change that would define the "smart" behavior as the default one.

**Pros:** It *completely eliminates* the pain for the `utoipa` users who have to use `#[schema(as = ...)]` types for struct fields.
**Cons:** It is a subtle breaking change *without a quick fix*.

**Note:** I don't see a world where someone would rely on the original behavior without understanding the risk. It's a weird behavior after all.

4. Make a breaking change that would define the "smart" behavior as the default one but add a feature that would "dumb down" the OpenAPI type name inference.

**Pros:** It *completely eliminates* the pain for the `utoipa` users who have to use `#[schema(as = ...)]` types for struct fields.
**Cons:** It still is a subtle breaking change but *with a quick fix*.

UPDATE:

There's another dimension: `utoipa` can temporarily start using both "smart" and "dumb" inference and in case of discrepancies emit warnings. This can set stage for a smooth breaking change. However, it can lead to a temporary feature creep, which will be resolved when the breaking changes are landed.

## Suggestion

I like the idea with running both types of inferences by default because it allows to relatively safely introduce a breaking change. However, opt-in and opt-out behaviors are complicated due to the additive nature of features.

juhaku:
This PR aims to change the logic how the prefix is defined for the request bodies and response bodies throughout the library. Prior to this the name is resolved from the value provided to the `request_body` or response `body` attribute. However this current logic becomes cumbersome when user needs to always provide the correct path to the attribute value that correspond the name in the OpenAPI specification.

This changes it so that the name itself will be resolved from the name of the schema itself thus eliminating the need for manual path prefixing. If schema itself is provided with `as` attribute then the name will be resolved from the same `as` attribute value.

Fixes #993

### DIFF
--- a/utoipa-gen/Cargo.toml
+++ b/utoipa-gen/Cargo.toml
@@ -23,7 +23,7 @@ ulid = { version = "1", optional = true, default-features = false }
 url = { version = "2", optional = true }
 
 [dev-dependencies]
-utoipa = { path = "../utoipa", features = ["debug", "uuid"], default-features = false }
+utoipa = { path = "../utoipa", features = ["debug", "uuid", "macros"], default-features = false }
 serde_json = "1"
 serde = "1"
 actix-web = { version = "4", features = ["macros"], default-features = false }
--- a/utoipa-gen/src/component.rs
+++ b/utoipa-gen/src/component.rs
@@ -2,12 +2,19 @@ use std::borrow::Cow;
 
 use proc_macro2::{Ident, Span, TokenStream};
 use quote::{quote, quote_spanned, ToTokens};
+use syn::punctuated::Punctuated;
 use syn::spanned::Spanned;
-use syn::{Attribute, GenericArgument, Path, PathArguments, PathSegment, Type, TypePath};
+use syn::{
+    AngleBracketedGenericArguments, Attribute, GenericArgument, GenericParam, Generics, Path,
+    PathArguments, PathSegment, Type, TypePath,
+};
 
 use crate::doc_comment::CommentAttributes;
 use crate::schema_type::{SchemaFormat, SchemaTypeInner};
-use crate::{as_tokens_or_diagnostics, AttributesExt, Diagnostics, OptionExt, ToTokensDiagnostics};
+use crate::{
+    as_tokens_or_diagnostics, AttributesExt, Diagnostics, GenericsExt, OptionExt,
+    ToTokensDiagnostics,
+};
 use crate::{schema_type::SchemaType, Deprecated};
 
 use self::features::attributes::{Description, Nullable};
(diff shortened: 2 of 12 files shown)


juhaku:
Blocked for new generics, Current aliases will break with this implementation.

juhaku:
Closed, this is implemented here #1034 along with new generics implementation

PR #1022, opened 2024-09-04
Use fs::read to overwrite swagger UI contents

joaquin041:
When using SWAGGER_UI_OVERWRITE_FOLDER, text files get properly overwritten, but other files like icons fail with:
`cannot read content from file: icon.png`

Using `fs::read` instead of `fs::read_file` solves the issue. It overwrites the files using bytes instead of strings.


juhaku (approved):
Neat, thanks for the patch :+1:

PR #1036, opened 2024-09-09
Add typos to CI

jayvdb:
Please merge https://github.com/juhaku/utoipa/pull/1038 first

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -91,6 +91,17 @@ jobs:
           if [[ ${{ steps.changes.outputs.changes }} == true ]]; then
             ./scripts/test.sh ${{ matrix.crate }}
           fi
+
+  check-typos:
+    name: typos
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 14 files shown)


.github/workflows/build.yaml
```diff
@@ -87,6 +91,17 @@ jobs:
           if [[ ${{ steps.changes.outputs.changes }} == true ]]; then
             ./scripts/test.sh ${{ matrix.crate }}
           fi
+
+  check-typos:
+    name: typos
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
juhaku, line 99: What does this do? Does it throw error if it finds typos?

jayvdb: yup , built with rust https://crates.io/crates/typos-cli

c.f. https://github.com/taiki-e/install-action

jayvdb: it is configurable if there are false positives

juhaku: Alright, I guess we can have it.

Issue #1037, opened 2024-09-09
Builds failing for individual packages utoipa-axum & utoipa-swagger-ui


jayvdb:
see https://github.com/juhaku/utoipa/actions/runs/10779351123/job/29892650558?pr=1036 and other jobs

The following error is repeated many times
```
if [[ true == true ]]; then
    cargo clippy --quiet --package utoipa-axum
  fi
  shell: /usr/bin/bash -e {0}
  env:
    CARGO_TERM_COLOR: always
error[E04[3](https://github.com/juhaku/utoipa/actions/runs/10779351123/job/29892651722?pr=1036#step:6:3)3]: failed to resolve: could not find `utoipa` in the crate root
   --> utoipa/src/openapi/schema.rs:908:[4](https://github.com/juhaku/utoipa/actions/runs/10779351123/job/29892651722?pr=1036#step:6:4)0
    |
908 |         pub multiple_of: Option<crate::utoipa::Number>,
    |                                        ^^^^^^ could not find `utoipa` in the crate root
    |
note: found an item that was configured out
   --> utoipa/src/lib.rs:46[5](https://github.com/juhaku/utoipa/actions/runs/10779351123/job/29892651722?pr=1036#step:6:5):5
    |
465 | mod utoipa {
    |     ^^^^^^
    = note: the item is gated behind the `macros` feature
```

juhaku:
Oh snap, I must have forgotten something..

PR #1039, opened 2024-09-09
Fix building utoipa-rapidoc & utoipa-scalar

jayvdb:
Fixes https://github.com/juhaku/utoipa/issues/1037 (more)

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -3,35 +3,39 @@
 set -e
 
 : "${CARGO:=cargo}"
+: "${CARGO_COMMAND:=test}"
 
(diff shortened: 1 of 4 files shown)


scripts/test.sh
```diff
@@ -3,35 +3,42 @@
 set -e
 
 : "${CARGO:=cargo}"
+: "${CARGO_COMMAND:=test}"
```
jayvdb, line 6: with this it is possible to do

```
CARGO_COMMAND=build ./scripts/test.sh
```

juhaku: Super, this is nice :+1: This provides a quick way to test whether the stack builds.

PR #1041, opened 2024-09-09
Fix testing without explicit features

jayvdb:


jayvdb:
Note there is another test failure - I am working on it

juhaku:
@jayvdb It seems to be that the reason is the line in the `test.sh`
```bash
    # Always test the crate itself first, without any features added.
    $CARGO "${CARGO_COMMAND}" -p "$crate"
```
If I comment this out, it works, what was this for?

scripts/test.sh
```diff
@@ -11,7 +11,9 @@ for crate in $crates; do
     echo "Testing crate: $crate..."
 
     # Always test the crate itself first, without any features added.
-    $CARGO ${CARGO_COMMAND} -p $crate
+    if [[ "$crate" != "utoipa-gen" ]]; then
+        $CARGO ${CARGO_COMMAND} -p $crate
+    fi
```
jayvdb, line 16: So this was added so that we always check that it is possible to build/check/test a crate without any explicit features being added on the command line.

However, that concept does not easy work with utoipa-gen , so I have added an exception here to avoid that.

utoipa-rapidoc/Cargo.toml
```diff
@@ -23,5 +23,8 @@ actix-web = { version = "4", optional = true, default-features = false }
 rocket = { version = "0.5", features = ["json"], optional = true }
 axum = { version = "0.7", default-features = false, features = ["json"], optional = true }
 
+[dev-dependencies]
+utoipa-rapidoc = { path = ".", features = ["actix-web", "axum", "rocket"] }
```
jayvdb, line 27: there is another way to solve this

https://github.com/async-rs/async-std/blob/main/src/lib.rs#L140-L141

```
#![cfg_attr(feature = "attributes", doc = "```")]
#![cfg_attr(not(feature = "attributes"), doc = "```ignore")]
```

That will take a bit longer

scripts/test.sh
juhaku: Ah, True it does not since there are those mutually exclusive features which is unfortunate, but kind of hard to move out.

utoipa-rapidoc/Cargo.toml
juhaku: So this ignores certain parts of the docs for it won't run the doc tests? Interesting concept probably could be leveraged on some cases through. :thinking:

jayvdb: Right - that approach means the doctests are optionally activated by feature flags as well, so all the features dont need to all added to dev-dependencies .

I havent done this before - I just saw the trick last weekend and it looks like it will help here.

juhaku: Neat, it probably could.

scripts/test.sh
```diff
@@ -11,7 +11,9 @@ for crate in $crates; do
     echo "Testing crate: $crate..."
 
     # Always test the crate itself first, without any features added.
-    $CARGO ${CARGO_COMMAND} -p $crate
+    if [[ "$crate" != "utoipa-gen" ]]; then
+        $CARGO ${CARGO_COMMAND} -p $crate
```
juhaku, line 15: Do you mind if you wrap the variable with double quotes. I have a shellcheck enabled and it does complain and rightfully so that unquoted variables might get globbed or splitted. Like follows but all entries that are not double quoted
```suggestion
        $CARGO "${CARGO_COMMAND}" -p $crate
```

juhaku: Though I can also wrap then later as well

jayvdb: Without quotes, the following works
```
CARGO_COMMAND="test --offline" ./scripts/test.sh
```
I see the shellcheck warning and will investigate.

I wonder why it doesnt also complain about `$CARGO`.

juhaku: > I wonder why it doesnt also complain about `$CARGO`.

That's interesting. My version does not error about that. I am actually using the bash lsp server.

juhaku:
Should we give this a go, or do you still do some tweaking? With the shellcheck lints?

jayvdb:
It should be good to merge.

I added a shellcheck ignore, as it is more useful to allow the variable to expand.

juhaku:
Alright, let's get the show on the road.

Issue #1042, opened 2024-09-10
Axum Swagger UI Handler Redirect Causing Invalid URI for "/"


kevinlutzer:
Hey Guys! I love this library, thanks for building and maintaining it! 🥳 

## Issue

Serving the Swagger UI via Axum on "/" causes a 303 redirect to an invalid URI: http:///.  When instantiation the Swagger UI with `SwaggerUI::new("/someroute")`, the internal router will always redirect to "/someroute/". 

The line of code in question: https://github.com/juhaku/utoipa/blob/4422cee81cf6b0811c4126a17cf800f7447298d1/utoipa-swagger-ui/src/axum.rs#L52C17-L52C24

Note that other routing tools like Actix do not do this: https://github.com/juhaku/utoipa/blob/4422cee81cf6b0811c4126a17cf800f7447298d1/utoipa-swagger-ui/src/actix.rs#L26C8-L37C35

## Solution

Remove the redirect and serve the Axum UI on `<route name>`, `<route name>/`, and `<route name>/*rest`. 

## Sample To Replicate

Main:

``` rust
use axum::{routing::get, Router};

use utoipa::OpenApi;
use utoipa_swagger_ui::SwaggerUi;

#[derive(OpenApi)]
#[openapi()]
struct ApiDoc;

/// Get health of the API.
#[utoipa::path(
    get,
    path = "/foo",
    responses(
        (status = OK, description = "Success", body = str, content_type = "text/plain")
    )
)]
async fn foo_handler() -> &'static str {
    "bar"
}

#[tokio::main]
async fn main() {
    let app = Router::new()
        .route("/foo", get(foo_handler))
        .merge(SwaggerUi::new("/").url("/api-docs/openapi.json", ApiDoc::openapi()));

    // run it
    let listener = tokio::net::TcpListener::bind("127.0.0.1:3000")
        .await
        .unwrap();
    println!("listening on {}", listener.local_addr().unwrap());

    axum::serve(listener, app).await.unwrap();
}
``` 
Cargo Manifest:

```toml
[package]
name = "utoipasample"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
axum = "0.7.5"
tokio = { version = "1.40.0", features = ["full"] }
utoipa = { version = "4.2.3", features = ["axum_extras"] }
utoipa-axum = "0.1.0-beta.2"
[...]

PR #1043, opened 2024-09-10
Remove Redirect Causing Invalid URIs for Swagger UIs Server on /

kevinlutzer:
Fixes https://github.com/juhaku/utoipa/issues/1042


juhaku:
This is okay, the tests were broken on master. But should work now.

PR #1045, opened 2024-09-11
Always use system `curl` by default when building `utoipa-swagger-ui`

### LINKED ISSUE #1044: `utoipa-swagger-ui`: cross compile to Windows on Linux fails
When cross-compiling to `x86_64-pc-windows-gnu` target on Linux, `utoipa-swagger-ui` fails to build with the following error.

```
error: failed to run custom build command for `utoipa-swagger-ui v7.1.0`

Caused by:
  process didn't exit successfully: `/repo/target/release/build/utoipa-swagger-ui-a31a62fed59f60ee/build-script-build` (exit status: 101)
  --- stdout
  OUT_DIR: /repo/target/x86_64-pc-windows-gnu/release/build/utoipa-swagger-ui-31dfd8e5fc31018f/out
  SWAGGER_UI_DOWNLOAD_URL: https://github.com/swagger-api/swagger-ui/archive/refs/tags/v5.17.12.zip
  start download to : "/repo/target/x86_64-pc-windows-gnu/release/build/utoipa-swagger-ui-31dfd8e5fc31018f/out/v5.17.12.zip"
  cargo:rerun-if-env-changed=SWAGGER_UI_DOWNLOAD_URL
  reqwest feature: Err(NotPresent)

  --- stderr
  thread 'main' panicked at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/utoipa-swagger-ui-7.1.0/build.rs:183:81:
  called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

I took a quick look at `build.rs`, and it seems like the original author made some implicit assumptions about tooling availability when targeting different OSes, but did not take into account the possibility of cross-compilation.

cyqsimon:
Also fixes #1044

### DIFF
--- a/utoipa-swagger-ui/Cargo.toml
+++ b/utoipa-swagger-ui/Cargo.toml
@@ -45,20 +45,12 @@ zip = { version = "2", default-features = false, features = ["deflate"] }
 regex = "1.7"
 
 # enabled optionally to allow rust only build with expense of bigger dependency tree and platform
-# independant build. By default `curl` system package is tried for downloading the Swagger UI.
+# independent build. By default `curl` system package is tried for downloading the Swagger UI.
 reqwest = { version = "0.12", features = [
     "blocking",
     "rustls-tls",
 ], default-features = false, optional = true }
 url = { version = "2", optional = true }
-utoipa-swagger-ui-vendored = { version = "0.1", path = "../utoipa-swagger-ui-vendored", optional = true }
-
-# Windows is forced to use reqwest to download the Swagger UI.
-[target.'cfg(windows)'.build-dependencies]
-reqwest = { version = "0.12", features = [
-    "blocking",
-    "rustls-tls",
-], default-features = false }
 utoipa-swagger-ui-vendored = { version = "0.1", path = "../utoipa-swagger-ui-vendored", optional = true }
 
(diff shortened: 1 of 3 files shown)


cyqsimon:
It seems to me that the original author of this code assumed conditional compilation within the build script (`#[cfg]`/`cfg!()`) and the environment variables presented to the build script at runtime (`CARGO_CFG_*`) can be used interchangeably, whereas in reality there is a semantic difference that becomes important in cross-compilation.

I haven't bothered scrutinizing other parts of `build.rs` (or for that matter, the build script of other crates), but I suspect there could be similar oversights too. Maybe this is something worth looking over.

juhaku (approved):
Great, it is much cleaner as so. :tada:

juhaku:
> I haven't bothered scrutinizing other parts of build.rs (or for that matter, the build script of other crates), but I suspect there could be similar oversights too. Maybe this is something worth looking over.

Sure if you see it worthwhile. (Though other crates does not have any build scripts as of this moment). The cross platform support is mainly relied on community support. Since I am only a linux user. And some choises I have made is based only on assumption of how it suppose to work :sweat_smile:

PR #1048, opened 2024-09-12
Add global config for `utiopa`

### LINKED ISSUE #429: Type aliases not being resolved correctly
Using the type alias

```rust
pub type DateTimeUtc = chrono::DateTime<chrono::Utc>;
```

Leads to any serialized struct field having the type:
```js
{recursive: DateTimeUtc}
```

Directly specifying the type as `DateTime<Utc>` in the struct leads to the type being correctly resolved. (type shows as `date-time`)

### LINKED ISSUE #547: Type Aliases in OpenApi: Resolver error
My struct has a field which is a type alias.
```rust
use serde::Deserialize;
use serde::Serialize;

type MeeasureName = String;

pub fn example() {
    ()
}


#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash, utoipa::ToSchema)]
#[serde(tag = "type")]
pub struct AggregationRequest {
    pub name: MeeasureName,
    pub measures: Vec<(String, String)>
}
```

I expose this struct to OpenApi like so:
```rust
#[derive(OpenApi)]
#[openapi(
    info(
        title = "Ultima BI"
    ),
    paths(
        routers::execute
    ),
    components(
        schemas(AggregationRequest)
    ),
    tags(
        (name = "Ultima BI", description = "Ultimate Business Intellegence endpoints.")
    ),
)]
pub(crate) struct ApiDoc;
```

What I see in the Swagger UI(ignore irrelevant stuff):
image

### LINKED ISSUE #766: Allow alias for serde_json::Value: Gobal Rusts Type Alias registry
Hello.

Instead of using `serde_json::Value` directly, we added an alias `Json` for `serde_json::Value`.

But the `Json` won't be recognized as `serde_json::Value`.

```rust
pub type Json = serde_json::Value;
```

This would make it recognizable, but that doesn't make much sense.

```rust
pub struct VO {
  #[schema(value_type = HashMap<String, serde_json::Value>)]
  pub fields: HashMap<String, Json>,
}
```

Could it be like `#[aliases]`, where to write it once, and it's consistently recognized as `serde_json::Value`?

### LINKED ISSUE #787: Support for rust type aliases
Perhaps there's a way already to workaround this, I just could not find it.

I'm working on a code generator for Redfish schemas that produces some code like this:

```
#[derive(Deserialize, Serialize, ToSchema, Builder, Clone, Debug)]
#[builder(setter(strip_option))]
#[builder(build_fn(error = "ModelBuildError"))]
pub struct IPv6Address {
    /// The IPv6 address.
    #[serde(rename = "Address")]
    #[builder(default)]
    pub address: Option<String>,

    // ... other stuff
    /// The IPv6 address prefix Length.
    #[serde(rename = "PrefixLength")]
    #[builder(default)]
    pub prefix_length: Option<PrefixLength>,              ///// <<<<< NOTICE THE USAGE OF PrefixLength here
}
```

The PrefixLength is generated as this:
```
pub type PrefixLength = i32;
```

When I generate the swagger-ui, there's this error for types like that:

```
Resolver error at paths./redfish/v1/Managers/{manager_id}/EthernetInterfaces/{ifc_id}.get.responses.200.content.application/json.schema.items.properties.IPv6StaticAddresses.items.properties.PrefixLength.$ref
Could not resolve reference: Could not resolve pointer: /components/schemas/PrefixLength does not exist in document
```

As it's an alias to a primitive type, I can't put the #[derive(ToSchema)] on it.

I could create the aliases as struct wrappers, but it would be a bit of a hassle to use it.

I'm about to change my codegenerator to do some lookup for aliases and translate it directly in the struct definition, but decided to check if there's another option available.

Thank you.

### LINKED ISSUE #984: `ToSchema` not implemented for primitive if the primitive is behind a type alias
You could test this trivially in an example from the repo:

`utoipa/examples/todo-axum/src/main.rs`: modify 
```rust
    /// Item to do.
    #[derive(Serialize, Deserialize, ToSchema, Clone)]
    pub(super) struct Todo {
        id: i32,
        #[schema(example = "Buy groceries")]
        value: String,
        done: bool,
    }
```

to 
```rust
    type OurNumber = u32;

    /// Item to do.
    #[derive(Serialize, Deserialize, ToSchema, Clone)]
    pub(super) struct Todo {
        id: OurNumber,
        #[schema(example = "Buy groceries")]
        value: String,
        done: bool,
    }
```

First, Swagger Docs will throw an error
```
Could not resolve reference: Could not resolve pointer: /components/schemas/OurNumber does not exist in document
```
and `Todo` will show a missing Schema component for its `id` field

If you try to add new type alias to the `ApiDoc` annotation like this:
```
        components(
            schemas(todo::Todo, todo::TodoError, todo::OurNumber)
        ),
```

you'll get
```rust
error[E0277]: the trait bound `u32: ToSchema<'_>` is not satisfied
   --> src/main.rs:31:50
    |
31  |             schemas(todo::Todo, todo::TodoError, todo::OurNumber)
    |                                                  ^^^^ the trait `ToSchema<'_>` is not implemented for `u32`
```

Expected behaviour: well I'd expect `ToSchema` to be implemented for primitives. 

Is there a way to work around this?

juhaku:
This PR adds global config for `utiopa` which allows defining global Rust type aliases that can be recognized by `utoipa`.

```rust
 let _ = Config::new()
     .alias_for("MyType", "i32");
     .alias_for("Json", "Value");
     .alias_for("NullableString", "Option<String>");
```

Closes #984 Closes #787 Closes #766 Closes #429 Closes #547

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -26,6 +26,7 @@ jobs:
           - utoipa-rapidoc
           - utoipa-scalar
           - utoipa-axum
+          - utoipa-config
       fail-fast: true
     runs-on: ubuntu-latest
 
@@ -68,6 +69,8 @@ jobs:
             elif [[ "$change" == "utoipa-scalar" && "${{ matrix.crate }}" == "utoipa-scalar" && $changes == false ]]; then
               changes=true
             elif [[ "$change" == "utoipa-axum" && "${{ matrix.crate }}" == "utoipa-axum" && $changes == false ]]; then
+              changes=true
+            elif [[ "$change" == "utoipa-config" && "${{ matrix.crate }}" == "utoipa-config" && $changes == false ]]; then
               changes=true
             fi
           done < <(git diff --name-only ${{ github.sha }}~ ${{ github.sha }} | grep .rs | awk -F \/ '{print $1}')
--- a/.github/workflows/draft.yaml
+++ b/.github/workflows/draft.yaml
@@ -21,6 +21,7 @@ jobs:
           - utoipa-rapidoc
           - utoipa-scalar
           - utoipa-axum
+          - utoipa-config
     runs-on: ubuntu-latest
 
     steps:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,10 +12,12 @@ members = [
     "utoipa-rapidoc",
     "utoipa-scalar",
     "utoipa-axum",
+    "utoipa-config",
 ]
 
 [workspace.metadata.publish]
 order = [
+    "utoipa-config",
     "utoipa-gen",
     "utoipa",
     "utoipa-swagger-ui-vendored",
--- a/README.md
+++ b/README.md
@@ -109,6 +109,7 @@ and the `ipa` is _api_ reversed. Aaand... `ipa` is also an awesome type of beer 
 - **`non_strict_integers`**: Add support for non-standard integer formats `int8`, `int16`, `uint8`, `uint16`, `uint32`, and `uint64`.
 - **`rc_schema`**: Add `ToSchema` support for `Arc<T>` and `Rc<T>` types. **Note!** serde `rc` feature flag must be enabled separately to allow
   serialization and deserialization of `Arc<T>` and `Rc<T>` types. See more about [serde feature flags](https://serde.rs/feature-flags.html).
+- **`config`** Enables [`utoipa-config`](./utoipa-config/README.md) for the project which allows defining global configuration options `utoipa`.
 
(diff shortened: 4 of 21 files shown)


juhaku:
> I'm not satisfied with this approach since it is still string based match-and-replace and need users to explicitly list all type aliases used. It is error-prone and frustrating when there are many nested type aliases that may also involving generics.

The string alias value will be parsed to `TypeTree` which is then processed same way as the type itself was written instead. I am not sure what actually can be done more to support the aliases, because they cannot be treated as real types and they cannot implement `ToSchema` nor we know what type it actually is at compile time when parsing the token stream. 

It is true that the aliases this way does not really support generic aliases. But I am not sure whether they could ever be supported. It would need some experimenting. My advice is to use the aliases sparingly. 

If this support causes more trouble to users perhaps it could be removed as well. :thinking:

PR #1059, opened 2024-09-21
Refactor enums processing

### LINKED ISSUE #324: Support mapping for Discriminator object.
Hi, first of all thanks for the amazing work 💯!

I would like to work on supporting defining custom mapping for Discriminator object. Here's a [minimally working PR](https://github.com/getditto/utoipa/pull/1/files) on a fork according to the below proposal (**TLDR;** `#[schema(mapping = "...")]`). 

The PR only works for Unnamed Field Struct and hardcoding the value of the mapping. If it's keen to have this feature added as part of utoipa, do let me know and I'm happy to open a PR here and work on it according to any suggestions.

### Context 
Here's a example use case:

```rust
#[derive(ToSchema, Debug, Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
pub struct CreateResult {
    pub id: String,
    pub amount: u64
}

#[derive(ToSchema, Debug, Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
pub struct UpdateResult {
    pub id: String,
    pub amount: u64
}

#[derive(ToSchema, Debug, Serialize, Deserialize)]
#[serde(tag = "method", rename_all = "camelCase")]
pub enum Result {
    Create(CreateResult),
    Update(UpdateResult),
}
```

Using `utoipa`, the following would be generated:

```json
{
  "oneOf": [
    {
      "allOf": [
        {
          "$ref": "#/components/schemas/CreateResult"
        },
        {
          "type": "object",
          "required": [
            "method"
          ],
          "properties": {
            "method": {
              "type": "string",
              "enum": [
                "create"
              ]
            }
          }
        }
      ]
    },
    {
      "allOf": [
        {
          "$ref": "#/components/schemas/DeleteResult"
        },
        {
          "type": "object",
          "required": [
            "method"
          ],
          "properties": {
            "method": {
              "type": "string",
              "enum": [
                "delete"
              ]
            }
          }
        }
      ]
    }
  ],
  "discriminator": {
    "propertyName": "method"
  }
}
```

and the expected json will be as follow:

```json
{
  "method": "create | delete",
  "id": "1",
}
```

However. according to the [Discriminator spec](https://spec.openapis.org/oas/latest.html#discriminator-object), it would need to method would need to match either `CreateResult` or `DeleteResult`. This is where `mapping` come in and help with it. Without the custom mapping, the generated client using OpenAPI geneartor will not work as expected.

### Proposal

Support additonal `mapping` attribute in `schema` macro:

```rust
#[derive(ToSchema, Debug, Serialize, Deserialize)]
#[serde(tag = "method", rename_all = "camelCase")]
pub enum Result {
    #[schema(mapping = "create")]
    Create(CreateResult),
    #[schema(mapping = "update")]
    Update(UpdateResult),
}
```
which will generate the following:

```json
{
  "oneOf": [
    {
      "allOf": [
        {
          "$ref": "#/components/schemas/CreateResult"
        },
        {
          "type": "object",
          "required": [
            "method"
          ],
          "properties": {
            "method": {
              "type": "string",
              "enum": [
                "create"
              ]
            }
          }
        }
      ]
    },
    {
      "allOf": [
        {
          "$ref": "#/components/schemas/DeleteResult"
        },
        {
          "type": "object",
          "required": [
            "method"
          ],
          "properties": {
            "method": {
              "type": "string",
              "enum": [
                "delete"
              ]
            }
          }
        }
      ]
    }
  ],
  "discriminator": {
    "propertyName": "method",
    "mapping": {
      "create": "#/components/schemas/CreateResult",
      "delete": "#/components/schemas/DeleteResult"
    }
  }
}
```
I'm not sure if this is the best way so do let me know if there's any better alternatives. Some of the inputs I might need is how do we handle the different kind of Enums we have and also what if those reference struct are not derive with `ToSchema`.

### LINKED ISSUE #346: Invalid inline schema defined in oneOf according to spec
Trying to use OpenAPI Generator CLI for my API which has some enums with type tag. I'm getting the following error:

```
WARN  o.o.codegen.DefaultCodegen - Invalid inline schema defined in oneOf/anyOf in 'CreateQuestionData'. Per the OpenApi spec, for this case when a composed schema defines a discriminator, the oneOf/anyOf schemas must use $ref. Change this inline definition to a $ref definition
```

This leads to invalid type generation for enums. I am assuming this is because utoipa generates an inline schema for enums whereas according to the spec it should use $refs?

Any workarounds?

### LINKED ISSUE #617: Schema for internally tagged structs does not adhere to OpenAPI spec
I would like to use the following enum (note that `NumberParam` has a "tag" parameter):

```rust
#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Clone, ToSchema)]
#[serde(rename_all = "camelCase", tag = "type")]
pub enum NumberParam {
    Static { value: usize },
    Derived(DerivedNumber),
}

#[derive(Debug, PartialEq, Serialize, Deserialize, Clone, ToSchema)]
#[serde(rename_all = "camelCase")]
pub struct DerivedNumber {
    pub attribute: String,
    pub factor: f64,
    pub default_value: f64,
}
```

At the moment Utoipa will generate this schema:

```json
{
  "openapi": "3.0.3",
  "info": {
    "title": "Dummy API",
    "version": "0.7.0"
  },
  "paths": {
  },
  "components": {
    "schemas": {
      "NumberParam": {
        "oneOf": [
          {
            "type": "object",
            "required": [
              "value",
              "type"
            ],
            "properties": {
              "type": {
                "type": "string",
                "enum": [
                  "static"
                ]
              },
              "value": {
                "type": "integer",
                "minimum": 0.0
              }
            }
          },
          {
            "allOf": [
              {
                "$ref": "#/components/schemas/DerivedNumber"
              },
              {
                "type": "object",
                "required": [
                  "type"
                ],
                "properties": {
                  "type": {
                    "type": "string",
                    "enum": [
                      "derived"
                    ]
                  }
                }
              }
            ]
          }
        ],
        "discriminator": {
          "propertyName": "type"
        }
      },
      "DerivedNumber": {
        "type": "object",
        "required": [
          "attribute",
          "factor",
          "defaultValue"
        ],
        "properties": {
          "attribute": {
            "type": "string"
          },
          "defaultValue": {
            "type": "number",
            "format": "double"
          },
          "factor": {
            "type": "number",
            "format": "double"
          }
        }
      }
    }
  }
}
```

As you can see, utoipa inlines the schemas of the `oneOf` variants. This is against the spec when there is a discriminator specified. The `oneOf` should only contain refs. This can also be seen when trying to generate a python client using the official program `openapi-generator`. It throws many errors and warnings related to this:

```
$ java -jar openapi-generator-cli.jar generate -i openapi_test.json -g python -o backend

[main] WARN  o.o.codegen.DefaultCodegen - Invalid inline schema defined in oneOf/anyOf in 'NumberParam'. Per the OpenApi spec, for this case when a composed s
chema defines a discriminator, the oneOf/anyOf schemas must use $ref. Change this inline definition to a $ref definition
[main] WARN  o.o.codegen.utils.ModelUtils - Failed to get the schema name: null
[main] ERROR o.o.codegen.DefaultCodegen - String to be sanitized is null. Default to ERROR_UNKNOWN
[main] ERROR o.o.codegen.DefaultCodegen - String to be sanitized is null. Default to ERROR_UNKNOWN
[main] ERROR o.o.codegen.DefaultCodegen - Failed to lookup the schema 'null' when processing oneOf/anyOf. Please check to ensure it's defined properly.       
[main] ERROR o.o.codegen.DefaultCodegen - Failed to lookup the schema 'null' when processing oneOf/anyOf. Please check to ensure it's defined properly.       
[main] WARN  o.o.codegen.DefaultCodegen - Invalid inline schema defined in oneOf/anyOf in 'NumberParam'. Per the OpenApi spec, for this case when a composed s
chema defines a discriminator, the oneOf/anyOf schemas must use $ref. Change this inline definition to a $ref definition
[main] ERROR o.o.codegen.DefaultCodegen - Failed to lookup the schema 'null' when processing oneOf/anyOf. Please check to ensure it's defined properly.       
[main] ERROR o.o.codegen.DefaultCodegen - Failed to lookup the schema 'null' when processing oneOf/anyOf. Please check to ensure it's defined properly.
```

To solve this problem, utoipa would need to generate named schemas for each variant and also specify a custom mapping. That can look like this:

```json
{
  "openapi": "3.0.3",
  "info": {
    "title": "Dummy API",
    "version": "0.7.0"
  },
  "paths": {
  },
  "components": {
    "schemas": {
      "NumberParam": {
        "oneOf": [
          {
            "$ref": "#/components/schemas/NumberParamStaticWithType"
          },
          {
            "$ref": "#/components/schemas/DerivedNumberWithType"
          }
        ],
        "discriminator": {
          "propertyName": "type",
          "mapping": {
            "static": "#/components/schemas/NumberParamStaticWithType",
            "derived": "#/components/schemas/DerivedNumberWithType"
          }
        }
      },
      "DerivedNumber": {
        "type": "object",
        "required": [
          "attribute",
          "factor",
          "defaultValue"
        ],
        "properties": {
          "attribute": {
            "type": "string"
          },
          "defaultValue": {
            "type": "number",
            "format": "double"
          },
          "factor": {
            "type": "number",
            "format": "double"
          }
        }
      },
      "DerivedNumberWithType": {
        "allOf": [
          {
            "$ref": "#/components/schemas/DerivedNumber"
          },
          {
            "type": "object",
            "required": [
              "type"
            ],
            "properties": {
              "type": {
                "type": "string",
                "enum": [
                  "derived"
                ]
              }
            }
          }
        ]
      },
      "NumberParamStaticWithType": {
        "type": "object",
        "required": [
          "value",
          "type"
        ],
        "properties": {
          "type": {
            "type": "string",
            "enum": [
              "static"
            ]
          },
          "value": {
            "type": "integer",
            "minimum": 0.0
          }
        }
      }
    }
  }
}
```

What are your thoughts on this? Can this feature be added to utoipa?

### LINKED ISSUE #621: Overly complicated discriminator
We have a Rust struct, where all of the variants have a member `report_type`.
```rust
    #[derive(..., utoipa::ToSchema)]
    #[serde(tag = "report_type")]
    pub enum ReportData {
        Report1(Report1),
        Report2(Report2),
        Report3(Report3),
        ...
    }
```
utoipa emits
```yaml
    ReportData:
      oneOf:
        - allOf:
            - $ref: "#/components/schemas/Report1"
            - type: object
              required:
                - report_type
              properties:
                report_type:
                  type: string
                  enum:
                    - Report1
        - allOf:
            - $ref: "#/components/schemas/Report2"
            - type: object
              required:
                - report_type
              properties:
                report_type:
                  type: string
                  enum:
                    - Report2
        - allOf:
            - $ref: "#/components/schemas/Report3"
            - type: object
              required:
                - report_type
              properties:
                report_type:
                  type: string
                  enum:
                    - Report3
      ...
      discriminator:
        propertyName: report_type

```

As the `report_type` fields in each schema, it should be

```yaml
    ReportData:
      oneOf:
        - $ref: "#/components/schemas/Report1"
        - $ref: "#/components/schemas/Report2"
        - $ref: "#/components/schemas/Report3"
      discriminator:
        propertyName: report_type
```

juhaku:
This commit refactors enum processing from ground up. This fully unifies the how each enum variant schema is resolved and the way variant schemas are generated. This will result easier debugging, changing and updating the enum processing in future and most of all the enums now will behave consistently due to removing bunch of duplication and adding correct abstractions instead.

This commit also unifies previously known `SimpleEnum` and `ReprEnum` to a single enum to furhter simplify the code. Also `ComplexEnum` is now known by `MixedEnum`.

This commit implements discriminator with support for custom mapping. Discriminator can only be used with `#[serde(untagged)]` enum having only unnamed field variants with one schema reference implementing `ToSchema` trait. It cannot be used with primitive types nor with inlined schemas.

Removed `#[serde(tag = ...)]` as discriminator support.

Update docs and add support for missing features for enum variants such as `Title`, `Deprecated`, `MinProperties` and `MaxProperties`.

 ### Breaking changes

* `#[serde(tag = ...)]` will not be used as discriminator.

Example of new discriminator syntax. It can be defined with discriminator field or optionally with custom mappping.
```rust
#[schema(discriminator = "bar_type")]
#[serde(untagged)]
enum BarInternal {
    Baz(BarBarInternal),
    FooBar(FooInternal),
}

// with custom mapping
#[schema(discriminator(property_name = "bar_type", mapping(
    ("bar" = "#/components/schemas/BarBarInternal"),
    ("foo" = "#/components/schemas/FooInternal"),
)))]
#[serde(untagged)]
enum BarInternal {
    Baz(BarBarInternal),
    FooBar(FooInternal),
}
```

Fixes #621 Fixes #324
Closes #617 Closes #346


1lutz:
Does that mean `BarBarInternal` and `FooInternal` structs need to include a string field `bar_type`?

juhaku:
> Does that mean `BarBarInternal` and `FooInternal` structs need to include a string field `bar_type`?

Yes, as it is by the OpenAPI spec.

1lutz:
Then I don't like this solution. I don't think there must be a 1:1 relation between the OpenAPI Schema and Rust code. You normally would not write code like in the following:

```rust
struct Dog {
    pet_type: String
}
```

juhaku:
That is true, however the enums work as previously they did, just the `#[serde(tag = ...)]` does not generate the discriminator field anymore. This at the beginning was wrong because the schema that is generated with the `tag` does include polymorphism with inlined schemas. This is wrong as the discriminator will not be adhered to inlined schemas. 

This previous "broken" discriminator implementation has caused errors the users have been facing and led to manual mapping and hacks anyway. Thus it is better to remove the broken implementation in the first place.

The assessment of supporting `#[serde(tag = ...)]` as a discriminator needs to be done in future and is in a better state after this PR code wise. To fully support such case still needs further work and planning of how to correctly leverage the `#[serde(tag = ...)]` attribute to generate the discriminator field in such a way that the generated spec is correct and will not cause issues.

1lutz:
When you assess the support of `#[serde(tag = ...)]` as a discriminator, you may look into what I programmed for my company: https://github.com/geo-engine/geoengine/blob/46a5336c0ec194fe67594b3727fad784873d3b38/services/src/util/apidoc.rs#L155

It takes the "broken" enums with inline schemas and a discriminator, generates new variant schemas with a nice name (or in some case edits existing schemas), links them in the enum schema and adapts the discriminator on-the-fly. There are tests below which show the results of the transformation.

Maybe you would like to integrate something like this in the future.

juhaku:
That is pretty neat. Sure something like this could be evaluated at least to the level of supporting unnamed field variants having one `ToSchema` reference. For what comes to the inlined variants like named field variant, it might be quite hard at compile time to convert it to a schema that suppose to be then added to the spec. Before even thinking such possibility I need to first solve the problem of registering schemas from their usage such as definition on `#[utoipa::path(....)]` macro and nested schemas within `ToSchema` types. This would most likely give me fundamentals to work on such functionality. 

But implementing this kind of transformer internally should not be too big of a thing though. Nevertheless there is a need to assess whats feasible and whats possible.

ChristianBeilschmidt:
> Removed #[serde(tag = ...)] as discriminator support.
Why? This is the common case for using serde to deserialize API calls with enums.

Is there a workaround to not change all our enums to untagged just because of OpenAPI specs?

Issue #643, opened 2023-06-09
How do I add custom x-properties


coffeebe4code:
I am using the actix example essentially.

We have custom requirements at our company to add several extra fields, usually for certain blocks. like the server block, or extra x-properties in the info block.

```rs
#[utoipa::path(
    get,
    path = "/echo",
    responses(
        (status = 200, description = "echos back all headers and request body in response"),
    ),
)]
#[get("/echo")]
```

```rs
    #[derive(OpenApi)]
    // ?insert more props on openapi?
    #[openapi(paths(echo))]
    struct ApiDoc;

    let mut openapi = ApiDoc::openapi();
    // ? insert more props on openapi?
    let mut file = std::fs::File::create("openapi.json")?;
    file.write_all(openapi.to_pretty_json().unwrap().as_bytes())?;
```

I would much rather build this file at compile time as well, but I can manage that on my end if there are no exposed means from this project.

Great freaking project btw, this all worked perfectly out of the box very quick to set up.

juhaku:
In short, at the moment `x-properties` are not supported and there is no build in way to insert extension properties to the schema. 

Perhaps this is something that could be explored in future, and of course PR:s are welcome. :slightly_smiling_face:

rlukata:
@coffeebe4code You can build the server object using [the modify trait](https://docs.rs/utoipa/latest/utoipa/trait.Modify.html).

lucasgranberg:
https://github.com/juhaku/utoipa/pull/763
Are these the same x-properties? My PR is for openapi extensions

juhaku:
@lucasgranberg Yes, 

@coffeebe4code `Info` block supports extensions now, but `Server` does not have support for those as of yet. Though adding that supports is quite quick
.

lucasgranberg:
https://github.com/juhaku/utoipa/commit/b62efb45c7d75f539024306128d073e1e1cd7286

Pretty much this but my fork is too messed up to merge, better to do separately

juhaku:
Right, 

It's worth to check https://spec.openapis.org/oas/latest.html which objects supports extensions and add on the same go the support for them if applicable. (I believe that not all objects can support in utoipa those extensions as of current state types.)
![image](https://github.com/user-attachments/assets/449c9bcf-8a0f-4847-97df-6869609d657d)

juhaku:
This PR adds the extensions to types currently missing the field #1104.

PR #1062, opened 2024-09-25
Add explicit `Extensions` type

heat1q:
I noticed when deserializing an existing spec into `OpenApi` that some fields are getting duplicated into the extensions. This seems to happen because extensions are currently defined as a simple `HashMap<String, serde_json::Value>` (see #763) which is flattened at the respective level. This causes serde to also fill the fields.

<details><summary>Example</summary>

Given the following spec

```json
{
  "openapi": "3.1.0",
  "info": {
    "title": "Sample API",
    "description": "API description in Markdown.",
    "version": "1.0.0"
  },
  "servers": [
    {
      "url": "https://api.example.com/v1"
    }
  ],
  "paths": {
    "/users": {
      "get": {
        "summary": "Returns a list of users.",
        "description": "Optional extended description in Markdown.",
        "responses": {
          "200": {
            "description": "OK",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object"
                }
              }
            }
          }
        }
      }
    }
  },
  "components": {}
}
```

deserializing it into `OpenApi`
```rust
let openapi: OpenApi = serde_json::from_str("{...}").unwrap();
println!("{#?}", openapi.extensions);
```
We see that the `extensions` fields get populated even though no extensions are specified.
```
Some(
    {
        "paths": Object {
            "/users": Object {
                "get": Object {
                    "description": String("Optional extended description in Markdown."),
                    "responses": Object {
                        "200": Object {
                            "content": Object {
                                "application/json": Object {
                                    "schema": Object {
                                        "type": String("object"),
                                    },
                                },
                            },
                            "description": String("OK"),
                        },
                    },
                    "summary": String("Returns a list of users."),
                },
            },
        },
    },
)
```
because of how extensions (or rather the deserialize behavior) is specified
https://github.com/juhaku/utoipa/blob/e0c5aa749ac548cca7fd75544293865efe5d0231/utoipa/src/openapi.rs#L129-L131

</details> 

### Proposed Solution
We wrap the extensions as they are currently implemented in an explicit type for which we control the `Deserialize` behavior, that is, we only allow key-value pair that have the specific `x-` key prefix mentioned in the [spec](https://spec.openapis.org/oas/latest.html#specification-extensions).

### DIFF
--- a/utoipa/src/openapi/extensions.rs
+++ b/utoipa/src/openapi/extensions.rs
@@ -1,0 +1,97 @@
+//! Implements [OpenAPI Extensions][external_docs].
+//!
+/// [extensions]: https://spec.openapis.org/oas/latest.html#specification-extensions
+use std::{
+    collections::HashMap,
+    ops::{Deref, DerefMut},
+};
+
+use serde::Serialize;
+
+use super::builder;
+
+const EXTENSION_PREFIX: &str = "x-";
+
+builder! {
+    ExtensionsBuilder;
+
+    /// Additional [data for extending][extensions] the OpenAPI specification.
+    ///
+    /// [extensions]: https://spec.openapis.org/oas/latest.html#specification-extensions
+    #[derive(Default, Serialize, Clone, PartialEq, Eq)]
+    #[cfg_attr(feature = "debug", derive(Debug))]
+    pub struct Extensions{
+        #[serde(flatten)]
+        extensions: HashMap<String, serde_json::Value>,
+    }
+}
+
+impl Deref for Extensions {
+    type Target = HashMap<String, serde_json::Value>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.extensions
+    }
+}
+
+impl DerefMut for Extensions {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.extensions
+    }
+}
+
+impl From<Extensions> for HashMap<String, serde_json::Value> {
+    fn from(value: Extensions) -> Self {
+        value.extensions
+    }
+}
+
+impl<'de> serde::de::Deserialize<'de> for Extensions {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let extensions: HashMap<String, _> = HashMap::deserialize(deserializer)?;
+        let extensions = extensions
+            .into_iter()
+            .filter(|(k, _)| k.starts_with(EXTENSION_PREFIX))
+            .collect();
+        Ok(Self { extensions })
+    }
+}
+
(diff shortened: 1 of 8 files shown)


utoipa/src/openapi/extensions.rs
```diff
@@ -0,0 +1,97 @@
...
+            .filter(|(k, _)| k.starts_with(EXTENSION_PREFIX))
+            .collect();
+        Ok(Self { extensions })
+    }
+}
+
+impl ExtensionsBuilder {
+    /// Adds a key-value pair to the extensions. Extensions keys are prefixed with `"x-"` if
+    /// not done already.
+    pub fn add<K, V>(mut self, key: K, value: V) -> Self
```
heat1q, line 66: I'm unsure about the `Extensions` public API. Should we expose the underlying `HashMap`? Or should it only be accessed via setters/getters to ensure that the extensions keys have the correct format?

juhaku: I think that is good for the builder to have this kind of adding method. However as what comes to the `Extension` type itself I think it is good to follow the common practice overall and just expose the fields as public like other types does. 

And for `Extension` there is nothing to expose as it is exposed via `Deref` and `DerefMut` directly so I think this is good as is. When the field is exposed as is it allows users to move vast approaches to operate on the underlying type. Though with the expense of not enforcing the `x- prefix for keys` as defined by the specification.

juhaku (approved):
Great, thanks for the PR :trophy: Looks legit.

However as a side not for future, Currently the extensions are not present in all places they suppose to be supported. It would be worth to add these fields to all types supporting the extensions. See issue #643

utoipa/src/openapi/extensions.rs
juhaku: As a second though I have no strict opinion whether the `getter/setter` should be enforced for the `Extension` type as well. At least then it could enforce the `x-` prefix, but at the same time is limiting how users can actually operate on the type. E.g. I believe it would be useful for example to allow adding extensions from `iter` as well directly to the map and to be able to operate directly with the map via Rust's Iterator API.

juhaku:
@heat1q Do you think this is ready for merge? Or do you have any input for this one?

heat1q:
I've added a `FromIterator` implementation for completeness as suggested. Feel free to merge it if it's fine for you.

>However as a side not for future, Currently the extensions are not present in all places they suppose to be supported. It would be worth to add these fields to all types supporting the extensions. See issue https://github.com/juhaku/utoipa/issues/643

I think with a proper `Extensions` type we now have the option to offer a nice API to add extensions to objects when creating specs. This might be something I can have a look in the future.

juhaku:
Nice, this is good.

juhaku:
Related #1062

PR #1079, opened 2024-10-03
Generate `ToSchema` generic bounds on impl instead of requiring it on definition

### LINKED ISSUE #1078: [5.0.0-rc] [generics]: add `T: ToSchema` bounds on `impl` instead of requiring it on definition?
I have generic types like this:

```rust
#[derive(utoipa::ToSchema)]
struct Signed<T> {
    payload: T,
    sig: String,
}

#[utoipa::path(post, path = "/api", request_body = Signed<String>)]
async fn post_message() {}
```

It fails to compile with:
```
error[E0277]: the trait bound `T: PartialSchema` is not satisfied
 --> src/main.rs:3:14
  |
3 |     payload: T,
  |              ^ the trait `ComposeSchema` is not implemented for `T`, which is required by `T: PartialSchema`
  |
  = note: required for `T` to implement `PartialSchema`
help: consider restricting type parameter `T`
  |
2 | struct Signed<T: utoipa::__dev::ComposeSchema> {
  |                ++++++++++++++++++++++++++++++

error[E0277]: the trait bound `T: ToSchema` is not satisfied
 --> src/main.rs:3:14
  |
3 |     payload: T,
  |              ^ the trait `ToSchema` is not implemented for `T`
  |
help: consider restricting type parameter `T`
  |
2 | struct Signed<T: utoipa::ToSchema> {
  |                ++++++++++++++++++
```

But the suggestion to define `struct Signed<T: utoipa::ToSchema>` would change the public API to require `ToSchema` for any `Signed<T>`. It would break other non-API-related code using `Signed<T>` on non-`ToSchema` types. My use case is to make `utoipa` dependency optional, so it's not possible to introduce breaking change on whether user enables `utoipa` feature.

I think a better solution would be generating the `ToSchema` bound on `impl` block, instead of requiring them on the struct definition. That is:
```rust
#[derive(ToSchema)]
struct Signed<T> { .. }
/// generates:
impl<T: ToSchema /*<- added by derive-macro*/> ToSchema for Signed<T> { .. }
```

oxalica:
Fixes #1078

```rust
#[derive(ToSchema)]
struct Signed<T> { .. }
/// generates
impl<T> ToSchema for Signed<T> where T: ToSchema /* <- introduced by this PR */ { .. }
```

- Each generic types now automatically get a `ToSchema` bound in where clause of generated `impl` blocks. Thus the user does not need to explicitly add `ToSchema` bounds on struct definitions.

- Attribute `schema(bound = ..)` is introduced for struct or enum containers to opt-out the automatic `ToSchema` bounds addition, by manually provide where clause bounds for the generated impl blocks. The attribute is inspired by [`serde(bound = ..)`](https://serde.rs/container-attrs.html#bound).
  Eg. `schema(bound = "")` can be used to remove bounds in case of `PhantomData` types.

This PR is based on 61c98bb9b40842ce84f6a67a572f4f4afc989610 instead of `master` because `master` is already failing the test after the rc version bump (snapshot assertion failure in `openapi_schemas_resolve_schema_references` and friends). It does not conflict anyway.

### DIFF
--- a/utoipa-gen/CHANGELOG.md
+++ b/utoipa-gen/CHANGELOG.md
@@ -20,6 +20,7 @@
 * Add support for description and summary overriding (https://github.com/juhaku/utoipa/pull/948)
 * Add nest `OpenApi` support (https://github.com/juhaku/utoipa/pull/930)
 * Add support for additional tags via `tags` (https://github.com/juhaku/utoipa/pull/916)
+* Add `bound` attribute for customizing generic impl bounds. (https://github.com/juhaku/utoipa/pull/1079)
 
 ### Fixed
 
--- a/utoipa-gen/src/component/features.rs
+++ b/utoipa-gen/src/component/features.rs
@@ -101,6 +101,7 @@ pub enum Feature {
     ContentEncoding(attributes::ContentEncoding),
     ContentMediaType(attributes::ContentMediaType),
     Discriminator(attributes::Discriminator),
+    Bound(attributes::Bound),
     MultipleOf(validation::MultipleOf),
     Maximum(validation::Maximum),
     Minimum(validation::Minimum),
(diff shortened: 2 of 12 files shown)


oxalica:
@juhaku Seems you are introducing some merge conflicts :thinking:

juhaku (changes requested):
This looks pretty good: :tada: Thanks for the patch. 

I updated it a bit, mainly I fixed the tests and rebased it above the master.

However could you update the docs for the new functionality in `/utoipa-gen/src/lib.rs` for what is changed in `ToSchema` derive macro. Also add entry to the `utoipa-gen/CHANGELOG.md` to top most of suitable category of unreleased changes. Maybe under `Added` title.

juhaku (approved):
Great, :bow: :1st_place_medal:

juhaku:
> Updated docs and added CHANGELOG. Seems the doctest for utoipa-gen (`cargo test -p utoipa-gen --doc`) is failing on master. Just manually checked all failures are unrelated.

Right, yeah there might still be some of those, I'll fix them later on.

PR #1082, opened 2024-10-03
Keep indentation in doc comments and simplify

oxalica:
Leading spaces is significant for Markdown and we should not blindlessly trim them, which may break Markdown lists or code blocks.

Here we strip the minimal indentation of all non-empty doc comment lines, so we can both keep Markdown semantics and avoid unnecessary leading spaces immediately after the `///` token.

Eg.
```rust
use utoipa::{OpenApi, ToSchema};

/// Summary
///
/// List:
/// - foo
///   bar
/// - baz
#[derive(ToSchema)]
struct Typ {}
```

The `bar` line is indented and should continue the previous paragraph in the first unordered list element `foo`. But if stripping the leading whitespaces, `bar` will start a new paragraph, which is incorrect.


juhaku (commented):
This is super :100:  Same drill could you mark the changes to the CHANGELOG in utoipa-gen since it touches it's functionality.

juhaku (approved):

PR #1096, opened 2024-10-05
Add a default impl of ToSchema::name()

ValentinColin:
I suggest adding a default implementation of the ToSchema trait.

This is a naive implementation that takes the type name and removes the module path and generic elements.

### DIFF
--- a/utoipa/CHANGELOG.md
+++ b/utoipa/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+* Add default implementation of `ToSchema` trait (https://github.com/juhaku/utoipa/pull/1096)
 * Add support for `property_names` for object (https://github.com/juhaku/utoipa/pull/1084)
 * Add auto collect schemas for utoipa-axum (https://github.com/juhaku/utoipa/pull/1072)
 * Add global config for `utiopa` (https://github.com/juhaku/utoipa/pull/1048)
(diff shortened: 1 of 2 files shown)


DenuxPlays (commented):
I am Not Sure how the Tests are done in this Library but I think a few Tests to Validate the Output would be a good idea.

juhaku:
Yup that's a valid point.

juhaku (changes requested):
While this is definitely handy for people manually implementing `ToSchema` it would be a good idea to write couple of tests to validate the output.

juhaku:
> This is a naive implementation that takes the type name and removes the module path and generic elements.

This naive implementation is enough since the name is by default for a type calculated only from the name without generics. Generic args in the name will come from the inner generic types respectively upon schema creation. When users want more advanced name, they can easily override the default functionality and provide their own solution IMO.

ValentinColin:
Absolutely right for unit test!

I've added some documentation too.

DenuxPlays (approved):
LGTM

juhaku (approved):
:100: super

juhaku:
One thing though, could you add a line (title o the PR) to CHANGELOG.md of in `utoipa/CHANGELOG.md` Add an new entry to top of the `Added` section of Unreleased release.

juhaku:
Great, thansk :slightly_smiling_face:

PR #1129, opened 2024-10-14
Implement schema traits for indexset

h7kanna:
Implemented traits for IndexSet when indexmap feature is enabled.


juhaku (approved):
Nice, thanks :100:

PR #1136, opened 2024-10-15
Fixed dependency name in doc comment

Kanerix:
Doc comment showed wrong dependency name ("_" instead of "-")


juhaku (commented):
Thanks, I have missed that one :+1:

PR #1137, opened 2024-10-15
Add `no_recursion` feature for `ToSchema`

### LINKED ISSUE #1134: Collecting recursive schemas fails with a stack overflow in 5.x
We have an api that uses a setup that boils to this:

```
use utoipa::{OpenApi, ToSchema};

#[derive(ToSchema)]
pub struct Instance {
    kind: Kind,
}

#[derive(ToSchema)]
pub enum Kind {
    MultipleNested(Vec<Instance>),
}

#[derive(ToSchema)]
pub struct Error {
    instance: Instance,
}

#[derive(OpenApi)]
#[openapi(components(schemas(Error)))]
pub struct ApiDoc {}

fn main() {
    let api = ApiDoc::openapi();
    println!("{:?}", api.to_json().unwrap());
}
```

This fails like that:

```
thread 'main' has overflowed its stack
fatal runtime error: stack overflow
```

in 5.x while it worked in 4.x.

Here's a backtrace from `gdb`:

```
[...]
#3762 0x0000555555565f1e in utoipa_recursion::{impl#3}::schemas (schemas=0x7fffffffced0) at src/main.rs:8
#3763 0x00005555555657be in utoipa_recursion::{impl#1}::schemas (schemas=0x7fffffffced0) at src/main.rs:3
#3764 0x0000555555565f1e in utoipa_recursion::{impl#3}::schemas (schemas=0x7fffffffced0) at src/main.rs:8
#3765 0x00005555555657be in utoipa_recursion::{impl#1}::schemas (schemas=0x7fffffffced0) at src/main.rs:3
#3766 0x000055555556653e in utoipa_recursion::{impl#5}::schemas (schemas=0x7fffffffced0) at src/main.rs:13
#3767 0x00005555555668d4 in utoipa_recursion::{impl#6}::openapi () at src/main.rs:18
#3768 0x0000555555564f66 in utoipa_recursion::main () at src/main.rs:23
```

Was there any change in how (bi-)recursive schemas are handled?

juhaku:
This commit adds a new `no_recursion` feature for `ToSchema` which is used to break out from recurring into infinite loop when collecting schema references. In case of looping schema tree this is mandatory to break the recursion. Failing to do so will be runtime panic.

Fixes #1134

### DIFF
--- a/utoipa-gen/CHANGELOG.md
+++ b/utoipa-gen/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
+* Add `no_recursion` feature for `ToSchema` (https://github.com/juhaku/utoipa/pull/1137)
+
 ### Fixed
 
 * Chore explicit FromIterator for edition 2018 (https://github.com/juhaku/utoipa/pull/1131)
--- a/utoipa-gen/src/component.rs
+++ b/utoipa-gen/src/component.rs
@@ -731,6 +731,7 @@ pub struct SchemaReference {
     pub tokens: TokenStream,
     pub references: TokenStream,
     pub is_inline: bool,
+    pub no_recursion: bool,
 }
 
 impl SchemaReference {
@@ -1188,6 +1189,9 @@ impl ComponentSchema {
                     let rewritten_path = type_path.rewrite_path()?;
                     let nullable_item = nullable_one_of_item(nullable);
                     let mut object_schema_reference = SchemaReference::default();
+                    object_schema_reference.no_recursion = features
+                        .iter()
+                        .any(|feature| matches!(feature, Feature::NoRecursion(_)));
 
                     if let Some(children) = &type_tree.children {
                         let children_name = Self::compose_name(
@@ -1468,6 +1472,7 @@ impl ComponentSchema {
                     tokens: quote! { <#rewritten_path as utoipa::PartialSchema>::schema() },
                     references: quote !{ <#rewritten_path as utoipa::ToSchema>::schemas(schemas) },
                     is_inline: false,
+                    no_recursion: false,
                 }))
                 )
             } else {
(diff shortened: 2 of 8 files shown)


jrudolph:
Awesome, thanks for the quick fix.

juhaku:
Yup, that was intended. It is allowed for unnamed variant because unnamed variant will actually generate schema for the unnamed field itself. I did not test it with tuple as value. It might not work as expected though. Otherwise it is expected to be given for a field that is not allowed to recurring any further.

juhaku:
For named schema it would not have any effect, unless it is treated as no recursion limit for all the fields of the struct itself. Such addition could be considered if there is much need for it.

jrudolph:
Ok, I see. I think, it's indeed not strictly required because you can put it on the field. It's just a bit inconvenient (and even hard to debug when you encounter it the first time) and understand why there's a diff between the different places where you can put those different attributes (allowed on top-level struct but not in enum struct). So, it would be good to have for consistency reasons. But I undestand that it might not be worth the additional effort.

juhaku:
```rust
    #[derive(ToSchema)]
    pub struct Instance {
        #[schema(no_recursion)]
        kind: Kind,
    }

    #[derive(ToSchema)]
    pub enum Recursion {
        Named {
            #[schema(no_recursion)]
            foobar: Box<Recur>,
        },
        #[schema(no_recursion)]
        Unnamed(Box<Recur>),
        NoValue,
    }

    #[derive(ToSchema)]
    #[schema(no_recursion)]
    pub struct UnnamedError(Kind);
```
Yeah it is allowed on these places, for it to be allowed on the named field struct or on the named field enum variant would be a different behavior to what its now being enabled for. For consistency it probably would be good to add and as of now might be a bit disorienting. It shouldn't be too big of an effort though.

juhaku:
@jrudolph Added a PR for this to further enhance the syntax: #1144

jrudolph:
Very nice, thanks :)

juhaku:
I might as well release it all the way, since I am not sure when the next release will be. :slightly_smiling_face:

PR #1155, opened 2024-10-21
Fix regression due to a985d8c1340f80ab69b2b0e5de799df98d567732

### LINKED ISSUE #1152: examples/axum-utoipa-bindings: the swagger ui cannot be shown properly when loading `/swagger-ui`
before a985d8c1340f80ab69b2b0e5de799df98d567732, the swagger ui can be shown when loading `/swagger-ui`.
the request for `/swagger-ui` is redirected to `/swagger-ui/` and the assets of the swagger ui are loaded properly.

a985d8c1340f80ab69b2b0e5de799df98d567732 breaks this behavior and the request for `/swagger-ui` is not redirected and urls of the assets are wrong.
because the url of the swagger ui is `/swagger-ui`  and relative urls of the assets are resolved with `/`.

i think that it's better to revert a985d8c1340f80ab69b2b0e5de799df98d567732 on the `master` branch. then fix #1042 properly.

the simplest solution is `panic!()` when `SwaggerUi::new()` is called with `/`.
i think that this is acceptable for most people because mounting the swagger ui to `/` is rare.

another solution is adding the route for redirect only when the `path` does not end with `/`.

```diff
diff --git a/utoipa-swagger-ui/src/axum.rs b/utoipa-swagger-ui/src/axum.rs
index 22a3179..7b76c7a 100644
--- a/utoipa-swagger-ui/src/axum.rs
+++ b/utoipa-swagger-ui/src/axum.rs
@@ -45,10 +45,20 @@ where
         let handler = routing::get(serve_swagger_ui).layer(Extension(Arc::new(config)));
         let path: &str = swagger_ui.path.as_ref();

-        router
-            .route(path, handler.clone())
-            .route(&format!("{}/", path), handler.clone())
-            .route(&format!("{}/*rest", path), handler)
+        if path.ends_with('/') {
+            router
+                .route(path, handler.clone())
+                .route(&format!("{}*rest", path), handler)
+        } else {
+            let slash_path = format!("{}/", path);
+            router
+                .route(
+                    path,
+                    routing::get(|| async move { axum::response::Redirect::to(&slash_path) }),
+                )
+                .route(&format!("{}/", path), handler.clone())
+                .route(&format!("{}/*rest", path), handler)
+        }
     }
 }

```

in this case, a request for `/swagger-ui` won't be redirected if `SwaggerUi::new()` is called with `/swagger-ui/`.
but i think that this is what the code is intended to do.

there is another solution.
currently, assets are specified with relative urls.
so, using absolute urls for the assets in the html document of the swagger ui can solve this issue without the redirect.
but, this kind of change affects servers other than axum and you may have to pay more cost for testing than other solutions.

i prefer the second solution, but you can choose one of any solutions.

masnagam:
Closes #1152

### DIFF
--- a/utoipa-swagger-ui/Cargo.toml
+++ b/utoipa-swagger-ui/Cargo.toml
@@ -34,7 +34,9 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0" }
 
 [dev-dependencies]
+axum-test = "16.2.0"
 similar = "2.5"
+tokio = { version = "1", features = ["macros"] }
 utoipa-swagger-ui = { path = ".", features = ["actix-web", "axum", "rocket"] }
 
 [package.metadata.docs.rs]
--- a/utoipa-swagger-ui/src/axum.rs
+++ b/utoipa-swagger-ui/src/axum.rs
@@ -45,10 +45,27 @@ where
         let handler = routing::get(serve_swagger_ui).layer(Extension(Arc::new(config)));
         let path: &str = swagger_ui.path.as_ref();
 
-        router
-            .route(path, handler.clone())
-            .route(&format!("{}/", path), handler.clone())
-            .route(&format!("{}/*rest", path), handler)
+        if path == "/" {
+            router
+                .route(path, handler.clone())
+                .route(&format!("{}*rest", path), handler)
+        } else {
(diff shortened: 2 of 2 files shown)


juhaku (approved):
Great, :+1:

PR #1171, opened 2024-10-28
Merge paths if they already are added to the paths map

hdoordt:
The old version of the diff would overwrite old routes to the paths map even if they had a different method, so this change merges them instead


utoipa-axum/src/router.rs
```diff
@@ -253,8 +253,15 @@ where
             })
         };
 
-        // add current paths to the OpenApi
-        self.1.paths.paths.extend(paths.paths.clone());
+        // add or merge current paths to the OpenApi
+        for (path, item) in paths.paths.clone() {
```
juhaku, line 257: I think we can do this without cloning as well. The `clone` there previously has been just forgotten there most likely.

juhaku (changes requested):
Nice, thanks for fixing, :+1: Could you add a CHANGELOG entry top of the file for this change.
```markdown
## Unreleased

### Changed

* This is the PR title (link)
```

hdoordt:
@juhaku Thanks for having a look. I've updated the CHANGELOG and removed the `clone` invocation

utoipa-axum/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Changed
 
 * Use OpenApiRouter::default for empty OpenApi (https://github.com/juhaku/utoipa/pull/1133)
+* Merge paths if they already are added to the paths map in utoipa-axum (https://github.com/juhaku/utoipa/pull/1171)
```
juhaku, line 8: Good except this has not been released yet :slightly_smiling_face:

hdoordt: Lol yeah fair

juhaku (approved):
Great, :+1:

Issue #1182, opened 2024-11-03
Generic type parameters are always inlined


JMLX42:
## Example

```rust
#[derive(ToSchema)]
pub struct ResponsePayload<T: ToSchema> {
    data: Vec<T>,
}

#[derive(ToSchema)]
pub struct PrimaryData {
    id: String,
    value: u32,
}
```

## Expected definition

```json
{
  "components": {
    "schemas": {
      "PrimaryData": {
        "properties": {
          "id": {
            "type": "string"
          },
          "value": {
            "format": "int32",
            "minimum": 0,
            "type": "integer"
          }
        },
        "required": [
          "id",
          "value"
        ],
        "type": "object"
      },
      "ResponsePayload_PrimaryData": {
        "properties": {
          "data": {
            "items": {
              "$ref": "#/components/schemas/PrimaryData"
            },
            "type": "array"
          }
        },
        "required": [
          "data"
        ],
        "type": "object"
      }
    }
  },
  "info": {
    "description": "",
    "license": {
      "name": ""
    },
    "title": "utoipa-generic-response",
    "version": "0.1.0"
  },
  "openapi": "3.1.0",
  "paths": {}
}
```

## Actual definition

```json
{
  "components": {
    "schemas": {
      "PrimaryData": {
        "properties": {
          "id": {
            "type": "string"
          },
          "value": {
            "format": "int32",
            "minimum": 0,
            "type": "integer"
          }
        },
        "required": [
          "id",
          "value"
        ],
        "type": "object"
      },
      "ResponsePayload_PrimaryData": {
        "properties": {
          "data": {
            "items": {
              "properties": {
                "id": {
                  "type": "string"
                },
                "value": {
                  "format": "int32",
                  "minimum": 0,
                  "type": "integer"
                }
              },
              "required": [
                "id",
                "value"
              ],
              "type": "object"
            },
            "type": "array"
          }
        },
        "required": [
          "data"
        ],
        "type": "object"
      }
    }
  },
  "info": {
    "description": "",
    "license": {
      "name": ""
    },
    "title": "utoipa-generic-response",
    "version": "0.1.0"
  },
  "openapi": "3.1.0",
  "paths": {}
}
```

```diff
25,39c25
<               "properties": {
<                 "id": {
<                   "type": "string"
<                 },
<                 "value": {
<                   "format": "int32",
<                   "minimum": 0,
<                   "type": "integer"
<                 }
<               },
<               "required": [
<                 "id",
<                 "value"
[...]

PR #1184, opened 2024-11-04
Do not inline non-primitive type parameters by default

JMLX42:
Fixes https://github.com/juhaku/utoipa/issues/1182

- [x] Implement a more comprehensive inlining strategy for generic type parameters
- [x] Fix the tests
- [x] Update the CHANGELOG

### DIFF
--- a/utoipa-gen/src/component.rs
+++ b/utoipa-gen/src/component.rs
@@ -1216,6 +1216,17 @@ impl ComponentSchema {
                     let title_tokens = as_tokens_or_diagnostics!(&title);
 
                     if is_inline {
+                        let schema_type = SchemaType {
+                            path: Cow::Borrowed(&rewritten_path),
+                            nullable,
+                        };
+                        let index =
+                            if !schema_type.is_primitive() || type_tree.generic_type.is_none() {
+                                container.generics.get_generic_type_param_index(type_tree)
+                            } else {
+                                None
+                            };
+
                         object_schema_reference.is_inline = true;
                         let items_tokens = if let Some(children) = &type_tree.children {
                             schema_references.extend(Self::compose_child_references(children)?);
@@ -1223,8 +1234,21 @@ impl ComponentSchema {
                             let composed_generics =
                                 Self::compose_generics(children, container.generics)?
                                     .collect::<Array<_>>();
-                            quote_spanned! {type_path.span()=>
-                                <#rewritten_path as utoipa::__dev::ComposeSchema>::compose(#composed_generics.to_vec())
+
+                            if index.is_some() {
+                                quote_spanned! {type_path.span()=>
+                                    let _ = <#rewritten_path as utoipa::PartialSchema>::schema;
+
+                                    if let Some(composed) = generics.get_mut(#index) {
+                                        composed.clone()
+                                    } else {
+                                        <#rewritten_path as utoipa::PartialSchema>::schema()
+                                    }
+                                }
+                            } else {
+                                quote_spanned! {type_path.span()=>
+                                    <#rewritten_path as utoipa::__dev::ComposeSchema>::compose(#composed_generics.to_vec())
+                                }
                             }
                         } else {
                             quote_spanned! {type_path.span()=>
@@ -1255,8 +1279,18 @@ impl ComponentSchema {
 
                         schema.to_tokens(tokens);
                     } else {
-                        let index = container.generics.get_generic_type_param_index(type_tree);
-                        // only set schema references tokens for concrete non generic types
+                        let schema_type = SchemaType {
+                            path: Cow::Borrowed(&rewritten_path),
+                            nullable,
+                        };
+                        let index =
+                            if !schema_type.is_primitive() || type_tree.generic_type.is_none() {
+                                container.generics.get_generic_type_param_index(type_tree)
+                            } else {
+                                None
+                            };
+
+                        // forcibly inline primitive type parameters, otherwise use references
                         if index.is_none() {
                             let reference_tokens = if let Some(children) = &type_tree.children {
                                 let composed_generics = Self::compose_generics(
@@ -1281,7 +1315,7 @@ impl ComponentSchema {
                                         let _ = <#rewritten_path as utoipa::PartialSchema>::schema;
 
                                         if let Some(composed) = generics.get_mut(#index) {
-                                            std::mem::take(composed)
+                                            composed.clone()
                                         } else {
                                             #item_tokens.into()
                                         }
--- a/utoipa-gen/tests/openapi_derive.rs
+++ b/utoipa-gen/tests/openapi_derive.rs
@@ -554,6 +554,64 @@ fn derive_nest_openapi_with_tags() {
 }
 
 #[test]
+fn openapi_schemas_resolve_generic_enum_schema() {
+    #![allow(dead_code)]
+    use utoipa::ToSchema;
+
+    #[derive(ToSchema)]
+    enum Element<T> {
+        One(T),
+        Many(Vec<T>),
+    }
+
+    #[derive(OpenApi)]
+    #[openapi(components(schemas(Element<String>)))]
+    struct ApiDoc;
+
+    let doc = ApiDoc::openapi();
+
+    let value = serde_json::to_value(&doc).expect("OpenAPI is JSON serializable");
+    let schemas = value.pointer("/components/schemas").unwrap();
+    let json = serde_json::to_string_pretty(&schemas).expect("OpenAPI is json serializable");
+    println!("{json}");
+
+    assert_json_eq!(
+        schemas,
+        json!({
+            "Element_String": {
+                "oneOf": [
+                    {
+                    "properties": {
+                        "One": {
+                        "type": "string"
+                        }
+                    },
+                    "required": [
+                        "One"
+                    ],
+                    "type": "object"
+                    },
+                    {
+                    "properties": {
+                        "Many": {
+                        "items": {
+                            "type": "string"
+                        },
+                        "type": "array"
+                        }
+                    },
+                    "required": [
+                        "Many"
+                    ],
+                    "type": "object"
+                    }
+                ]
+            }
+        })
+    )
+}
+
+#[test]
 fn openapi_schemas_resolve_schema_references() {
     #![allow(dead_code)]
     use utoipa::ToSchema;
--- a/utoipa-gen/tests/schema_generics.rs
+++ b/utoipa-gen/tests/schema_generics.rs
@@ -1,7 +1,9 @@
 use std::borrow::Cow;
 use std::marker::PhantomData;
 
+use assert_json_diff::assert_json_eq;
 use serde::Serialize;
+use serde_json::json;
 use utoipa::openapi::{Info, RefOr, Schema};
 use utoipa::{schema, OpenApi, PartialSchema, ToSchema};
 
@@ -21,6 +23,132 @@ fn generic_schema_custom_bound() {
     fn assert_is_to_schema<T: ToSchema>() {}
 
     assert_is_to_schema::<Type<NoToSchema>>();
+}
+
+#[test]
+fn generic_request_body_schema() {
+    #![allow(unused)]
+
+    #[derive(ToSchema)]
+    #[schema(as = path::MyType<T>)]
+    struct Type<T> {
+        #[schema(inline)]
+        t: T,
+    }
+
+    #[derive(ToSchema)]
+    struct Person<T: Sized, P> {
+        field: T,
+        #[schema(inline)]
+        t: P,
+    }
+
+    #[utoipa::path(
+        get,
+        path = "/handler",
+        request_body = inline(Person<String, Type<i32>>),
+    )]
+    async fn handler() {}
+
+    #[derive(OpenApi)]
+    #[openapi(
+        components(
+            schemas(
+                Person::<String, Type<i32>>,
+            )
+        ),
+        paths(
+            handler
+        )
+    )]
+    struct ApiDoc;
+
+    let mut doc = ApiDoc::openapi();
+    doc.info = Info::new("title", "version");
+
+    let actual = serde_json::to_value(&doc).expect("operation is JSON serializable");
+    let json = serde_json::to_string_pretty(&actual).unwrap();
+
+    println!("{json}");
+
+    assert_json_eq!(
+        actual,
+        json!({
+            "openapi": "3.1.0",
+            "info": {
+                "title": "title",
+                "version": "version"
+            },
+            "paths": {
+                "/handler": {
+                    "get": {
+                        "tags": [],
+                        "operationId": "handler",
+                        "requestBody": {
+                            "content": {
+                                "application/json": {
+                                    "schema": {
+                                        "type": "object",
+                                        "required": [
+                                            "field",
+                                            "t"
+                                        ],
+                                        "properties": {
+                                            "field": {
+                                                "type": "string"
+                                            },
+                                            "t": {
+                                                "type": "object",
+                                                "required": [
+                                                    "t"
+                                                ],
+                                                "properties": {
+                                                    "t": {
+                                                        "type": "integer",
+                                                        "format": "int32"
+                                                    }
+                                                }
+                                            }
+                                        }
+                                    }
+                                }
+                            },
+                            "required": true
+                        },
+                        "responses": {}
+                    }
+                }
+            },
+            "components": {
+                "schemas": {
+                    "Person_String_path.MyType_i32": {
+                        "type": "object",
+                        "required": [
+                            "field",
+                            "t"
+                        ],
+                        "properties": {
+                            "field": {
+                                "type": "string"
+                            },
+                            "t": {
+                                "type": "object",
+                                "required": [
+                                    "t"
+                                ],
+                                "properties": {
+                                    "t": {
+                                        "type": "integer",
+                                        "format": "int32"
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        })
+    );
 }
 
 #[test]
@@ -107,6 +235,7 @@ fn schema_with_non_generic_root() {
 
     #[derive(ToSchema)]
     struct Bar<T> {
+        #[schema(inline)]
         value: T,
     }
 
@@ -231,6 +360,7 @@ fn high_order_types() {
 
     #[derive(ToSchema)]
     pub struct High<T> {
+        #[schema(inline)]
         high: T,
     }
 
--- a/utoipa-gen/tests/testdata/openapi_schemas_resolve_inner_schema_references
+++ b/utoipa-gen/tests/testdata/openapi_schemas_resolve_inner_schema_references
@@ -40,7 +40,7 @@
           "properties": {
             "Many": {
               "items": {
-                "type": "object"
+                "type": "string"
               },
               "type": "array"
             }
@@ -95,6 +95,32 @@
           "properties": {
             "Many": {
               "items": {
+                "properties": {
+                  "accounts": {
+                    "items": {
+                      "oneOf": [
+                        {
+                          "type": "null"
+                        },
+                        {
+                          "$ref": "#/components/schemas/Account"
+                        }
+                      ]
+                    },
+                    "type": "array"
+                  },
+                  "foo_bar": {
+                    "$ref": "#/components/schemas/Foobar"
+                  },
+                  "name": {
+                    "type": "string"
+                  }
+                },
+                "required": [
+                  "name",
+                  "foo_bar",
+                  "accounts"
+                ],
                 "type": "object"
               },
               "type": "array"
--- a/utoipa-gen/tests/testdata/schema_generics_openapi
+++ b/utoipa-gen/tests/testdata/schema_generics_openapi
@@ -193,7 +193,7 @@
               "Many": {
                 "type": "array",
                 "items": {
-                  "type": "object"
+                  "type": "string"
                 }
               }
             }


JMLX42:
@juhaku I attempted to fix the `openapi_schemas_resolve_schema_references` to add `#[schema(inline)]` where the reference expects it. But I think the reference is wrong.

Here is the difference:

```diff
43c43
<                 "type": "object"
---
>                 "type": "string"
97a98,123
>                 "properties": {
>                   "accounts": {
>                     "items": {
>                       "oneOf": [
>                         {
>                           "type": "null"
>                         },
>                         {
>                           "$ref": "#/components/schemas/Account"
>                         }
>                       ]
>                     },
>                     "type": "array"
>                   },
>                   "foo_bar": {
>                     "$ref": "#/components/schemas/Foobar"
>                   },
>                   "name": {
>                     "type": "string"
>                   }
>                 },
>                 "required": [
>                   "name",
>                   "foo_bar",
>                   "accounts"
>                 ],
```

I think the reference is wrong:

- `Element_String.oneOf[1].properties.Many.items.type` should be `string` (just like the `One` variant), but it is `object` in the reference.
- `Element_Yeah.oneOf[1].properties.Many.items.type` should be the inlined schema of `Yeah`, but it is `{ "type": "object" }` in the reference.

Should I fix the reference?

juhaku:
True it is indeed wrong, I definitely would appreciate if you want to take the initiative to enter to the rabbit hole of generics and fix the broken references as well. :laughing: For now I have no idea as of yet that what is causing that misalignment.

juhaku:
But maybe we can just leave it as is. Like without the `inline` it will now create correct reference. And with `inline` it does inline the actual type. So I think this is probably something that we want and we can just settle the fact that for now on the generics are not inlined by default? :thinking:

JMLX42:
> True it is indeed wrong, I definitely would appreciate if you want to take the initiative to enter to the rabbit hole of generics and fix the broken references as well. 😆 For now I have no idea as of yet that what is causing that misalignment.

@juhaku that's the neat part : my PR fixes it AFAIK. But I just wanted your approval before changing the reference JSON.

> But maybe we can just leave it as is. Like without the inline it will now create correct reference. And with inline it does inline the actual type. So I think this is probably something that we want and we can just settle the fact that for now on the generics are not inlined by default? 

My point exactly. The result with those changes is IMO better and fixes that broken test/reference. So if I just change the reference JSON:

1. the expected result is correct
2. the way inlining + generics works is IMHO more sound
3. **but** generics are now using refs instead of inlining

But you can't have 2. without 3.

So if that's ok with you I will:

- Fix the reference JSON for the broken test.
- Update the CHANGELOG

juhaku:
Yes, that is ok for me, thanks for the work :slightly_smiling_face: :+1:

JMLX42:
If I check only with `schema.is_primitive()`, I get the following failed test:

```
test derive_generic_schema_enum_variants ... FAILED

failures:

---- derive_generic_schema_enum_variants stdout ----
{
  "openapi": "3.1.0",
  "info": {
    "title": "title",
    "version": "version"
  },
  "paths": {},
  "components": {
    "schemas": {
      "BTreeMap": {
        "type": "object",
        "additionalProperties": {
          "type": "string"
        },
        "propertyNames": {
          "type": "string"
        }
      },
      "BTreeSet": {
        "type": "array",
        "items": {
          "type": "integer",
          "format": "int32"
        },
        "uniqueItems": true
      },
      "FooStruct_Option_i32": {
        "type": "object",
        "required": [
          "foo"
        ],
        "properties": {
          "foo": {
            "$ref": "#/components/schemas/Option"
          }
        }
      },
      "FoosEnum": {
        "oneOf": [
          {
            "type": "object",
            "required": [
              "ThingNoAliasOption"
            ],
            "properties": {
              "ThingNoAliasOption": {
                "$ref": "#/components/schemas/FooStruct_Option_i32"
              }
            }
          },
          {
            "type": "object",
            "required": [
              "FooEnumThing"
            ],
            "properties": {
              "FooEnumThing": {
                "type": "object",
                "required": [
                  "foo"
                ],
                "properties": {
                  "foo": {
                    "$ref": "#/components/schemas/Vec"
                  }
                }
              }
            }
          },
          {
            "type": "object",
            "required": [
              "FooThingOptionVec"
            ],
            "properties": {
              "FooThingOptionVec": {
                "type": "object",
                "required": [
                  "foo"
                ],
                "properties": {
                  "foo": {
                    "$ref": "#/components/schemas/Option"
                  }
                }
              }
            }
          },
          {
            "type": "object",
            "required": [
              "FooThingLinkedList"
            ],
            "properties": {
              "FooThingLinkedList": {
                "type": "object",
                "required": [
                  "foo"
                ],
                "properties": {
                  "foo": {
                    "$ref": "#/components/schemas/LinkedList"
                  }
                }
              }
            }
          },
          {
            "type": "object",
            "required": [
              "FooThingBTreeMap"
            ],
            "properties": {
              "FooThingBTreeMap": {
                "type": "object",
                "required": [
                  "foo"
                ],
                "properties": {
                  "foo": {
                    "$ref": "#/components/schemas/BTreeMap"
                  }
                }
              }
            }
          },
          {
            "type": "object",
            "required": [
              "FooThingHashMap"
            ],
            "properties": {
              "FooThingHashMap": {
                "type": "object",
                "required": [
                  "foo"
                ],
                "properties": {
                  "foo": {
                    "$ref": "#/components/schemas/HashMap"
                  }
                }
              }
            }
          },
          {
            "type": "object",
            "required": [
              "FooThingHashSet"
            ],
            "properties": {
              "FooThingHashSet": {
                "type": "object",
                "required": [
                  "foo"
                ],
                "properties": {
                  "foo": {
                    "$ref": "#/components/schemas/HashSet"
                  }
                }
              }
            }
          },
          {
            "type": "object",
            "required": [
              "FooThingBTreeSet"
            ],
            "properties": {
              "FooThingBTreeSet": {
                "type": "object",
                "required": [
                  "foo"
                ],
                "properties": {
                  "foo": {
                    "$ref": "#/components/schemas/BTreeSet"
                  }
                }
              }
            }
          }
        ]
      },
      "HashMap": {
        "type": "object",
        "additionalProperties": {
          "type": "string"
        },
        "propertyNames": {
          "type": "integer",
          "format": "int32"
        }
      },
      "HashSet": {
        "type": "array",
        "items": {
          "type": "integer",
          "format": "int32"
        },
        "uniqueItems": true
      },
      "LinkedList": {
        "type": "array",
        "items": {
          "type": "integer",
          "format": "int32"
        }
      },
      "Option": {
        "oneOf": [
          {
            "type": "null"
          },
          {
            "type": "array",
            "items": {
              "type": "integer",
              "format": "int32"
            }
          }
        ]
      },
      "Vec": {
        "type": "array",
        "items": {
          "type": "integer",
          "format": "int32"
        }
      }
    }
  }
}
```

Of course, we do not want refs to known/native generics. So I will add a test for `type_tree.generic_type.is_none()`.

juhaku:
Sure, it also seems that the it still creates a reference to `String` for some reason. 
```diff
diff --git a/utoipa-gen/tests/testdata/openapi_schemas_resolve_inner_schema_references b/utoipa-gen/tests/testdata/openapi_schemas_resolve_inner_schema_references
index 7368bc5..718e792 100644
--- a/utoipa-gen/tests/testdata/openapi_schemas_resolve_inner_schema_references
+++ b/utoipa-gen/tests/testdata/openapi_schemas_resolve_inner_schema_references
@@ -28,7 +28,7 @@
         {
           "properties": {
             "One": {
-              "type": "string"
+              "$ref": "#/components/schemas/String"
             }
           },
           "required": [
@@ -40,7 +40,7 @@
           "properties": {
             "Many": {
               "items": {
-                "type": "object"
+                "$ref": "#/components/schemas/String"
               },
               "type": "array"
             }
@@ -57,33 +57,7 @@
         {
           "properties": {
             "One": {
-              "properties": {
-                "accounts": {
-                  "items": {
-                    "oneOf": [
-                      {
-                        "type": "null"
-                      },
-                      {
-                        "$ref": "#/components/schemas/Account"
-                      }
-                    ]
-                  },
-                  "type": "array"
-                },
-                "foo_bar": {
-                  "$ref": "#/components/schemas/Foobar"
-                },
-                "name": {
-                  "type": "string"
-                }
-              },
-              "required": [
-                "name",
-                "foo_bar",
-                "accounts"
-              ],
-              "type": "object"
+              "$ref": "#/components/schemas/Yeah"
             }
           },
           "required": [
@@ -95,7 +69,7 @@
           "properties": {
             "Many": {
               "items": {
-                "type": "object"
+                "$ref": "#/components/schemas/Yeah"
               },
               "type": "array"
             }
@@ -231,6 +205,9 @@
       ],
       "type": "object"
     },
+    "String": {
+      "type": "string"
+    },
     "ThisIsNone": {
       "default": null
     },

```

JMLX42:
> Sure, it also seems that the it still creates a reference to `String` for some reason.

@juhaku weird, that change should fix it:

https://github.com/juhaku/utoipa/pull/1184/files#diff-e437cad9daa59d1d04f3511c45a3ff26a0626b664485d236fca0944598b77a24R563-R564

juhaku:
> https://github.com/juhaku/utoipa/pull/1184/files#diff-e437cad9daa59d1d04f3511c45a3ff26a0626b664485d236fca0944598b77a24R563-R564

Yup that fixes, it but if you remove the `inline` from the arguments, the reference is created correctly for the `Yeah`, but not for the `String`. It has something to do with the way the schema references is created :thinking:

JMLX42:
> Yup that fixes, it but if you remove the `inline` from the arguments, the reference is created correctly for the `Yeah`, but not for the `String`. It has something to do with the way the schema references is created 🤔

Correct. But if you add the `type_tree.generic_type.is_none()` it covers this case too. I just updated my branch:

https://github.com/juhaku/utoipa/pull/1184/files#diff-4eda00dd4f48f35e49679edd4d11d86a327ce484703e1352743c069c083e01e7R1263

JMLX42:
The only failing test AFAIK is `openapi_schemas_resolve_schema_references` .

```diff
43c43
<                 "type": "string"
---
>                 "type": "object"
98,123d97
<                 "properties": {
<                   "accounts": {
<                     "items": {
<                       "oneOf": [
<                         {
<                           "type": "null"
<                         },
<                         {
<                           "$ref": "#/components/schemas/Account"
<                         }
<                       ]
<                     },
<                     "type": "array"
<                   },
<                   "foo_bar": {
<                     "$ref": "#/components/schemas/Foobar"
<                   },
<                   "name": {
<                     "type": "string"
<                   }
<                 },
<                 "required": [
<                   "name",
<                   "foo_bar",
<                   "accounts"
<                 ],
```

And I'm a bit puzzled by this one...

JMLX42:
> And I'm a bit puzzled by this one...

I have created a separate test: [`openapi_schemas_resolve_generic_enum_schema`](https://github.com/juhaku/utoipa/pull/1184/commits/dce8e512cd773635b9940b5e0bbdb5815b718fc0#diff-e437cad9daa59d1d04f3511c45a3ff26a0626b664485d236fca0944598b77a24R556-R613)

juhaku:
True, that is a bit weird, I am trying to debug it as well.

JMLX42:
Could it be specific to how enum schemas are handled?

JMLX42:
FYI here is the expanded code:

```rust
mod response {
    use std::ops::Deref;
    use utoipa::{openapi::schema::ArrayItems, ToSchema};
    pub enum Element<T> {
        One(T),
        Many(Vec<T>),
    }
    impl<T> utoipa::__dev::ComposeSchema for Element<T>
    where
        T: utoipa::ToSchema,
    {
        fn compose(
            mut generics: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
            Into::<
                utoipa::openapi::schema::OneOfBuilder,
            >::into(utoipa::openapi::OneOf::with_capacity(2usize))
                .item(
                    utoipa::openapi::schema::Object::builder()
                        .property(
                            "One",
                            {
                                let _ = <T as utoipa::PartialSchema>::schema;
                                if let Some(composed) = generics.get_mut(0usize) {
                                    std::mem::take(composed)
                                } else {
                                    utoipa::openapi::schema::RefBuilder::new()
                                        .ref_location_from_schema_name(
                                            ::alloc::__export::must_use({
                                                let res = ::alloc::fmt::format(
                                                    format_args!("{0}", <T as utoipa::ToSchema>::name()),
                                                );
                                                res
                                            }),
                                        )
                                        .into()
                                }
                            },
                        )
                        .required("One"),
                )
                .item(
                    utoipa::openapi::schema::Object::builder()
                        .property(
                            "Many",
                            utoipa::openapi::schema::ArrayBuilder::new()
                                .items({
                                    let _ = <T as utoipa::PartialSchema>::schema;
                                    if let Some(composed) = generics.get_mut(0usize) {
                                        std::mem::take(composed)
                                    } else {
                                        utoipa::openapi::schema::RefBuilder::new()
                                            .ref_location_from_schema_name(
                                                ::alloc::__export::must_use({
                                                    let res = ::alloc::fmt::format(
                                                        format_args!("{0}", <T as utoipa::ToSchema>::name()),
                                                    );
                                                    res
                                                }),
                                            )
                                            .into()
                                    }
                                }),
                        )
                        .required("Many"),
                )
                .into()
        }
    }
    impl<T> utoipa::ToSchema for Element<T>
    where
        T: utoipa::ToSchema,
    {
        fn name() -> std::borrow::Cow<'static, str> {
            std::borrow::Cow::Borrowed("Element")
        }
        fn schemas(
            schemas: &mut Vec<
                (String, utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>),
            >,
        ) {
            schemas.extend([]);
            <T as utoipa::ToSchema>::schemas(schemas);
            <T as utoipa::ToSchema>::schemas(schemas);
        }
    }
}
```

juhaku:
> Could it be specific to how enum schemas are handled?

Could be, but hard to tell right away.

JMLX42:
```rust
        fn schemas(
            schemas: &mut Vec<
                (String, utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>),
            >,
        ) {
            schemas.extend([]);
            <T as utoipa::ToSchema>::schemas(schemas);
            <T as utoipa::ToSchema>::schemas(schemas);
        }
```

Not sure if it's related, but the `<T as utoipa::ToSchema>::schemas(schemas)` x2 looks like an issue.

Otherwise the code generated for the `One` and `Many` alternative are pretty much exactly the same...

juhaku:
> ```rust
> ```rust
> 
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> Not sure if it's related, but the <T as utoipa::ToSchema>::schemas(schemas) x2 looks like an issue.
> ```

Yeah, it should not be an issue, but could also be fixed though, it in theory should result same spec but just redundantly calls the compose twice. There is no check whether the same reference call is already placed so it blindly adds schema reference call for each variant of the enum. Perhaps this could be fixed to the enum schema references implementation.

JMLX42:
Alignment aside, the generated code is the same for the `One` and the `Many.items` variants:

image

Could it be that `ArrayBuilder::items()` behaves unexpectedly?

juhaku:
> Could it be that `ArrayBuilder::items()` behaves unexpectedly?

Could be, as what `dbg!(<Yeah as utoipa::PartialSchema>::schema());` returns, it returns correct schema, but when it is placed within the compose:
```rust
    let v = <Element<Yeah> as utoipa::__dev::ComposeSchema>::compose(
        [<Yeah as utoipa::PartialSchema>::schema()].to_vec(),
    );
```
This will result the invalid code for somehow.

juhaku:
It might have something to do with the `utiopa/src/schema.rs` `ArrayItems` default implementation, as It creates an empty object. I this might result the empty schema to be used instead :thinking: 
```rust
impl Default for ArrayItems {
    fn default() -> Self {
        Self::RefOrSchema(Box::new(Object::with_type(SchemaType::AnyValue).into()))
    }
}
```

JMLX42:
> > Could it be that `ArrayBuilder::items()` behaves unexpectedly?
> 
> Could be, as what `dbg!(<Yeah as utoipa::PartialSchema>::schema());` returns, it returns correct schema, but when it is placed within the compose:
> 
> ```rust
>     let v = <Element<Yeah> as utoipa::__dev::ComposeSchema>::compose(
>         [<Yeah as utoipa::PartialSchema>::schema()].to_vec(),
>     );
>     dbg!(&v);
> ```
> 
> This will result the invalid code for somehow.

IMO that `compose()` leads to [`impl<T: ComposeSchema> ComposeSchema for Vec<T>`](https://github.com/juhaku/utoipa/blob/c45215c303ab288ed132a487c4b86b00b9513573/utoipa/src/lib.rs#L1373). And for `T = String`, it leads to the code generated by `impl_compose_schema!(String)` which generates:

```rust
impl ComposeSchema for String {
                fn compose(_: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
                    schema!( String ).into()
                }
            }
```

But I cannot find the `schema!` macro...

JMLX42:
> IMO that `compose()` leads to [`impl<T: ComposeSchema> ComposeSchema for Vec<T>`](https://github.com/juhaku/utoipa/blob/c45215c303ab288ed132a487c4b86b00b9513573/utoipa/src/lib.rs#L1373). And for `T = String`, it leads to the code generated by `impl_compose_schema!(String)` which generates:

My bad. In the broken case, `T = Vec<String>`.

JMLX42:
> It might have something to do with the `utoipa/src/schema.rs` `ArrayItems` default implementation, as It creates an empty object. I this might result the empty schema to be used instead 🤔
> 
> ```rust
> impl Default for ArrayItems {
>     fn default() -> Self {
>         Self::RefOrSchema(Box::new(Object::with_type(SchemaType::AnyValue).into()))
>     }
> }
> ```

I've added a `dbg!("ArrayItems::default");` in `ArrayItems::default()` and it *is* called when running the faulty test.

JMLX42:
> I've added a `dbg!("ArrayItems::default");` in `ArrayItems::default()` and it _is_ called when running the faulty test.

But changing its impl does not change the outcome of the test:

```rust
impl Default for ArrayItems {
    fn default() -> Self {
        dbg!("ArrayItems::default");
        Self::False
    }
}
```

So IMO the problem is elsewhere. No?

juhaku:
Now I know what happens :facepalm: the problem is `mem::take` it does not work for enums which are reusing same type for multiple variants. What ends up happening is that `std::mem::take` will remove the `<Yeah as PartialSchema>::schema()` for the first variant called `One` and replace it with empty object schema and the empty object schema is returned for the `Many` variant. :smile:

JMLX42:
> Now I know what happens 🤦 the problem is `mem::take` it does not work for enums which are reusing same type for multiple variants. What ends up happening is that `std::mem::take` will remove the `<Yeah as PartialSchema>::schema()` for the first variant called `One` and replace it with empty object schema and the empty object schema is returned for the `Many` variant. 😄

That makes perfect sense. Why taking away the generic in the first place?

Shouldn't that trigger the same problem for a struct with multiple field typed with the same type parameter?

juhaku:
> That makes perfect sense. Why taking away the generic in the first place?

Idea was just to reuse memory, and avoid cloning, But true that would happen for structs using same type over and over again.

It seems the idea was more beautiful than the implementation.

juhaku:
Yeah, nice one :+1: It was some hard core debugging

JMLX42:
> Probably should consider migrating to vec of `Rc<RefOr<Schema>>` for reusing memory? 🤔

Probably. But out of scope for this PR if you don't mind :rofl: 

The last failing test is `derive_generic_schema_collect_recursive_schema_not_inlined`, and IMO the ref is wrong.

```diff
77,89c77
<                   "type": "object",
<                   "required": [
<                     "t"
<                   ],
<                   "properties": {
<                     "t": {
<                       "type": "string",
<                       "enum": [
<                         "One",
<                         "Two"
<                       ]
<                     }
<                   }
---
>                   "type": "string"
100,112c88,92
<                   "type": "object",
<                   "required": [
<                     "t"
<                   ],
<                   "properties": {
<                     "t": {
<                       "type": "string",
<                       "enum": [
<                         "One",
<                         "Two"
<                       ]
<                     }
<                   }
---
>                   "type": "string",
>                   "enum": [
>                     "One",
>                     "Two"
>                   ]
```
 `FooStruct_BTreeMap_Ty_Ky_Person_Value.properties.foo.additionalProperties.properties.t`  matches with `Person<Value>` and `Value` is a newtype for `String`. So the proper type is `"string"`, but the reference says:

```json
              "t": {
                  "type": "object",
                  "required": [
                    "t"
                  ],
                  "properties": {
                    "t": {
                      "type": "string",
                      "enum": [
                        "One",
                        "Two"
                      ]
                    }
                  }
                }
````

`FooStruct_BTreeMap_Ty_Ky_Person_Value.properties.foo.propertyNames` matches with `Ty<Ky>`, and `Ky` is the `One | Two` enum. So there is no reason to have a nested `FooStruct_BTreeMap_Ty_Ky_Person_Value.properties.foo.propertyNames.t.properties.t`.

I fixed the reference in 58405c8ea14e95e622e429696440e1eabcd4689b

@juhaku I'll you double check my changes made to the reference JSON files.

**Comment by juhaku** [comment]
> @juhaku I'll you double check my changes made to the reference JSON files.

Sure I check them out.

> The last failing test is `derive_generic_schema_collect_recursive_schema_not_inlined`, and IMO the ref is wrong.

That is interesting, with my changes there is no more failing tests :thinking:  Though I did not add any `inline` args.

**Comment by juhaku** [comment]
```diff
diff --git a/utoipa-gen/src/component.rs b/utoipa-gen/src/component.rs
index 5165196..70fb0a5 100644
--- a/utoipa-gen/src/component.rs
+++ b/utoipa-gen/src/component.rs
@@ -1259,11 +1259,12 @@ impl ComponentSchema {
                             path: Cow::Borrowed(&rewritten_path),
                             nullable,
                         };
-                        let index = if schema_type.is_primitive() {
-                            container.generics.get_generic_type_param_index(type_tree)
-                        } else {
-                            None
-                        };
+                        let index =
+                            if schema_type.is_primitive() || type_tree.generic_type.is_none() {
+                                container.generics.get_generic_type_param_index(type_tree)
+                            } else {
+                                None
+                            };
 
                         // forcibly inline primitive type parameters, otherwise use references
                         if index.is_none() {
@@ -1290,7 +1291,7 @@ impl ComponentSchema {
                                         let _ = <#rewritten_path as utoipa::PartialSchema>::schema;
 
                                         if let Some(composed) = generics.get_mut(#index) {
-                                            std::mem::take(composed)
+                                            composed.clone()
                                         } else {
                                             #item_tokens.into()
                                         }
diff --git a/utoipa-gen/tests/openapi_derive.rs b/utoipa-gen/tests/openapi_derive.rs
index bda75b9..d34d569 100644
--- a/utoipa-gen/tests/openapi_derive.rs
+++ b/utoipa-gen/tests/openapi_derive.rs
@@ -4,7 +4,7 @@ use assert_json_diff::{assert_json_eq, assert_json_include};
 use serde::Serialize;
 use serde_json::{json, Value};
 use utoipa::{
-    openapi::{schema, RefOr, Response, ResponseBuilder},
+    openapi::{schema, RefOr, Response, ResponseBuilder, Schema},
     OpenApi, ToResponse,
 };
 use utoipa_gen::ToSchema;
@@ -208,7 +208,7 @@ fn derive_openapi_with_servers() {
     #[openapi(
         servers(
             (url = "http://localhost:8989", description = "this is description"),
-            (url = "http://api.{username}:{port}", description = "remote api", 
+            (url = "http://api.{username}:{port}", description = "remote api",
                 variables(
                     ("username" = (default = "demo", description = "Default username for API")),
                     ("port" = (default = "8080", enum_values("8080", "5000", "3030"), description = "Supported ports for the API"))
@@ -560,8 +560,8 @@ fn openapi_schemas_resolve_schema_references() {
 
     #[derive(ToSchema)]
     enum Element<T> {
-        One(#[schema(inline)] T),
-        Many(#[schema(inline)] Vec<T>),
+        One(T),
+        Many(Vec<T>),
     }
 
     #[derive(ToSchema)]
diff --git a/utoipa-gen/tests/testdata/openapi_schemas_resolve_inner_schema_references b/utoipa-gen/tests/testdata/openapi_schemas_resolve_inner_schema_references
index 7368bc5..9e0cd74 100644
--- a/utoipa-gen/tests/testdata/openapi_schemas_resolve_inner_schema_references
+++ b/utoipa-gen/tests/testdata/openapi_schemas_resolve_inner_schema_references
@@ -40,7 +40,7 @@
           "properties": {
             "Many": {
               "items": {
-                "type": "object"
+                "type": "string"
               },
               "type": "array"
             }
@@ -95,6 +95,32 @@
           "properties": {
             "Many": {
               "items": {
+                "properties": {
+                  "accounts": {
+                    "items": {
+                      "oneOf": [
+                        {
+                          "type": "null"
+                        },
+                        {
+                          "$ref": "#/components/schemas/Account"
+                        }
+                      ]
+                    },
+                    "type": "array"
+                  },
+                  "foo_bar": {
+                    "$ref": "#/components/schemas/Foobar"
+                  },
+                  "name": {
+                    "type": "string"
+                  }
+                },
+                "required": [
+                  "name",
+                  "foo_bar",
+                  "accounts"
+                ],
                 "type": "object"
               },
               "type": "array"
diff --git a/utoipa-gen/tests/testdata/schema_generics_openapi b/utoipa-gen/tests/testdata/schema_generics_openapi
index 95c429c..f69eda2 100644
--- a/utoipa-gen/tests/testdata/schema_generics_openapi
+++ b/utoipa-gen/tests/testdata/schema_generics_openapi
@@ -193,7 +193,7 @@
               "Many": {
                 "type": "array",
                 "items": {
-                  "type": "object"
+                  "type": "string"
                 }
               }
 ```

**Comment by juhaku** [comment]
With this, if i run `./scripts/test.sh` it will pass the checks.

File: utoipa-gen/tests/testdata/schema_generic_collect_non_inlined_schema [hunk 1]
```diff
@@ -140,7 +140,7 @@
                 },
                 "t": {
                   "type": "integer",
-                  "format": "int32"
+                  "format": "int64"
```
juhaku, line 143: These JSON changes seems correct except how this `int32` changes to `int64`

juhaku:
True, that `additionalProperties` `t` value seems incorrect.

utoipa-gen/tests/testdata/schema_generic_collect_non_inlined_schema
JMLX42: `FooStruct_HashMap_i32_Person_i64` is the schema for `HashMap(FooStruct<std::collections::HashMap<i32, Person<i64>>>)`. So `Person<i64>::t` is typed as `i64`. No?

JMLX42:
> With this, if i run `./scripts/test.sh` it will pass the checks.

I was testing with `cargo test -p utoipa-gen`. Let me run `./scripts/test.sh` and remove the superfluous changes.

JMLX42:
> > With this, if i run `./scripts/test.sh` it will pass the checks.
> 
> I was testing with `cargo test -p utoipa-gen`. Let me run `./scripts/test.sh` and remove the superfluous changes.

@juhaku `./scripts/test.sh utoipa-gen` fails on `generic_schema_full_api` with:

```rust
    #[derive(ToSchema)]
    struct Person<'p, T: Sized, P> {
        id: usize,
        name: Option<Cow<'p, str>>,
        field: T,
        t: P,
    }
```

but passes with:

```rust
    #[derive(ToSchema)]
    struct Person<'p, T: Sized, P> {
        id: usize,
        name: Option<Cow<'p, str>>,
        #[schema(inline)]
        field: T,
        #[schema(inline)]
        t: P,
    }
```

juhaku:
> @juhaku `./scripts/test.sh utoipa-gen` fails on `generic_schema_full_api` with:

Okay, I'll check.

JMLX42:
Without the inline on `Person::field`:

```rust
#[derive(ToSchema)]
    struct Person<'p, T: Sized, P> {
        id: usize,
        name: Option<Cow<'p, str>>,
        field: T,
        #[schema(inline)]
        t: P,
    }
```

the test fails and there is an unexpected `$ref` to `#/components/schemas/String`

JMLX42:
> > @juhaku `./scripts/test.sh utoipa-gen` fails on `generic_schema_full_api` with:
> 
> Okay, I'll check.

Ok you *are* right. Removing all the `#[schema(inline)]` in `generic_schema_full_api` make the test pass.

But adding some of them leads to unexpected behaviors such as [this](https://github.com/juhaku/utoipa/pull/1184#issuecomment-2455533360). or 

It looks like the following change helps:

```diff
diff --git a/utoipa-gen/src/component.rs b/utoipa-gen/src/component.rs
index 70fb0a5..f51ea75 100644
--- a/utoipa-gen/src/component.rs
+++ b/utoipa-gen/src/component.rs
@@ -1259,12 +1259,13 @@ impl ComponentSchema {
                             path: Cow::Borrowed(&rewritten_path),
                             nullable,
                         };
-                        let index =
-                            if schema_type.is_primitive() || type_tree.generic_type.is_none() {
-                                container.generics.get_generic_type_param_index(type_tree)
-                            } else {
-                                None
-                            };
+                        object_schema_reference.is_inline =
+                            !schema_type.is_primitive() && type_tree.generic_type.is_some();
+                        let index = if !object_schema_reference.is_inline {
+                            container.generics.get_generic_type_param_index(type_tree)
+                        } else {
+                            None
+                        };
```

juhaku:
> It looks like the following change helps:

Oh, nice, great.

juhaku:
@JMLX42 Getting back to the `derive_generic_schema_collect_recursive_schema_not_inlined` test that was failing with invalid schema for the `BTreeMap`. The following changes fixes this for the `BTreeMap` but this also indicates that the default implementation needs to be also changed for other type than the `BTreeMap` as well.
```diff
diff --git a/utoipa/src/lib.rs b/utoipa/src/lib.rs
index 7c9d666..57b16c1 100644
--- a/utoipa/src/lib.rs
+++ b/utoipa/src/lib.rs
@@ -1254,6 +1254,7 @@ impl_from_for_number!(
 pub mod __dev {
     use utoipa_gen::schema;
 
+    use crate::openapi::RefOr;
     use crate::{utoipa, OpenApi, PartialSchema};
 
     pub trait PathConfig {
@@ -1431,13 +1432,32 @@ pub mod __dev {
         }
     }
 
-    impl<K: ComposeSchema, T: ComposeSchema> ComposeSchema for std::collections::BTreeMap<K, T> {
+    impl<K: ComposeSchema + utoipa::ToSchema, T: ComposeSchema + utoipa::ToSchema> ComposeSchema
+        for std::collections::BTreeMap<K, T>
+    {
         fn compose(
             schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
         ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
+            fn schema_or_ref<V: ComposeSchema + utoipa::ToSchema>(
+                schemas: &[utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>],
+                index: usize,
+            ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
+                let _ = <V as utoipa::PartialSchema>::schema;
+                if let Some(composed) = schemas.get(index) {
+                    composed.clone()
+                } else {
+                    utoipa::openapi::schema::RefBuilder::new()
+                        .ref_location_from_schema_name(format!(
+                            "{}",
+                            <V as utoipa::ToSchema>::name()
+                        ))
+                        .into()
+                }
+            }
+
             utoipa::openapi::ObjectBuilder::new()
-                .property_names(Some(K::compose(schemas.clone())))
-                .additional_properties(Some(T::compose(schemas)))
+                .property_names(Some(schema_or_ref::<K>(&schemas, 0)))
+                .additional_properties(Some(schema_or_ref::<T>(&schemas, 1)))
                 .into()
         }
     }
```

JMLX42:
> The following changes fixes this for the `BTreeMap` but this also indicates that the default implementation needs to be also changed for other type than the `BTreeMap` as well.

Can you list me the other types to fix?

juhaku:
> > The following changes fixes this for the `BTreeMap` but this also indicates that the default implementation needs to be also changed for other type than the `BTreeMap` as well.
> 
> Can you list me the other types to fix?

Generally all the default `ComposeSchema` implementations in `utiopa::__dev {}` module according to the changes above. All of those implementations calls wrongly `T::compose(..)` But they need to be changed to above format to get the schema or to create ref.

juhaku:
It could be another PR though and I can also do that as well.

juhaku:
There is no need to change this though:
```rust
impl<T: ComposeSchema + ?Sized> PartialSchema for T {
    fn schema() -> crate::openapi::RefOr<crate::openapi::schema::Schema> {
        T::compose(Vec::new())
    }
}
```

juhaku:
All these should be changed:
```rust
    impl<T: ComposeSchema> ComposeSchema for Option<T> {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
            utoipa::openapi::schema::OneOfBuilder::new()
                .item(
                    utoipa::openapi::schema::ObjectBuilder::new()
                        .schema_type(utoipa::openapi::schema::Type::Null),
                )
                .item(T::compose(schemas))
                .into()
        }
    }

    impl<T: ComposeSchema> ComposeSchema for Vec<T> {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
            utoipa::openapi::schema::ArrayBuilder::new()
                .items(T::compose(schemas))
                .into()
        }
    }

    impl<T: ComposeSchema> ComposeSchema for std::collections::LinkedList<T> {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
            utoipa::openapi::schema::ArrayBuilder::new()
                .items(T::compose(schemas))
                .into()
        }
    }

    impl<T: ComposeSchema> ComposeSchema for [T] {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
            utoipa::openapi::schema::ArrayBuilder::new()
                .items(T::compose(schemas))
                .into()
        }
    }

    impl<T: ComposeSchema> ComposeSchema for &[T] {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
            utoipa::openapi::schema::ArrayBuilder::new()
                .items(T::compose(schemas))
                .into()
        }
    }

    impl<T: ComposeSchema> ComposeSchema for &mut [T] {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
            utoipa::openapi::schema::ArrayBuilder::new()
                .items(T::compose(schemas))
                .into()
        }
    }

    impl<K: ComposeSchema, T: ComposeSchema> ComposeSchema for std::collections::HashMap<K, T> {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
                .property_names(Some(K::compose(schemas.clone())))
                .additional_properties(Some(T::compose(schemas)))
                .into()
        }
    }

    impl<K: ComposeSchema + utoipa::ToSchema, T: ComposeSchema + utoipa::ToSchema> ComposeSchema
        for std::collections::BTreeMap<K, T>
    {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
            fn schema_or_ref<V: ComposeSchema + utoipa::ToSchema>(
                schemas: &[utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>],
                index: usize,
            ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
                let _ = <V as utoipa::PartialSchema>::schema;
                if let Some(composed) = schemas.get(index) {
                    composed.clone()
                } else {
                    utoipa::openapi::schema::RefBuilder::new()
                        .ref_location_from_schema_name(format!(
                            "{}",
                            <V as utoipa::ToSchema>::name()
                        ))
                        .into()
                }
            }

                .property_names(Some(schema_or_ref::<K>(&schemas, 0)))
                .additional_properties(Some(schema_or_ref::<T>(&schemas, 1)))
                .into()
        }
    }

    impl<K: ComposeSchema> ComposeSchema for std::collections::HashSet<K> {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
            utoipa::openapi::schema::ArrayBuilder::new()
                .items(K::compose(schemas))
                .unique_items(true)
                .into()
        }
    }

    impl<K: ComposeSchema> ComposeSchema for std::collections::BTreeSet<K> {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
            utoipa::openapi::schema::ArrayBuilder::new()
                .items(K::compose(schemas))
                .unique_items(true)
                .into()
        }
    }

    #[cfg(feature = "indexmap")]
    #[cfg_attr(doc_cfg, doc(cfg(feature = "macros", feature = "indexmap")))]
    impl<K: ComposeSchema, T: ComposeSchema> ComposeSchema for indexmap::IndexMap<K, T> {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
                .property_names(Some(K::compose(schemas.clone())))
                .additional_properties(Some(T::compose(schemas)))
                .into()
        }
    }

    #[cfg(feature = "indexmap")]
    #[cfg_attr(doc_cfg, doc(cfg(feature = "macros", feature = "indexmap")))]
    impl<K: ComposeSchema> ComposeSchema for indexmap::IndexSet<K> {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
            utoipa::openapi::schema::ArrayBuilder::new()
                .items(K::compose(schemas))
                .unique_items(true)
                .into()
        }
    }

    impl<'a, T: ComposeSchema + Clone> ComposeSchema for std::borrow::Cow<'a, T> {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
            T::compose(schemas)
        }
    }

    impl<T: ComposeSchema> ComposeSchema for std::boxed::Box<T> {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
            T::compose(schemas)
        }
    }

    impl<T: ComposeSchema> ComposeSchema for std::cell::RefCell<T> {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
            T::compose(schemas)
        }
    }

    #[cfg(feature = "rc_schema")]
    #[cfg_attr(doc_cfg, doc(cfg(feature = "macros", feature = "rc_schema")))]
    impl<T: ComposeSchema> ComposeSchema for std::rc::Rc<T> {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
            T::compose(schemas)
        }
    }

    #[cfg(feature = "rc_schema")]
    #[cfg_attr(doc_cfg, doc(cfg(feature = "macros", feature = "rc_schema")))]
    impl<T: ComposeSchema> ComposeSchema for std::sync::Arc<T> {
        fn compose(
            schemas: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
        ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
            T::compose(schemas)
        }
    }
```

utoipa-gen/tests/testdata/schema_generic_collect_non_inlined_schema
juhaku: Ah yes, true. :smile:

juhaku: I think this is solved, and will be fixed by changing the default implementations.

JMLX42:
I tried to remove some `#[schema(inline)]`. Apparently:

- it works when the schema is passed to `request_body` (cf `derive_path_test_collect_generic_array_request_body`) :heavy_check_mark: 
- it does *not* work when passed to`components(schemas())` (cf `derive_generic_schema_collect_recursive_schema_not_inlined`).

I've added a new test that highlights the issue: `generic_request_body_schema`.

juhaku:
> I tried to remove some `#[schema(inline)]`. Apparently:
> 
>     * it works when the schema is passed to `request_body` (cf `derive_path_test_collect_generic_array_request_body`) ✔️
> 
>     * it does _not_ work when passed to`components(schemas())` (cf `derive_generic_schema_collect_recursive_schema_not_inlined`) ❌.
> 
> 
> I've added a new test that highlights the issue: `generic_request_body_schema`.
> 
> Update: I think those problems now come from [the `if is_inlined` branch](https://github.com/juhaku/utoipa/blob/c45215c303ab288ed132a487c4b86b00b9513573/utoipa-gen/src/component.rs#L1219).

Right, indeed it creates the weird references., I'll check this too. You are right it might just be in the `is_inline` side the error now.

juhaku:
@JMLX42 
Now this comes to a bit annoying point. The `inline` was not defined for the `compose` at all. So what the `inline` does it actually calls `::schema()` function in hopes of getting the schema. while this works for normal types, it does not work for generic types which are composed.

E.g. If we have type like this below. And provide a generic type for `P` called `Type<i32>`. What ends up happening is it ends up calling `P::schema()` which is same as calling `<Type<i32> as PartialSchema>::schema()` which equals to `<Type<i32> as ComposeSchema>::compose(Vec::new())`. So what happens here is that it calls `compose` with empty generics array and it makes the type a reference. 
```rust
 // #[derive(ToSchema)]
struct Person<T: Sized, P> {
    field: T,
    // #[schema(inline)]
    t: P,
}

let person = <Person<String, Type<i32>> as utoipa::__dev::ComposeSchema>::compose(
    [
        <String as utoipa::PartialSchema>::schema(),
        <Type<i32> as utoipa::__dev::ComposeSchema>::compose(
            [<i32 as utoipa::PartialSchema>::schema()].to_vec(),
        ),
    ]
    .to_vec(),
);
dbg!(person);
```
This is what gets generated for `Person` and note here because of aforementioned `#[schema(inline)]` the `P` is called as  `P::schema()` but because it is a generic type it actually should be called with `P::compose(args)`.
```rust
impl<T: Sized, P> utoipa::__dev::ComposeSchema for Person<T, P>
where
    T: utoipa::ToSchema,
    P: utoipa::ToSchema,
{
    fn compose(
        mut generics: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
    ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
        {
            let mut object = utoipa::openapi::ObjectBuilder::new();
            object = object
                .property("field", {
                    let _ = <T as utoipa::PartialSchema>::schema;
                    if let Some(composed) = generics.get_mut(0usize) {
                        composed.clone()
                    } else {
                        utoipa::openapi::schema::RefBuilder::new()
                            .ref_location_from_schema_name(format!(
                                "{}",
                                <T as utoipa::ToSchema>::name()
                            ))
                            .into()
                    }
                })
                .required("field");
            let p = <P as utoipa::PartialSchema>::schema();
            dbg!("got p", &p);
            object = object
                .property("t", p)
                .required("t");
            object
        }
        .into()
    }
}
```

juhaku:
The problem is here: 
https://github.com/JMLX42/utoipa/blob/4ca70bdf0146bb05a8da4035d6a06ec4f5cbcbae/utoipa-gen/src/component.rs#L1220-L1233
We somehow should resolve here the actual type of `P` and depending whether it is a generic type it would call `compose` instead if `schema`. But I am not sure whether it actually is possible. Now the `compose` is called when has children` meaning that it actually has generic children.

juhaku:
If I change the implementation to this:
```diff
diff --git a/utoipa-gen/src/component.rs b/utoipa-gen/src/component.rs
index 70fb0a5..f6f1f68 100644
--- a/utoipa-gen/src/component.rs
+++ b/utoipa-gen/src/component.rs
@@ -1228,7 +1228,8 @@ impl ComponentSchema {
                             }
                         } else {
                             quote_spanned! {type_path.span()=>
-                                <#rewritten_path as utoipa::PartialSchema>::schema()
+                                // <#rewritten_path as utoipa::PartialSchema>::schema()
+                                <#rewritten_path as utoipa::__dev::ComposeSchema>::compose(generics.clone())
                             }
                         };
                         object_schema_reference.tokens = items_tokens.clone();
```
I get following error:
```bash
1. the trait bound `P: utoipa::__dev::ComposeSchema` is not satisfied
   the trait `utoipa::__dev::ComposeSchema` is not implemented for `P` [E0277]
```
Because `P` does not implement `ComposeSchema` and it is not required to be implemented. :face_exhaling: 

This kind of gets to point where the generics implementation in general needs some adjustment e.g. it should go `ComposeSchema` first and make sure everything implements it in first place. :thinking:  
```rust
impl<T: Sized, P> utoipa::__dev::ComposeSchema for Person<T, P>
where
    T: utoipa::ToSchema,
    P: utoipa::ToSchema,
{
    fn compose(
        mut generics: Vec<utoipa::openapi::RefOr<utoipa::openapi::schema::Schema>>,
    ) -> utoipa::openapi::RefOr<utoipa::openapi::schema::Schema> {
        {
            let mut object = utoipa::openapi::ObjectBuilder::new();
            object = object
                .property("field", {
                    let _ = <T as utoipa::PartialSchema>::schema;
                    if let Some(composed) = generics.get_mut(0usize) {
                        composed.clone()
                    } else {
                        utoipa::openapi::schema::RefBuilder::new()
                            .ref_location_from_schema_name(format!(
                                "{}",
                                <T as utoipa::ToSchema>::name()
                            ))
                            .into()
                    }
                })
                .required("field");
            object = object
                .property(
                    "t",
                    <P as utoipa::__dev::ComposeSchema>::compose(generics.clone()),
                )
                .required("t");
            object
        }
        .into()
    }
}
```

JMLX42:
Another problem: I did all of this because I wanted my responses to properly use ref for there type parameter.

Here is my code:

```rust
#[derive(Debug, Serialize, ToSchema)]
pub struct SuccessResponsePayload<T: ResourceObject> {
    #[schema(inline = false)]
    pub data: Vec<T>,
    #[schema(ignore = Self::ignore_included)]
    pub included: Option<Vec<T::RelationshipValue>>,
    #[schema(inline)]
    pub links: LinksObject<T>,
}

impl<T: ResourceObject> SuccessResponsePayload<T> {
    fn ignore_included() -> bool {
        !T::HAS_RELATIONSHIPS
    }
}

#[derive(Debug, Serialize, IntoResponses, ToSchema)]
#[serde(untagged)]
pub enum Response<T: ResourceObject> {
    #[response(
        status = 204,
        content_type = CONTENT_TYPE,
        description = "Success response with no content.",
    )]
    Empty,
    #[response(
        status = 200,
        content_type = CONTENT_TYPE,
        description = "Success response."
    )]
    Success(#[to_schema] SuccessResponsePayload<T>),
    #[response(
        status = "4XX",
        content_type = CONTENT_TYPE,
        description = "Client error response.",
        example = json!(EXAMPLE_ERROR_NOT_FOUND)
    )]
    ClientError(#[to_schema] ClientErrorResponse),
    #[response(
        status = "5XX",
        content_type = CONTENT_TYPE,
        description = "Server error response.",
        example = json!(EXAMPLE_ERROR_INTERNAL_SERVER_ERROR)
    )]
    ServerError(#[to_schema] ServerErrorResponse),
}
```

For some reason, `SuccessResponsePayload::data` is *still* inlined :cry: . I'll write a specific test for this scenario tomorrow.

> Now this comes to a bit annoying point. The inline was not designed for the compose at all 😆. It was there before the compose was a thing. So what the inline does it actually calls ::schema() function in hopes of getting the schema. while this works for normal types, it does not work for generic types which are composed.

The `generic_request_body_schema` passes if we inline all the way down though:

```rust
    #[derive(ToSchema)]
    #[schema(as = path::MyType<T>)]
    struct Type<T> {
        #[schema(inline)] // <= this is the change
        t: T,
    }

    #[derive(ToSchema)]
    struct Person<T: Sized, P> {
        field: T,
        #[schema(inline)]
        t: P,
    }
```

juhaku:
Yeah those `ToResponse` and `IntoResponses` have not been touched in this in this 5.0.0 release other than just to make it work. But they are not guaranteed to actually do the right thing. That is perhaps a topic of own. 

> The generic_request_body_schema passes if we inline all the way down though:

That works because the `compose` implementation for the `Type<i32>` returns either given composed generic type or a reference by default. If `inline` is used within the `Type<i32>` it will no use the compose implementation but directly call `i32::schema()` eventually and use that as the return value for `Type<i32>` compose function instead. This results the fact that whether <Type<i32>>::schema() or <Type<i32>>::compose(any_vec) is called they both return the `i32::schema()`.

JMLX42:
> Can the generated code check if the type is in `generics` and call `compose()` or `schema()` accordingly?

That's probably the expected goal of `generics.get_mut()` [here](https://github.com/juhaku/utoipa/blob/c45215c303ab288ed132a487c4b86b00b9513573/utoipa-gen/src/component.rs#L1283).

But the same check is not performed for [the `if is_inline` branch](https://github.com/juhaku/utoipa/blob/c45215c303ab288ed132a487c4b86b00b9513573/utoipa-gen/src/component.rs#L1219). Maybe that's what's missing.

juhaku:
> > Can the generated code check if the type is in `generics` and call `compose()` or `schema()` accordingly?
> 
> That's probably the expected goal of `generics.get_mut()` [here](https://github.com/juhaku/utoipa/blob/c45215c303ab288ed132a487c4b86b00b9513573/utoipa-gen/src/component.rs#L1283).
> 
> But the same check is not performed for [the `if is_inline` branch](https://github.com/juhaku/utoipa/blob/c45215c303ab288ed132a487c4b86b00b9513573/utoipa-gen/src/component.rs#L1219). Maybe that's what's missing.

https://github.com/juhaku/utoipa/blob/c45215c303ab288ed132a487c4b86b00b9513573/utoipa-gen/src/component.rs#L1220
The check is actually here it checks whether it has chidren, but in above case at that point it sees `P` as type which does not have children and not `Type<i32>`.

JMLX42:
> The check is actually here it checks whether it has chidren, but in above case at that point it sees `P` as type which does not have children and not `Type<i32>`.

Yes that's what I figured. But that's done at the macro time, not at runtime. Can't we have a better check at runtime using `generics.get_mut()`?

JMLX42:
> Yes that's what I figured. But that's done at the macro time, not at runtime. Can't we have a better check at runtime using `generics.get_mut()`?

I mean at some point, some code generates a `"$ref": "#/components/schemas/i32"` which does not make sense. I am having a hard time pinpointing where that code is.

juhaku:
> > `generics.get_mut()`?
> 
> I mean at some point, some code generates a `"$ref": "#/components/schemas/i32"` which does not make sense. I am having a hard time pinpointing where that code is.

Yeah, not sure whether there is any way to actually get around this.But the `inline` is taken out form the `t` it will work just fine, that is because it will go to the else block where it generates the type like shown below with the `generics.get_mut(generics)`.
> ```rust
>    #[derive(ToSchema)]
>    #[schema(as = path::MyType<T>)]
>    struct Type<T> {
>        #[schema(inline)] // <= this is the change
>        t: T,
>    }
>
>    #[derive(ToSchema)]
>    struct Person<T: Sized, P> {
>        field: T,
>        //#[schema(inline)]       // <--- here 
>        t: P,
>    }
>```
```rust
object = object
    .property("t", {
        let _ = <P as utoipa::PartialSchema>::schema;
        if let Some(composed) = generics.get_mut(1usize) {
            composed.clone()
        } else {
            utoipa::openapi::schema::RefBuilder::new()
                .ref_location_from_schema_name(format!(
                    "{}",
                    <P as utoipa::ToSchema>::name()
                ))
                .into()
        }
})
.required("t");
```

juhaku:
What's more is that when type type is E.g. wrapped with another type like in below example. If it is inlined with `#[schema(inline)]` it will create a reference to the type :rofl: But if it is not inlined, It will not create a reference but will just directly inline the type when `Person<T, P>` is being used. After all it was designed so that it will directly inline the generic arguments. 
```rust
    #[derive(ToSchema)]
    struct Wrapper {
        value: i32
    }

    #[derive(ToSchema)]
    struct Person<T: Sized, P> {
        field: T,
        #[schema(inline)]
        t: P,
    }
```

juhaku:
If we change the [`if is_inline` branch](https://github.com/juhaku/utoipa/blob/c45215c303ab288ed132a487c4b86b00b9513573/utoipa-gen/src/component.rs#L1219) like the following, then whether `#[schema(inline)]` is defined or not, the result will always be inlined.
```rust
diff --git a/utoipa-gen/src/component.rs b/utoipa-gen/src/component.rs
index 70fb0a5..216e22d 100644
--- a/utoipa-gen/src/component.rs
+++ b/utoipa-gen/src/component.rs
@@ -1227,8 +1227,20 @@ impl ComponentSchema {
                                 <#rewritten_path as utoipa::__dev::ComposeSchema>::compose(#composed_generics.to_vec())
                             }
                         } else {
-                            quote_spanned! {type_path.span()=>
-                                <#rewritten_path as utoipa::PartialSchema>::schema()
+                            let index = container.generics.get_generic_type_param_index(type_tree);
+                            dbg!("inline got index", index, &rewritten_path);
+                            if index.is_some() {
+                                quote_spanned! {type_path.span()=>
+                                    if let Some(composed) = generics.get(#index) {
+                                        composed.clone()
+                                    } else {
+                                        <#rewritten_path as utoipa::PartialSchema>::schema()
+                                    }
+                                }
+                                quote_spanned! {type_path.span()=>
+                                    <#rewritten_path as utoipa::PartialSchema>::schema()
+                                }
                             }
                         };
                         object_schema_reference.tokens = items_tokens.clone();
```

JMLX42:
@juhaku how about:

- we document this edge case
- we open a separate issue to track the fix

?

juhaku:
@JMLX42 I guess that is what we need to do. Otherwise this will never end. :laughing:

JMLX42:
> I guess that is what we need to do. Otherwise this will never end.

#1197

JMLX42:
@juhaku ready for review!

juhaku:
I guess with this the `schema(inline)` behavior will become bit unstable for generic arguments as can be seen from above discussions. I'll add more documentation for the `inline` attribute regarding this.

liningpan:
Could we get this in a new release? Thanks!

juhaku:
> Could we get this in a new release? Thanks!

Yup, it is there by now, and there is a fix coming in related to this as well.

PR #1188, opened 2024-11-05
Allow disabling syntax highlighting

xitep:


juhaku (changes requested):
Great :1st_place_medal: 

Could you add new CHANGLOG entry for this commit in: https://github.com/juhaku/utoipa/blob/master/utoipa-swagger-ui/CHANGELOG.md

On top of the file
```markdown

## Unreleased

### Added

* title (http://PR)

juhaku (approved):
Many thanks

PR #1199, opened 2024-11-09
fix: diverging axum route and openapi spec

boemjay:
Currently there is the possibility that the swagger-ui does not work with `utoipa_axum` in certain conditions.
The reason for this is, that the path resolution in the `.nest(router)` function is different in axum and utoipa.

This pull request fixes it by applying the same logic for the OpenAPI spec generation, that is used in axum.

## How to reproduce this issue
With the following minimal reproducible example, open the swagger ui on `localhost:3000/docs`.
The only endpoint is `/a/`, but when you perform the request, it returns `404 Not Found`.

```rust
use std::io;
use utoipa::OpenApi;
use utoipa_axum::{router::OpenApiRouter, routes};
use utoipa_swagger_ui::SwaggerUi;

#[derive(OpenApi)]
#[openapi()]
struct ApiDoc;

#[tokio::main]
async fn main() -> Result<(), io::Error> {
    #[utoipa::path(get, path = "/")]
    async fn nested_get_all() -> String {
        String::from("nested get all")
    }

    let (router, api) = OpenApiRouter::with_openapi(ApiDoc::openapi())
        .nest("/a", OpenApiRouter::new().routes(routes!(nested_get_all)))
        .split_for_parts();

    let router = router.merge(SwaggerUi::new("/docs").url("/apidoc/openapi.json", api));

    let listener = tokio::net::TcpListener::bind(("0.0.0.0", 3000)).await?;
    axum::serve(listener, router).await
}
```

## Cause of the issue
In Axum the `.nest` logic joins the prefix and nested path in a different way than utopia. E.g. for nested routes the trailing `/` is stripped away and other logic to prevent double `//` in the joined path.

### DIFF
--- a/utoipa-axum/CHANGELOG.md
+++ b/utoipa-axum/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog - utoipa-axum
 
+## Unreleased
+
+### Fixed
+
+* Fix diverging axum route and openapi spec (https://github.com/juhaku/utoipa/pull/1199)
+
 ## 0.1.2 - Oct 29 2024
 
 ### Changed
--- a/utoipa-axum/src/lib.rs
+++ b/utoipa-axum/src/lib.rs
@@ -313,7 +313,7 @@ mod tests {
         let paths = router.to_openapi().paths;
         let expected_paths = utoipa::openapi::path::PathsBuilder::new()
             .path(
-                "/api/customer/",
+                "/api/customer",
                 utoipa::openapi::PathItem::new(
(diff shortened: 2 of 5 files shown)


juhaku (changes requested):
This is pretty neat, :+1: 

However could you add a line to `CHANGELOG.md` for the addition at beginning of file.

```markdown

## Unreleased

### Fixed

* title (http;//PR)
```

juhaku:
I added the CHANGELOG markings for the PR :ok_hand:

Issue #1209, opened 2024-11-18
Cache assets downloaded in build.rs


thewh1teagle:
Eg. https://github.com/juhaku/utoipa/blob/master/utoipa-swagger-ui/build.rs#L21
Example how to cache them: https://github.com/thewh1teagle/sherpa-rs/blob/main/sys/build.rs#L459

PR #1214, opened 2024-11-21
feat(swagger-ui): cache swagger zip

thewh1teagle:
Enabled with

```console
cargo build -p utoipa-swagger-ui --features "cache,reqwest"
```

Resolve https://github.com/juhaku/utoipa/issues/1209#issuecomment-2491646267


utoipa-swagger-ui/src/internal/dirs.rs
juhaku, line 1: Maybe I am not educated enough, do we need all that C crap and pointer juggling? Aren't there a better and perhaps more simply way to the the same result or at least similar result?

As I understand this tries to use the users local cache dir for caching the data. I'd rather place it just the  `OUTPUT` dir of the `crate` in question. Or `target` dir since the `OUTPUT` dir can change?

juhaku: I'd expect at least some level of documentation what all that black magic does. Whats more is that I can guarantee that there are few that even dare to touch this code and I don't want to maintain this code.

utoipa-swagger-ui/src/internal/mod.rs
juhaku, line 1: Rather than using `path/to/internal/mod.rs` use `path/to/intenral.rs` This is a better syntax and is the recommended syntax over the `mod.rs` files. Eventually the code base ends up having a billion `mod.rs` files and that is not really compelling. 

See more here:
https://doc.rust-lang.org/book/ch07-02-defining-modules-to-control-scope-and-privacy.html#modules-cheat-sheet
https://doc.rust-lang.org/stable/book/ch07-05-separating-modules-into-different-files.html#alternate-file-paths
https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-modrs

The `mod.rs` syntax is old relic pre `2018 Edition` time (before `Rust 1.31`) just to make the Rust compiler support the modules.

juhaku (changes requested):
Still needs update to CHANGELOG.md

utoipa-swagger-ui/build.rs
```diff
@@ -24,6 +31,11 @@ const SWAGGER_UI_DOWNLOAD_URL_DEFAULT: &str =
 const SWAGGER_UI_DOWNLOAD_URL: &str = "SWAGGER_UI_DOWNLOAD_URL";
 const SWAGGER_UI_OVERWRITE_FOLDER: &str = "SWAGGER_UI_OVERWRITE_FOLDER";
 
+// wget <url> && sha256sum <file> | tr '[a-z]' '[A-Z]'
+#[cfg(feature = "cache")]
+const SWAGGER_UI_FILE_HASH: &str =
```
juhaku, line 36: But what happens if the user changes the url? I guess the validation will then just fail?

juhaku: Ah, it is only used as a suffix, well this is not very good to be hard coded because users can change the `SWAGGER_UI_DOWNLOAD_URL` which effectively is changing the version that changes the hash as well.

utoipa-swagger-ui/src/internal/dirs.rs
thewh1teagle: > Maybe I am not educated enough, do we need all that C crap and pointer juggling? Aren't there a better and perhaps more simply way to the the same result or at least similar result?

I also doesn't like that. though it's that expressive because the original dev wanted to support as many platforms as he can. but it's just for caching and I guess windows/linux/macos support is enough so we can simplify it into few lines

thewh1teagle: > As I understand this tries to use the users local cache dir for caching the data. I'd rather place it just the `OUTPUT` dir of the `crate` in question. Or `target` dir since the `OUTPUT` dir can change?

We shouldn't use OUTPUT dir the whole idea is to cache the downlloaded file in known location that will be still cached even if we rebuild the whole project / delete target folder and rebuild

thewh1teagle: > I'd expect at least some level of documentation what all that black magic does. Whats more is that I can guarantee that there are few that even dare to touch this code and I don't want to maintain this code.

It just returns the cache directory on different platforms

utoipa-swagger-ui/src/internal/mod.rs
thewh1teagle: Nice! didn't knew we can do the same thing without using mod.rs files. thanks

utoipa-swagger-ui/build.rs
thewh1teagle: > Ah, it is only used as a suffix, well this is not very good to be hard coded because users can change the `SWAGGER_UI_DOWNLOAD_URL` which effectively is changing the version that changes the hash as well.

I think we should avoid use cache if user supplied the URL

thewh1teagle:
Added documentation and simplified the implementation. If this looks good, I'll squash the commits.

thewh1teagle:
@juhaku 
Just a reminder, the PR now is really simple let me know if you can review / merge.

utoipa-swagger-ui/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### Added
 
+* Add `cache` feature to cache swagger ui zip in build script (https://github.com/juhaku/utoipa/pull/1214)
+
+### Added
+
```
juhaku, line 10:
```suggestion
```
We can remove these lines

juhaku (approved):
Nice, I like the approach better, simplier and easier to maintain. :+1:

juhaku:
> @juhaku Just a reminder, the PR now is really simple let me know if you can review / merge.

Yeah, thanks for poking around, it has been quite hectic few weeks which has let me to neglect the project.

thewh1teagle:
Added the suggested changes. You can squash directly in the Github UI just before merge

juhaku:
> Added the suggested changes. You can squash directly in the Github UI just before merge

Yup, I'll do that.

juhaku (approved):
Great :+1:

PR #1218, opened 2024-11-22
Add default to Example string fields

### LINKED ISSUE #1217: `RefOr<Example>` can't be deserialized if one of summary, description and external_value is missing
It makes OpenApi JSONs with multiple examples unable to be deserialized

Feiryn:
Closes #1217


juhaku (approved):
Nice, thanks for fixing :+1:

PR #1221, opened 2024-12-01
[swagger-ui] add basic authentication for internal use

realradical:
Add a basic auth config to SwaggerUI struct. In case we'd like to protect it by username and password.
 add the support for Axum, Actix and Rocket.

### DIFF
--- a/utoipa-swagger-ui/Cargo.toml
+++ b/utoipa-swagger-ui/Cargo.toml
@@ -22,7 +22,7 @@ vendored = ["dep:utoipa-swagger-ui-vendored"]
 [dependencies]
 rust-embed = { version = "8" }
 mime_guess = { version = "2.0" }
-actix-web = { version = "4", optional = true, default-features = false }
+actix-web = { version = "4", optional = true, default-features = false, features = ["macros"] }
 rocket = { version = "0.5", features = ["json"], optional = true }
 axum = { version = "0.7", default-features = false, features = [
     "json",
--- a/utoipa-swagger-ui/Cargo.toml
+++ b/utoipa-swagger-ui/Cargo.toml
@@ -32,6 +32,7 @@ utoipa = { version = "5.0.0", path = "../utoipa", default-features = false, feat
 ] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0" }
+base64 = { version = "0.22.1" }
 
 [dev-dependencies]
 axum-test = "16.2.0"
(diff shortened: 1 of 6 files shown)


utoipa-swagger-ui/Cargo.toml
```diff
@@ -32,6 +32,7 @@ utoipa = { version = "5.0.0", path = "../utoipa", default-features = false, feat
 ] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0" }
+base64 = { version = "0.22.1" }
```
juhaku, line 35: Since this is only enabled by in `axum` by now, I would make this dependency optional and enable it when `axum` is enabled.

juhaku (changes requested):
Quite nice, however still needs CHANGELOG entry to top of unreleased added section with title and link to the PR.

utoipa-swagger-ui/Cargo.toml
realradical: i'll try to make it work for rocket and acix

realradical:
> Quite nice, however still needs CHANGELOG entry to top of unreleased added section with title and link to the PR.

done

utoipa-swagger-ui/src/lib.rs
```diff
@@ -1317,6 +1341,12 @@ impl From<String> for Config<'_> {
     }
 }
 
+#[derive(Serialize, Clone)]
```
juhaku, line 1344:
```suggestion
/// Basic auth options for Swagger UI. By providing `BasicAuth` to `Config::basic_auth` the access to the
/// Swagger UI can be restricted behind given basic authentication. 
#[derive(Serialize, Clone)]
```

utoipa-swagger-ui/src/lib.rs
```diff
@@ -1317,6 +1341,12 @@ impl From<String> for Config<'_> {
...
+pub struct BasicAuth {
+    pub username: String,
```
juhaku, line 1346:
```suggestion
    /// Username for the `BasicAuth`
    pub username: String,
```

utoipa-swagger-ui/src/lib.rs
```diff
@@ -1317,6 +1341,12 @@ impl From<String> for Config<'_> {
...
+    pub password: String,
```
juhaku, line 1347:
```suggestion
    /// Password of the _`username`_ for the `BasicAuth`
    pub password: String,
```

PR #1223, opened 2024-12-01
Exhaustive pattern matching (include `MacroArg::Query` case)

SamuelMarks:
Got a bunch of warnings when running `cargo make`; resolved the ones which seem to be True Positives.


juhaku (approved):
Nice, thanks, I might have missed those. :+1:

PR #1233, opened 2024-12-12
`Info::from_env()` sets `License::identifier`

masnagam:
This PR changes `Info::from_env()` to set `info.license.identifier`.

### DIFF
--- a/utoipa-gen/Cargo.toml
+++ b/utoipa-gen/Cargo.toml
@@ -43,6 +43,7 @@ smallvec = { version = "1.10", features = ["serde"] }
 rust_decimal = { version = "1", default-features = false }
 chrono = { version = "0.4", features = ["serde"] }
 assert-json-diff = "2"
+assert_matches = "1.5.0"
 time = { version = "0.3", features = ["serde-human-readable"] }
 serde_with = "3.0"
 
--- a/utoipa-gen/Cargo.toml
+++ b/utoipa-gen/Cargo.toml
@@ -43,7 +43,6 @@ smallvec = { version = "1.10", features = ["serde"] }
 rust_decimal = { version = "1", default-features = false }
 chrono = { version = "0.4", features = ["serde"] }
 assert-json-diff = "2"
-assert_matches = "1.5.0"
 time = { version = "0.3", features = ["serde-human-readable"] }
 serde_with = "3.0"
 
--- a/utoipa-gen/src/openapi/info.rs
+++ b/utoipa-gen/src/openapi/info.rs
@@ -41,7 +41,15 @@ impl Info<'_> {
                     Some(contact)
                 }
             });
-        let license = std::env::var("CARGO_PKG_LICENSE").ok().map(License::from);
+        let license = std::env::var("CARGO_PKG_LICENSE")
+            .ok()
+            .map(|spdx_expr| License {
+                name: Cow::Owned(spdx_expr.clone()),
+                // CARGO_PKG_LICENSE contains an SPDX expression as described in the Cargo Book.
+                // It can be set to `info.license.identifier`.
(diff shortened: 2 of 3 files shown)


utoipa-gen/src/openapi/info.rs
```diff
@@ -41,7 +41,15 @@ impl Info<'_> {
                     Some(contact)
                 }
             });
-        let license = std::env::var("CARGO_PKG_LICENSE").ok().map(License::from);
+        let license = std::env::var("CARGO_PKG_LICENSE")
```
juhaku, line 44: I'd perhaps just update the `From<T>` implementation for License to support the `info.license.identifier` as well.

utoipa-gen/Cargo.toml
```diff
@@ -43,6 +43,7 @@ smallvec = { version = "1.10", features = ["serde"] }
 rust_decimal = { version = "1", default-features = false }
 chrono = { version = "0.4", features = ["serde"] }
 assert-json-diff = "2"
+assert_matches = "1.5.0"
```
juhaku, line 46: Perhaps there is no need for additional dependency? :thinking: Even making this kind of macro by hand should be quite trivial. Or perhaps there is no need to even a macro. Just typing `match {}` with `panic` on any branch.

juhaku (changes requested):
Looks good, just missing CHANGELOG.md update. Add to `Unreleased` section new  title of not exists `Changed` and then under that a a bullet point for the PR. Try to keep the order, newest first.

utoipa-gen/src/openapi/info.rs
masnagam: According to [the openapi specification](https://swagger.io/specification/), the `name` of the `license` object is a *name* of the license. For example `Apache 2.0`. And it's may not a SDPX license identifier.

If the `string` argument of `License::from()` is a SDPX license expression, it's ok to set the `string` value to the `name` and `identifier`. However, if the `string` argument is a license name, its value should not be set to the `identifier`.

utoipa-gen/Cargo.toml
masnagam: You're right. I'll modify the test I added.

[`assert_matches!()`](https://doc.rust-lang.org/std/?search=assert_matches) is one of standard macros, but available only in nightly at this point (will be stabilized in the future). So, I usually add the [`assert_matches`](https://crates.io/crates/assert_matches) crate and use the same name macro in my tests by habit.

masnagam:
> Looks good, just missing CHANGELOG.md update. Add to `Unreleased` section new title of not exists `Changed` and then under that a a bullet point for the PR. Try to keep the order, newest first.

updated in 556264140854957cf7b1abc647619e603e33ae87

utoipa-gen/src/openapi/info.rs
masnagam: modified in 4c501ce97105e23a99bc43168e8845fa16942961

utoipa-gen/Cargo.toml
masnagam: modified in 4c501ce97105e23a99bc43168e8845fa16942961

juhaku: Yup, seems a good crate though, I just try to keep the dependencies as minimal as possible for what comes to the library. Though eventually when utoipa will use the minimum rust version containing this macro it can be naturally used. Now though for users we need to support rust some generations backwards.

utoipa-gen/src/openapi/info.rs
juhaku: True :+1:

juhaku (approved):
Nice, thanks for the patch :100:

PR #1235, opened 2024-12-13
[utoipa-gen] request_body: Simplify ToTokensDiagnostics

jsoo1:
I noticed the tokens were traversed multiple times in confusing ways, and thought it could use a touch up.


juhaku (changes requested):
Could use a CHANGELOG.md entry under `Unreleased` title and `Changed` subtitle. Newest on top order.

Great, thanks for patch. :+1:

Issue #1087, opened 2024-10-04
Missing "encoding" field


JMLX42:
Hello,

I want to make sure the Swagger UI file dialog will limit the selectable file(s) to a specific extension (in this case `.glb`).
Based on my discussion with ChatGPT and my read of the OAS spec, it should be done using [the `encoding` object](https://spec.openapis.org/oas/v3.0.3#encoding-object).

## How to reproduce

```rust
#[derive(ToSchema, MultipartForm)]
pub(crate) struct GlbFileUpload {
    /// The binary glTF (`.glb`) file to import.
    #[schema(format = Binary, encoding = "model/gltf-binary", value_type=Vec<u8>)]
    file: TempFile,
}

/// Import a glTF file
///
/// Imports a glTF file from a multipart form data.
/// This endpoint accepts a `.glb` file and supports additional query parameters.
#[utoipa::path(
    tag = "glTF",
    params(
        GltfResourceQueryParameters,
    ),
    request_body(content = inline(GlbFileUpload), content_type = "multipart/form-data"),
    responses(GltfResourceResponse),
)]
#[post("/gltf/import")]
pub async fn import_gltf_file(
    req: HttpRequest,
    db: web::Data<Database>,
    query: web::Query<GltfResourceQueryParameters>,
    MultipartForm(payload): MultipartForm<GlbFileUpload>,
) -> Result<HttpResponse<actix_web::body::BoxBody>, ErrorList> {
    info!(
        "received `{}` ({} bytes) for import",
        payload.file.file_name.unwrap(),
        payload.file.size
    );

    let file = Arc::new(payload.file.file);
    let db = db.into_inner();

    Ok(
        ResponseBuilder::<GltfResourceObject>::new(req.uri().clone())
            .relationship_resolver(AssetResolver::new(db.clone()))
            .relationship_resolver(SceneResolver::new(db.clone()))
            .query_parameters(query.into_inner())
            .data(Box::new(move || {
                let db = db.clone();
                let file = file.clone();

                Box::pin(async move {
                    gltf_live_importer::import_from_reader(file.as_ref(), &db)
                        .await
                        .map(|r| r.into_resource_object())
                })
            }))
            .build()
            .await
            .into(),
    )
}
```

## OAS Spec

[...]

PR #1237, opened 2024-12-13
[utoipa-gen] MediaTypeAttr, Encoding: Parse encodings

jsoo1:
Though only in a limited fashion - notable fields are still not
implemented.

Though this does let you do:

```rs
   request_body(
       content(...)
       encoding(
           ("field_name", (
               content_type = "application/json",
               explode = true,
               allow_reserved = true,
           ))
       )
   )
```

Partially addresses #1087

### DIFF
--- a/utoipa-gen/src/path/media_type.rs
+++ b/utoipa-gen/src/path/media_type.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::collections::BTreeMap;
 
 use proc_macro2::TokenStream;
 use quote::{quote, ToTokens};
@@ -16,11 +17,15 @@ use crate::{parse_utils, AnyValue, Array, Diagnostics, ToTokensDiagnostics};
 use super::example::Example;
 use super::PathTypeTree;
 
+pub mod encoding;
+
+use encoding::Encoding;
+
 /// Parse OpenAPI Media Type object params
 /// ( Schema )
 /// ( Schema = "content/type" )
 /// ( "content/type", ),
-/// ( "content/type", example = ..., examples(..., ...), encoding(...) )
+/// ( "content/type", example = ..., examples(..., ...), encoding(("exampleField" = (...)), ...) )
 #[derive(Default)]
 #[cfg_attr(feature = "debug", derive(Debug))]
 pub struct MediaTypeAttr<'m> {
@@ -28,7 +33,7 @@ pub struct MediaTypeAttr<'m> {
     pub schema: Schema<'m>,
     pub example: Option<AnyValue>,
     pub examples: Punctuated<Example, Comma>,
-    // econding: String, // TODO parse encoding
+    pub encoding: BTreeMap<String, Encoding>,
 }
 
 impl Parse for MediaTypeAttr<'_> {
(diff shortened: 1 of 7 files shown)


utoipa-gen/src/path/media_type/encoding.rs
```diff
@@ -0,0 +1,92 @@
...
+use syn::parse::{Parse, ParseStream};
+use syn::{parenthesized, Error, LitBool, LitStr, Token};
+
+use crate::parse_utils;
+
+// (content_type = "...", explode = true, allow_reserved = false,)
+#[derive(Default)]
+#[cfg_attr(feature = "debug", derive(Debug))]
+pub struct Encoding {
+    pub(super) content_type: Option<String>,
```
juhaku, line 12: This could be `parse_utils::LitStrOrExpr`. This allows users to use a valid rust expression instead just literal str inlined to the macro.

utoipa-gen/src/path/media_type/encoding.rs
```diff
@@ -0,0 +1,92 @@
...
+
+        let mut encoding = Encoding::default();
+
+        while !content.is_empty() {
+            let ident = content.parse::<Ident>()?;
+            let attribute_name = &*ident.to_string();
+            match attribute_name {
+                "content_type" => {
+                    encoding.content_type = Some(
+                        parse_utils::parse_next(&content, || content.parse::<LitStr>())?
```
juhaku, line 33: Could use `parse_utils::parse_next_literal_str_or_expr(&inner)?`

utoipa-gen/src/path/media_type/encoding.rs
```diff
@@ -0,0 +1,92 @@
...
+                "explode" => {
+                    encoding.explode = Some(
+                        parse_utils::parse_next(&content, || content.parse::<LitBool>())?
+                            .value(),
+                    )
+                }
+                "allow_reserved" => {
+                    encoding.allow_reserved = Some(
+                        parse_utils::parse_next(&content, || content.parse::<LitBool>())?.value()
+                    )
```
juhaku, line 48: These could use `parse_utils::parse_bool_or_true(...)`. This allows users to omit the `bool` value if true is expected. This behavior is already in many places of utoipa.

utoipa-gen/src/path/media_type.rs
```diff
@@ -90,28 +95,56 @@ impl<'m> MediaTypeAttr<'m> {
...
+
+                impl Parse for KV {
+                    fn parse(input: ParseStream) -> syn::Result<Self> {
+                        let key_val;
+
+                        syn::parenthesized!(key_val in input);
+
+                        let k = key_val.parse::<syn::LitStr>()?.value();
+
+                        key_val.parse::<Comma>()?;
```
juhaku, line 127: I think instead of comma, this could be `=` sign. This would then match better with the syntax already at present in elsewhere e.g. links.

juhaku (changes requested):
Looks good, it is missing CHANGELOG.md entry for the addition. Perhaps under `Unreleased` and `Added` section by placing newest to top.

utoipa-gen/src/path/media_type.rs
jsoo1: Sounds good. In that case I'll go with the `parse_next` functions from the utils.

utoipa-gen/src/path/media_type/encoding.rs
jsoo1: Um wait, I am not going to do this yet because I am actually pretty certain the default for `explode` and `allow_reserved` would not be true, but I don't use them. If they are supposed to default to true I would be happy to change it.

jsoo1: Nvm, did it.

juhaku (approved):
Super, thanks for putting the effort :slightly_smiling_face:

PR #1238, opened 2024-12-13
utoipa-axum: Allow trailing comma in `routes!()` macro

Turbo87:
Once the `routes!()` invocation contains more than a couple of routes, it is likely that each route will be on its own line. Currently, the macro does not support trailing commas, which causes the git diff when adding a new route to be unnecessarily large. With this change trailing commas are supported, which shrinks the diff down to a single-line change.

### DIFF
--- a/utoipa-axum/CHANGELOG.md
+++ b/utoipa-axum/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Changed
+
+* Allow trailing comma in `routes!()` macro (https://github.com/juhaku/utoipa/pull/1238)
+
 ### Fixed
 
 * Fix axum path nesting (https://github.com/juhaku/utoipa/pull/1231)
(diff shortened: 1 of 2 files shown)


Turbo87:
Admittedly, I misunderstood from the documentation that `.routes(routes!(...))` is supposed to be called per-endpoint instead of having one long list... but this change IMHO still seems useful :D

Turbo87:
@juhaku any thoughts on this? :)

juhaku:
Well, I don't have a strong opinion on trailing comma. Though with such support users can decide whether they want to use the trailing comma or not, so why not.

juhaku (changes requested):
Small and simple, just needs CHANGELOG.md entry. Newest entry to the top. This should come under title `Unreleased` and sub title `Changed`. 

Also seems to need `cargo fmt`. I should setup pre commit hook, but haven't got it done :smiling_face_with_tear:

Turbo87:
rebased and feedback addressed :)

juhaku (approved):
Seems solid, thanks :100:

PR #1243, opened 2024-12-17
Use a re-exported serde_json dependency in macros

### LINKED ISSUE #1172: Undocumented serde_json dependency
I'm a new to rust, so not sure if this is an actual issue or I missed some documentation somewhere, but I get a compile error saying `use of undeclared crate or module serde_json` when using #[schema(default = something)]] in a struct deriving ToSchema.

It's easy enough to resolve by adding serde_json manually, just couldn't find any documentation of it as a dependency or features I might be missing.

Package versions and code for minimal replication.

```
[dependencies]
serde = { version = "1.0.214", features = ["derive"] }
utoipa = { version = "5.1.3", features = ["uuid", "axum_extras", "macros"] }
uuid = { version = "1.11.0", features = ["serde"] }
```

```
use serde::Deserialize;
use utoipa::{IntoParams, ToSchema};

#[derive(Deserialize, IntoParams, ToSchema)]
pub struct Pagable {
    #[schema(default = 0)]
    pub page: Option<i64>,
    #[schema(default = 10)]
    pub page_size: Option<i64>,
}
```

Sufflope:
Hello, here is a proposal to avoid shadow-depending on `serde_json` when using derives.

Fixes #1172

### DIFF
--- a/examples/actix-web-multiple-api-docs-with-scopes/Cargo.toml
+++ b/examples/actix-web-multiple-api-docs-with-scopes/Cargo.toml
@@ -13,7 +13,6 @@ authors = [
 [dependencies]
 actix-web = "4"
 serde = { version = "1.0", features = ["derive"] }
-serde_json = "1.0"
 env_logger = "0.11.0"
 log = "0.4"
 futures = "0.3"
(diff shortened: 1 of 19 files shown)


juhaku (approved):
Great, thanks for the patch :tada:

juhaku:
Just one thing, a CHANGELOG.md entry for the change would be good to add in `utoipa/CHANGELOG.md`and `utoipa-gen/CHANGELOG.md`. Newest entry to the top and should be under the `Unreleased` and `Changed` title.

juhaku (approved):
Great, thanks :+1:

Issue #1040, opened 2024-09-09
Switch to using cargo-make


jayvdb:
Would you be interested in me replacing test.sh with a Makefile.toml

https://github.com/sagiegurari/cargo-make

I can do all the CI stuff to make this work.

juhaku:
Why not, I have been actually weighting upon that or justfile or just plain make. :laughing: But I haven't got anywhere with that choice. Sure if you want to set things up that would be great. That test file get's anyway quite cumbersome to maintain. One thing to add is probably coverage as well at somepoint but haven't really got to it.

jayvdb:
Both cargo-make and just are supported by https://github.com/taiki-e/install-action/blob/main/TOOLS.md , which speeds up the CI installing of the tool.

just has more stars - I havent used that yet, so I will try it first.

juhaku:
Yeah, sure thing, for what I understand the justfile basically 60 % makefile and then the rest is build on top of it to enhance the quirks of old fashioned makefile to give smooth command running experience. 

For example the acitx-web is using justfile

Drifter-J:
The proposed change is moot because a justfile was adopted. Hence closing.

PR #1247, opened 2024-12-19
utopia-gen: Use `insta` for snapshot testing

Turbo87:
I saw in https://github.com/juhaku/utoipa/discussions/145 that snapshot testing had been considered in the past, but then `assert-json-diff` was used instead. This PR migrates a couple of the `utoipa-gen` tests to use `insta` instead of performing a basic string comparison.

`insta` produces diffs like below, which makes it much easier to see at first glance what changed. The `cargo-insta` extension also makes it very fast to update snapshots in case of any desired changes.

```
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot file: utoipa-gen/tests/snapshots/schema_generics__generic_schema_full_api.snap
Snapshot: generic_schema_full_api
Source: utoipa-gen/tests/schema_generics.rs:221
────────────────────────────────────────────────────────────────────────────────
Expression: doc
────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────────
   64    64 │                     "items"
   65    65 │                   ],
   66    66 │                   "properties": {
   67    67 │                     "items": {
   68       │-                      "type": "object",
         68 │+                      "type": "array",
   69    69 │                       "items": {
   70    70 │                         "type": "object",
   71    71 │                         "required": [
   72    72 │                           "id",
────────────┴───────────────────────────────────────────────────────────────────
To update snapshots run `cargo insta review`
```

If this is considered useful we could potentially also migrate the other tests to use snapshots instead of `assert-json-diff`.

### DIFF
--- a/utoipa-gen/CHANGELOG.md
+++ b/utoipa-gen/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ### Changed
 
+* Use `insta` for snapshot testing (https://github.com/juhaku/utoipa/pull/1247)
 * Make `parse_named_attributes` a method of `MediaTypeAttr` (https://github.com/juhaku/utoipa/pull/1236)
 * Use a re-exported `serde_json` dependency in macros instead of implicitly requiring it as dependency in end projects (https://github.com/juhaku/utoipa/pull/1243)
 * Simplified `ToTokensDiagnostics` for `request_body` (https://github.com/juhaku/utoipa/pull/1235)
--- a/utoipa-gen/Cargo.toml
+++ b/utoipa-gen/Cargo.toml
@@ -45,6 +45,7 @@ chrono = { version = "0.4", features = ["serde"] }
 assert-json-diff = "2"
 time = { version = "0.3", features = ["serde-human-readable"] }
 serde_with = "3.0"
+insta = { version = "1.41", features = ["json"] }
 
 [features]
 # See README.md for list and explanations of features
--- a/utoipa-gen/tests/openapi_derive.rs
+++ b/utoipa-gen/tests/openapi_derive.rs
@@ -1,6 +1,7 @@
 use std::{borrow::Cow, marker::PhantomData};
 
 use assert_json_diff::{assert_json_eq, assert_json_include};
+use insta::assert_json_snapshot;
 use serde::Serialize;
 use serde_json::{json, Value};
 use utoipa::{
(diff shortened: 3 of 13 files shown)


juhaku:
Try running tests with `scripts/test.sh` you can also specify crate by `scripts/test.sh utoipa` for example. I should setup a proper build and testing with Makefile or justfile or cargo make and there is an issue for it https://github.com/juhaku/utoipa/issues/1040 but it hasn't been started yet.

juhaku:
The tests seems to fail because there are snapshots that are not generated for tests run with `rc_schema` feature flag on seemingly.

Turbo87:
yeah, I must've missed running them for some reason. should be fixed now though :)

juhaku:
And like before, the CHANGELOG.md entry would be useful :slightly_smiling_face: It is bit annoying extra step but allows better change tracking for other developers who are interested of changed details.

Turbo87:
> And like before, the CHANGELOG.md entry would be useful 🙂 It is bit annoying extra step but allows better change tracking for other developers who are interested of changed details.

hmm, which changelog would this go into (utoipa-gen?)? and under which category (Changes?)?




> Perhaps we could push this `insta` bit further and migrate to it all the way. Though not necessarily in this PR as it is quite a lot of tests but could be done in future. 🤔

yep, should be relatively straight-forward to incrementally migrate the other tests too. 

open question would be whether to use [inline snapshots](https://insta.rs/docs/snapshot-types/#inline-snapshots) or regular ones like in this PR. the inline snapshots are closer to the current usage with `json!()` directly in the test code, but they are slightly more verbose and require recompiling the tests whenever a snapshot is changed.

I personally tend to use inline snapshots when the snapshot is relatively short, but with the amount of data that we're snapshotting here, I think I would use regular ones.

juhaku:
> hmm, which changelog would this go into (utoipa-gen?)? and under which category (Changes?)?

`Changed` is good enough category, and yes to the CHANGELOG of a project the changes are made to. Should write the code of conduct files etc. for this purpose :smile:

Turbo87:
okay, should be added correctly now :)

juhaku:
> yep, should be relatively straight-forward to incrementally migrate the other tests too.
> 
> open question would be whether to use [inline snapshots](https://insta.rs/docs/snapshot-types/#inline-snapshots) or regular ones like in this PR. the inline snapshots are closer to the current usage with `json!()` directly in the test code, but they are slightly more verbose and require recompiling the tests whenever a snapshot is changed.
> 
> I personally tend to use inline snapshots when the snapshot is relatively short, but with the amount of data that we're snapshotting here, I think I would use regular ones.

Yeah, probably which ever comes easier, since there will be quite a few tests. I guess convention to keep the test file name same as the test function name makes it easier to find.

Issue #1215, opened 2024-11-22
Query params should not be nullable


sergiimk:
Consider a query params struct like this:
```rust
#[derive(Debug, serde::Deserialize, utoipa::IntoParams)]
#[serde(rename_all = "camelCase", deny_unknown_fields)]
#[into_params(parameter_in = Query)]
pub struct QueryParams {
    pub schema_format: Option<SchemaFormat>,
}

#[derive(
    Debug,
    Default,
    Clone,
    Copy,
    PartialEq,
    Eq,
    serde::Serialize,
    serde::Deserialize,
    utoipa::ToSchema,
)]
pub enum SchemaFormat {
    #[default]
    ArrowJson,
    Parquet,
    ParquetJson,
}
```

Utoipa generates a spec like this:

```json
{
  "openapi": "3.1.0",
  "paths": {
    "/query": {
      "get": {
        "operationId": "query_handler",
        "parameters": [
          {
            "description": "How to encode the schema of the result",
            "in": "query",
            "name": "schemaFormat",
            "required": false,
            "schema": {
              "oneOf": [
                {
                  "type": "null"
                },
                {
                  "$ref": "#/components/schemas/SchemaFormat"
                }
              ]
            }
          }
       ]
      }
    }
  },
  "components": {
    "schemas": {
      "SchemaFormat": {
        "enum": [
          "ArrowJson",
          "Parquet",
          "ParquetJson"
        ],
        "type": "string"
      }
    }
  }
}
```

When we run [IBM openapi-validator](https://github.com/IBM/openapi-validator) on this spec it gives an error:
```
Errors:

  Message :   Invalid type; valid types are: array, boolean, integer, number, object, string
  Rule    :   ibm-schema-type-format
  Path    :   paths./query.get.parameters.5.schema.oneOf.0
```
I think the validator is correct here, as in case of a query parameter there cannot be `null` value, as it would be non-distinguishable from a `"null"` string.

My expectation is that for an optional query parameter utoipa should only specify `"required": false` but not make value nullable.

Issue #622, opened 2023-05-18
Wrapping axum query params makes them required


TheDan64:
I have some optional query params in axum:
```rust
#[derive(Debug, Deserialize, IntoParams, Validate)]
pub struct ReadManyParams {
    /// Name to start after
    start_after: Option<String>,
    /// Limits the number of results
    #[validate(range(min = 1, max = 1000))]
    per_page: Option<usize>,
}
```

I created a wrapper validation type to call the [validator](https://crates.io/crates/validator) crate. For example, in this case it ensures `per_page` query param is in [1, 1000]:

```rust
#[derive(Clone, Copy, Debug)]
pub struct Validated<T>(pub T);

#[async_trait]
impl<T, S> FromRequestParts<S> for Validated<T>
where
    T: Deref + FromRequestParts<S>,
    T::Target: Validate,
    S: Sync,
{
    type Rejection = ValidationErrorOr<T::Rejection>;

    async fn from_request_parts(parts: &mut Parts, state: &S) -> Result<Self, Self::Rejection> {
        let value = T::from_request_parts(parts, state)
            .await
            .map_err(ValidationErrorOr::Other)?;
        value.validate()?;
        Ok(Validated(value))
    }
}
```

(This is recommended in the [axum examples](https://github.com/tokio-rs/axum/blob/main/examples/validator/src/main.rs#L60))

My endpoint looks something like this:
```rust
#[utoipa::path(
    get,
    path = "/api/whatever",
    params(ReadManyParams),
)]
pub async fn read_many(
    Validated(Query(params)): Validated<Query<ReadManyParams>>,
) -> ... {
    ...
}
```

This mostly works, without `Validated`, both of the query parameters are marked optional by utoipa, which is correct. However, after wrapping it in my `Validated` type, they still show up but are now marked as required, which is incorrect since they're both optional and swagger UI is now too strict to use

![image](https://github.com/juhaku/utoipa/assets/3749239/e7b62281-1609-4911-a95c-456d810f75d9)

TheDan64:
I also just noticed that, even if you specify those incorrectly required fields, they won't show up in the request

eblocha:
I've found I need to wrap the struct with `#[into_params(parameter_in = Query)]`:

```rust
#[derive(Debug, Deserialize, Default, IntoParams)]
#[into_params(parameter_in = Query)]
pub struct TreeQueryParams {
    pub parent_id: Option<Uuid>,
}
```

[...]

Issue #609, opened 2023-05-02
IntoParams marks Options as required


FloppyDisck:
Hello, the below snippet shows all of the values as required in the Swagger doc
```rust
#[derive(Deserialize, IntoParams)]
struct Request {
    ids: Option<String>,
    with_price_ids: Option<String>,
    without_flags: Option<String>,
    with_flags: Option<String>,
}
```

This is the generated Open API spec
```json
"parameters": [
          {
            "name": "ids",
            "in": "path",
            "required": true,
            "schema": {
              "type": "string",
              "nullable": true
            }
          },
          {
            "name": "with_price_ids",
            "in": "path",
            "required": true,
            "schema": {
              "type": "string",
              "nullable": true
            }
          },
          {
            "name": "without_flags",
            "in": "path",
            "required": true,
            "schema": {
              "type": "string",
              "nullable": true
            }
          },
          {
            "name": "with_flags",
            "in": "path",
            "required": true,
            "schema": {
              "type": "string",
              "nullable": true
            }
          }
        ],
```

My expectation is that any optional value is not set as not required by default, or at least have some sort of param to explicitly set as not required, such as `#[param(not-required)]`

carueda:
This seems like a duplicate of https://github.com/juhaku/utoipa/issues/529

juhaku:
@carueda @FloppyDisck 

> This seems like a duplicate of https://github.com/juhaku/utoipa/issues/529

Yes.

The field nullability and requirability is defined here https://docs.rs/utoipa/latest/utoipa/derive.ToSchema.html#field-nullability-and-required-rules. Any other behavior should be considered a bug. 

But in short just defining a `Option<Type>` does not make it non-required to better support the code generation libraries and based on ideology that `null` serializes to `None` and vise versa. If a parameter need to be non-required then it **must** have a default value.

carueda:
Thanks @juhaku .

In case it helps @FloppyDisck, I faced some similar issue:

```rust
#[derive(Deserialize, IntoParams, Debug)]
#[serde(rename_all = "camelCase")]
pub struct PositionsQuery {
    /// Maximum number of last positions to report
    last_number_of_fixes: Option<u32>,
    /// Lower limit for time range restriction
    start_date: Option<String>,
    /// Upper limit for time range restriction
    end_date: Option<String>,
}

/// Get latest platform positions.
#[utoipa::path(
    get,
    path = "/trackdb/platforms/{platform_id}/positions",
    params(
[...]

PR #1248, opened 2024-12-19
utopia-gen: Adjust params code to not set `nullable` on `Option` for `Query` params

Turbo87:
This PR resolves https://github.com/juhaku/utoipa/issues/1215 by ensuring that `Option<_>` usage in `Query` params is not transformed into a nullable feature.

https://swagger.io/docs/specification/v3_0/describing-parameters/#default-parameter-values gives an example on how `required: false` query parameters are described and there is no `nullable: true` or `schema.type: null` in there, which indicates that `required: false` should be sufficient to describe an `Option<_>` type for query parameters.

(Potentially) Related:

- https://github.com/juhaku/utoipa/issues/1215
- https://github.com/juhaku/utoipa/pull/530
- https://github.com/juhaku/utoipa/issues/609
- https://github.com/juhaku/utoipa/issues/1058
- https://github.com/juhaku/utoipa/issues/622

### DIFF
--- a/utoipa-gen/CHANGELOG.md
+++ b/utoipa-gen/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ### Changed
 
+* Adjust params code to not set `nullable` on `Option` for `Query` params (https://github.com/juhaku/utoipa/pull/1248)
 * Use `insta` for snapshot testing (https://github.com/juhaku/utoipa/pull/1247)
 * Make `parse_named_attributes` a method of `MediaTypeAttr` (https://github.com/juhaku/utoipa/pull/1236)
 * Use a re-exported `serde_json` dependency in macros instead of implicitly requiring it as dependency in end projects (https://github.com/juhaku/utoipa/pull/1243)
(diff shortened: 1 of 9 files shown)


juhaku:
> https://swagger.io/docs/specification/v3_0/describing-parameters/#default-parameter-values gives an example on how `required: false` query parameters are described and there is no `nullable: true` or `schema.type: null` in there, which indicates that `required: false` should be sufficient to describe an `Option<_>` type for query parameters.

Although this spec is for OpenAPI 3.0 and not for OpenAPI 3.1.x it is probably right to expect that `Option<T>` query parameters should not be nullable but only `required: false`. Except in case of complex query parameters via `content` but they are not supported as of yet. https://github.com/juhaku/utoipa/issues/620

Turbo87:
> Although this spec is for OpenAPI 3.0 and not for OpenAPI 3.1.x

yeah, unfortunately I couldn't find an equivalent example in the 3.1 spec, but given the validator results in https://github.com/juhaku/utoipa/issues/1215 I guess it's a safe assumption

juhaku (approved):
Seems solid to me, :+1: Same here just a CHANGELOG.md entry probably under `Changed` category

Turbo87:
> Same here just a CHANGELOG.md entry probably under `Changed` category

done :)

juhaku (approved):
Let's set the ball rolling and merge it to master :slightly_smiling_face:

PR #1252, opened 2024-12-20
scripts/test.sh: Fix `auto_into_responses` feature declaration

Turbo87:
https://github.com/juhaku/utoipa/pull/665 renamed the feature, but for some reason the `test.sh` script contained `utoipa/` prefixes for the feature name in some cases, which effectively disabled these tests from even getting compiled.

This PR fixes the feature name in the `cargo test` invocations and "fixes" the actix tests as best as I was able to. Two tests weren't compiling at all, and another couple of tests were resulting in different output than the test was expecting. This should probably be fixed properly by someone with more knowledge of actix than me, but since it's already broken on `master` I guess we may as well merge it in this slightly less broken state 😅

### DIFF
--- a/utoipa-gen/CHANGELOG.md
+++ b/utoipa-gen/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog - utoipa-gen
 
+## Unreleased
+
+### Changed
+
+* scripts/test.sh: Fix `auto_into_responses` feature declaration (https://github.com/juhaku/utoipa/pull/1252)
+
 ## 5.3.0 - Dec 19 2024
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


juhaku:
> #665 renamed the feature, but for some reason the `test.sh` script contained `utoipa/` prefixes for the feature name in some cases, which effectively disabled these tests from even getting compiled.
> 
> This PR fixes the feature name in the `cargo test` invocations and "fixes" the actix tests as best as I was able to. Two tests weren't compiling at all, and another couple of tests were resulting in different output than the test was expecting. This should probably be fixed properly by someone with more knowledge of actix than me, but since it's already broken on `master` I guess we may as well merge it in this slightly less broken state 😅

Yup, It's quite long ago, since that feature was on the table. :sweat_smile: Thus no wonder if there are some tests that does not work as expected. So I agree we can merge this as is in slightly broken state and whenever the automatic responses will be looked again, it is more likely beneficial start fresh while keeping the original attempt as a reference. Anyhow that is something for some unknown future :smile: I believe that probably no one has ever used that feature.

utoipa-gen/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 
 ### Changed
 
+* scripts/test.sh: Fix `auto_into_responses` feature declaration (https://github.com/juhaku/utoipa/pull/1252)
```
juhaku, line 16: Otherwise great, but should be under new title `Unreleased` and subtitle `Changed` at the top before `5.3.0`

juhaku (approved):
Looks fine and dandy, the `auto_into_responses` is something that needs to be looked at in the future.

utoipa-gen/CHANGELOG.md
Turbo87: oops, I missed that this was for a released version 😅

juhaku (approved):

PR #1253, opened 2024-12-20
utoipa-gen: Replace `assert-json-diff` with snapshot testing via `insta`

Turbo87:
This PR continues what https://github.com/juhaku/utoipa/pull/1247 has started by migrating all tests in `utoipa-gen` that use `assert-json-diff` to use `insta::assert_json_snapshot!()` instead.

Note that this PR currently includes https://github.com/juhaku/utoipa/pull/1252 to avoid conflicts from the `path_derive_auto_into_responses_actix` test suite.

PR #1254, opened 2024-12-23
utoipa: Replace `assert-json-diff` with snapshot testing via `insta`

Turbo87:
- same as #1253, but for the `utoipa` crate this time

### DIFF
--- a/utoipa/src/openapi.rs
+++ b/utoipa/src/openapi.rs
@@ -656,13 +656,11 @@ pub(crate) use builder;
 
 #[cfg(test)]
 mod tests {
-    use assert_json_diff::assert_json_eq;
-    use serde_json::json;
-
     use crate::openapi::{
         info::InfoBuilder,
         path::{OperationBuilder, PathsBuilder},
     };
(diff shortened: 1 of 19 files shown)


utoipa/src/openapi.rs
```diff
@@ -795,61 +780,10 @@ mod tests {
...
-                      }
-                    }
-                  }
-                }
-            )
-        )
+
+        assert_json_snapshot!(api_1, {
+            ".paths" => insta::sorted_redaction()
+        });
```
juhaku, line 786: Does this keep the `paths` sorted always in same order?

juhaku (approved):
Super, looks good :+1:

utoipa/src/openapi.rs
Turbo87: yep. the `merge()` call caused them to be a bit random for some reason, but this fixed it.

Turbo87: https://insta.rs/docs/redactions/ has more info on this :)

Issue #1157, opened 2024-10-22
utoipa-axum: Make OpenApiRouter internal data public 


ValentinColin:
I think an easier way of accessing the router and OpenApi from OpenApiRouter would be very convenient. Either by making the internal data public, or with an accessor (get_router, get_mut_router ...). I will try to make a PR but maybe not soon.

juhaku:
Okay, sure go ahead.

PR #1262, opened 2024-12-27
Added mutable getter for the openapi instance of a OpenApiRouter

zannabianca1997:
There is currently no way of changing the `OpenApi` without breaking the router apart and recomposing it.

This enables modification of the API while mounting the routes.

I encountered this limitation while trying to build a router, and add a common security before nesting the router inside another. This will enable code like this:
```Rust
let mut router = OpenApiRouter::default()
    .routes(routes!(handler1))
    .routes(routes!(handler2));

RequireUserToken.modify(router.get_openapi_mut());

let outer_router = OpenApiRouter::default()
    .nest("/inner", router)
```

### DIFF
--- a/utoipa-axum/CHANGELOG.md
+++ b/utoipa-axum/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog - utoipa-axum
 
+## Unreleased
+
+### Changed
+
+* Added mutable getter for the openapi instance of a OpenApiRouter (https://github.com/juhaku/utoipa/pull/1262)
+
 ## 0.1.3 - Dec 19 2024
 
 ### Changed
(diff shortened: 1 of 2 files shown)


zannabianca1997:
This would also help with #1157. Most of the `Router` API is already accessible, this will enable access to the other half.

Another possibility would be to create a method that takes a type that implements `Modify`, but that would be a unnecessary restriction.

zannabianca1997:
I opened an issue, #1263 

This closes #1263

juhaku (approved):
Just missing a CHANGELOG.md entry from `utoipa-axum/CHANGELOG.md`. Add a new entry under `Unreleased` section (create if does not exists) and then add sub title `Changed` if does not exist then add the a bullet with `* title (https://PR)`

juhaku (approved):
Awesome :+1:

PR #1268, opened 2024-12-31
utoipa-axum: Add basic path params test

Turbo87:
This ensures that `/pet/{petId}` is correctly turned into a path that `axum` understands.

Note that in axum 0.8 the path syntax will change to the same syntax that OpenAPI uses, so, once we update, this test will catch the issue. I have a work-in-progress branch for the update already.


juhaku (approved):
Super, this is great :tada:

PR #1269, opened 2025-01-01
Update axum to v0.8.0

Turbo87:
This PR updates our `axum` dependencies to v0.8.0 and adjusts to the breaking changes in this release.

Specifically, it removes the `colonized_params()` fn from the router in `utoipa-axum`, since axum 0.8 now uses the same path syntax as the OpenAPI spec. It also adds a couple of `Sync` constraints.

Until `axum-test` has been updated to support the new version we won't be able to update this yet, which is why I'm opening this PR in draft mode. This PR currently also includes https://github.com/juhaku/utoipa/pull/1268 to prevent merge conflicts.

### DIFF
--- a/examples/axum-multipart/Cargo.toml
+++ b/examples/axum-multipart/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-axum = { version = "0.7", features = ["multipart"] }
+axum = { version = "0.8.0", features = ["multipart"] }
 tokio = { version = "1", features = ["full"] }
 tower = "0.5"
 utoipa = { version = "5", features = ["axum_extras"] }
--- a/examples/axum-multipart/Cargo.toml
+++ b/examples/axum-multipart/Cargo.toml
@@ -7,9 +7,9 @@ edition = "2021"
 axum = { version = "0.8.0", features = ["multipart"] }
 tokio = { version = "1", features = ["full"] }
 tower = "0.5"
-utoipa = { version = "5", features = ["axum_extras"] }
-utoipa-swagger-ui = { version = "8", features = ["axum"] }
-utoipa-axum = "0.1"
+utoipa = { path = "../../utoipa", features = ["axum_extras"] }
+utoipa-swagger-ui = { path = "../../utoipa-swagger-ui", features = ["axum"] }
+utoipa-axum = { path = "../../utoipa-axum" }
 serde = { features = ["derive"], version = "1" }
 
 [workspace]
--- a/utoipa-axum/src/router.rs
+++ b/utoipa-axum/src/router.rs
@@ -11,14 +11,6 @@ use tower_layer::Layer;
 use tower_service::Service;
 
 #[inline]
-fn colonized_params<S: AsRef<str>>(path: S) -> String
-where
-    String: From<S>,
-{
-    String::from(path).replace('}', "").replace('{', ":")
-}
-
-#[inline]
 fn path_template<S: AsRef<str>>(path: S) -> String {
     path.as_ref()
         .split('/')
@@ -245,11 +237,11 @@ where
             };
             let path = if path.is_empty() { "/" } else { path };
 
-            self.0.route(&colonized_params(path), method_router)
+            self.0.route(path, method_router)
         } else {
             paths.paths.iter().fold(self.0, |this, (path, _)| {
                 let path = if path.is_empty() { "/" } else { path };
-                this.route(&colonized_params(path), method_router.clone())
+                this.route(path, method_router.clone())
             })
         };
 
@@ -273,7 +265,7 @@ where
 
     /// Pass through method for [`axum::Router<S>::route`].
     pub fn route(self, path: &str, method_router: MethodRouter<S>) -> Self {
-        Self(self.0.route(&colonized_params(path), method_router), self.1)
+        Self(self.0.route(path, method_router), self.1)
     }
 
     /// Pass through method for [`axum::Router::route_layer`].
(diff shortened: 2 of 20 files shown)


examples/axum-multipart/Cargo.toml
```diff
@@ -4,12 +4,12 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-axum = { version = "0.7", features = ["multipart"] }
+axum = { version = "0.8.0", features = ["multipart"] }
```
Snapstromegon, line 7: I think leaving this at 0.8 instead of 0.8.0 is better. As far as I'm aware axum doesn't break on patch increases and there's already 0.8.1 out. This is not only for this case, but all axum versions in this PR.

Turbo87: `0.8` and `0.8.0` are equivalent (see https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html)

utoipa-axum/src/router.rs
```diff
@@ -10,14 +10,6 @@ use axum::Router;
...
-#[inline]
-fn colonized_params<S: AsRef<str>>(path: S) -> String
-where
-    String: From<S>,
-{
-    String::from(path).replace('}', "").replace('{', ":")
-}
-
 #[inline]
 fn path_template<S: AsRef<str>>(path: S) -> String {
```
juhaku, line 14: We can most likely remove this `path_template` fn as well since it is used to counter the colonized syntax and translate colonized paths to path template syntax, but since axum 0.8 comes with path template syntax this because redundant :thinking:

examples/axum-multipart/Cargo.toml
juhaku: Yeah, as the default resolution style is with the ^ which indicates it allows versions between `0.8.0 - 0.8.x`. If it was declared with = it would explicitly state the specific version. Though above, this is just matter of style, I deem to remove the unnecessary zeros but the effect is same.

juhaku (changes requested):
Looks quite nice. Thanks for putting the effort. :slightly_smiling_face: 

Still missing the famous `CHANGELOG.md` entries :wink: 

I don't have the time to put much effort into this project at the moment but perhaps when I get my work sorted out I have more time to focus on this one as well. :disappointed:

Turbo87:
feedback should be addressed and CI is green now :)

examples/axum-multipart/Cargo.toml
Snapstromegon: THX, TIL.
Either way, it probably should already be the newest one.

Turbo87: 0.8.1 was only a small readme change, so probably not worth the extra constraint :)

juhaku: Yup, if you do a clean build with it, it will and should pull the latest available version from crates :slightly_smiling_face:

examples/axum-multipart/Cargo.toml
```diff
@@ -4,12 +4,12 @@ version = "0.1.0"
...
 tokio = { version = "1", features = ["full"] }
 tower = "0.5"
-utoipa = { version = "5", features = ["axum_extras"] }
-utoipa-swagger-ui = { version = "8", features = ["axum"] }
-utoipa-axum = "0.1"
+utoipa = { path = "../../utoipa", features = ["axum_extras"] }
+utoipa-swagger-ui = { path = "../../utoipa-swagger-ui", features = ["axum"] }
+utoipa-axum = { path = "../../utoipa-axum" }
```
juhaku, line 12: :+1:

juhaku (approved):
Seems good, probably should copy paste the `CHANGELOG.md entry from `utoipa-swagger-ui` to `utoipa-redoc` and `utoipa-scalar` and `utoipa-rapidoc` as well.

juhaku (approved):

juhaku:
Great, thanks for the hard work :+1:

examples/axum-multipart/Cargo.toml
Snapstromegon, line 12: Maybe I'm wrong here an the docs I found are outdated, but I think this will create a problem during publishing:

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#local-paths-in-published-crates

This should probably include the path AND the version from crates.io.

juhaku: > Note: [crates.io](https://crates.io/) does not allow packages to be published with dependencies on code outside of [crates.io](https://crates.io/), except for [dev-dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies). See the [Multiple locations](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations) section for a fallback alternative for git and path dependencies.

Yes that is a case when you use the local path dependencies within crates that are supposed to be released. However this is an example crate which is not released thus not creating issues there. 

This path change does not affect releasability of the referenced crates, allowing them to be released as normal but only prohibiting release of the crate using the references without version or git ref.

Snapstromegon: Oh, I missed which Cargo.toml this was in - sorry. Thanks for clarifying.

Turbo87:
@juhaku any chance we could get this released in the next couple of days? happy to help, if I can :)

juhaku:
> @juhaku any chance we could get this released in the next couple of days? happy to help, if I can :)

Sure thing, there is not much to do actually, as I got fixed one bug that was caused by changes introduced to generic schemas behavior. Only thing to do is the define the next versions and update the CHANGELOGS with them and update Cargo.toml files with new versions. After pushing the CI does the rest (after pressing the publish button in releases) :slightly_smiling_face:

PR #1270, opened 2025-01-01
Remove `axum-test` dependency

Turbo87:
The `axum-test` crate has not yet been updated to work with axum v0.8 and is blocking us from releasing support for it ourselves. Since we use very little functionality from it, we can just replace it with direct calls to the `ServiceExt` trait of `tower`.

This should unblock https://github.com/juhaku/utoipa/pull/1269

### DIFF
--- a/utoipa-swagger-ui/Cargo.toml
+++ b/utoipa-swagger-ui/Cargo.toml
@@ -37,9 +37,10 @@ serde_json = { version = "1.0" }
 base64 = { version = "0.22.1" }
 
 [dev-dependencies]
-axum-test = "16.2.0"
+http = "1.2"
 similar = "2.5"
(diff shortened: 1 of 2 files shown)


juhaku (approved):
Great, :1st_place_medal: Thanks for the patch

juhaku (commented):
Perhaps this could also be mentioned in CHANGELOG.md as well, :thinking:

Turbo87:
it's only used internally though, so not sure how useful it is for users of the library to know 😅

juhaku:
Yup, valid point, wondering the same, it does not change how the lib is used in anyways, and if they want to see they have access to full changelog anyway from the release, so they can find it, it is just bit more effort.

juhaku:
Of we go with the merge.  :smile:

PR #1283, opened 2025-01-10
Exclude snapshots from published packages

### LINKED ISSUE #1281: Published versions of `utoipa` and `utoipa-gen` contain insta snapshot files
The published versions of `utoipa` (5.3.1 at the time of writing) and `utoipa-gen` (5.3.1 as well) contain insta snapshot files. These files are not required to use these crates, but they are only required to run tests. As it is exceptionally rare that people run tests on crates downloaded from crates.io (cargo not even provides a command for that) these files only increase the crate size and therefore "waste" storage space. It would be great to exclude these files from future published versions. See the cargo documentation on how to exclude/include files explicitly: https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields

* https://diff.rs/utoipa-gen/4.3.1/5.3.1 (navigate to src/snapshots)
* https://diff.rs/utoipa/4.2.3/5.3.1 (navigate to src/snapshots and src/openapi/snapshots)

weiznich:
This commit excludes insta snapshot files from the published packages. I've chosen to go with an explicit list of included files instead of an exclude list as this also excludes other possibly files present by accident at the publish time.

Fixes #1281

### DIFF
--- a/utoipa-gen/Cargo.toml
+++ b/utoipa-gen/Cargo.toml
@@ -9,6 +9,14 @@ keywords = ["openapi", "codegen", "proc-macro", "documentation", "compile-time"]
 repository = "https://github.com/juhaku/utoipa"
 authors = ["Juha Kukkonen <juha7kukkonen@gmail.com>"]
 rust-version.workspace = true
+include = [
+    "README.md",
+    "src/**/*.rs",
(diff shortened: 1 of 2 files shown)


juhaku (approved):
Thanks :+1:

PR #1291, opened 2025-01-15
Add HTTP method macros to simplify path attribute macro

mokeyish:
- Added `get`, `post`, `put`, `delete`, `options`, `head`, `patch`, `trace`, and `any` macros
  to simplify the `path` attribute configuration.
- These macros allow developers to define routes more concisely and improve code readability.


mokeyish:
@juhaku Hi, Do you have any comments on this PR?


I want to maintain a consistent style in functions through this simplification, although there is currently no intermediate adapter that can keep the function definition unchanged and quickly switch web frameworks, such as axum to artix.

PR #1293, opened 2025-01-15
Support servers in path macro

infiniteregrets:
Hi! Just want to say thanks for developing this amazing library. I realized that this piece was missing as per these docs https://swagger.io/docs/specification/v3_0/api-host-and-base-path/ and is my use case to be able to override the servers. 
I see it is available in the corresponding trait but not the macro, so adding it here.

~Very much a WIP, will test and add a few more commits later~

### DIFF
--- a/utoipa-gen/src/server.rs
+++ b/utoipa-gen/src/server.rs
@@ -1,0 +1,155 @@
+use proc_macro2::TokenStream;
+use quote::quote;
+use quote::ToTokens;
+use syn::Token;
+use syn::{
+    parenthesized,
+    parse::{Parse, ParseStream},
+    punctuated::Punctuated,
+    token::Comma,
+    Error, Ident, LitStr,
+};
+
+use crate::{parse_utils, Array};
+
+// (url = "http:://url", description = "description", variables(...))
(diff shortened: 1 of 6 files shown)


utoipa-gen/src/server.rs
```diff
@@ -0,0 +1,155 @@
...
+        parenthesized!(server_stream in input);
+        let mut server = Server::default();
+        while !server_stream.is_empty() {
+            let ident = server_stream.parse::<Ident>()?;
+            let attribute_name = &*ident.to_string();
+
+            match attribute_name {
+                "url" => {
+                    server.url = parse_utils::parse_next_literal_str_or_expr(&server_stream)?
+                }
```
infiniteregrets, line 36: rest of the impl is copied, I think URLs can be common amongst endpoints,  so changing it to `LitStrOrExpr` just like how tags work!

juhaku: Yup, that's fine

juhaku (approved):
Great work, only missing CHANGELOG entry, but I can also add that, and this is good to go. :+1:

Issue #1260, opened 2024-12-26
The generated spec for an Option field is unpleasant and breaks swift-openapi-generator


grahamc:
The spec that is generated today is:

```yaml
    AuthenticationStatus:
      type: object
      properties:
        token:
          oneOf:
            - type: 'null'
            - "$ref": "#/components/schemas/TokenStatus"
```

Since the token is both `null`able and also not required, the swagger UI double-indicates this in an ugly way, by showing no `*` to say it isn't required, and also showing oneOf null / the object:

![CleanShot 2024-12-26 at 15 15 39@2x](https://github.com/user-attachments/assets/d12661fc-3651-45da-b0ce-be054c1a9e1f)

and in the sample output, it decides to show the `null` option instead a usefully populated example:

![CleanShot 2024-12-26 at 15 16 31@2x](https://github.com/user-attachments/assets/511f31b6-50d1-4137-a16d-ca5d65878499)

and further, the swift openai generator pukes:

> warning: Schema "null" is not supported, reason: "schema type", skipping [context: foundIn=Components.Schemas.AuthenticationStatus (#/components/schemas/AuthenticationStatus)/token]

but this all seems a bit silly anyway, since the field is not marked as required, so we can skip the oneOf null entirely and get better docs, and better examples.

With:

```yaml
    AuthenticationStatus:
      type: object
      required:
      properties:
        token:
          "$ref": "#/components/schemas/TokenStatus"
```

![CleanShot 2024-12-26 at 15 18 00@2x](https://github.com/user-attachments/assets/c84171c3-bff1-4833-923d-60bcdb39d586)

![CleanShot 2024-12-26 at 15 17 51@2x](https://github.com/user-attachments/assets/6522a600-a971-447b-ae11-a3602f66782d)

and finally, the swift generator handles it just fine.

juhaku:
Technically the `oneOf` with `type: null` is correct. Though not sure what OpenAPI level the swift OpenAPI generator does aim for. `type: null` does not exist in OpenAPI 3.0 but it is added in OpenAPI 3.1. It combined with `oneOf` should allow defining explicitly `null` as a value or a type itself. The `required` attribute is different from the nullable type. Required controls whether the type can be absent (or wants to control). In Rust the explicit `null` will translate to `Option::None` or an unit type. 

[...]

PR #1299, opened 2025-01-23
Emit nullable_item last for OneOfBuilder

frederikrothenberger:
This PR changes the order in which the `#nullable_item` is emitted when processing the `OneOfBuilder`. This leads to much better example values in the Swagger UI, as by default the example just picks the first one of the options.

This change is inspired by this comment: https://github.com/juhaku/utoipa/issues/1260#issuecomment-2568216834
> What we could at least do is to change the order of the types in oneOf to make the null variant below, which will help the Swagger UI to use the correct spec as an example because it chooses the first one from the list.

Please take a look!


frederikrothenberger:
@juhaku: Any feedback on this?

juhaku (approved):
Great, let's get this merged.

PR #1306, opened 2025-02-03
modify operation-id default gen rule

qyvip:
Modify the rule for generating operation-id to default to pathw_ith_comtext_math, and automatically append method as a suffix when generating openAPI documents to address the issue of supporting multiple request methods simultaneously, such as post, get,head， Ensure the uniqueness of operationId and solve the problem of OpenAPI Generator generating prompt operationId is repeated


CommanderStorm (changes requested):
> address the issue of supporting multiple request methods simultaneously, such as post, get,head， Ensure the uniqueness of operationId and solve the problem of OpenAPI Generator generating prompt operationId is repeated

Can you attach a testcase for this?
Seems important to not accidentally regresss on this.

I also don't entirely understand all the changes you did and why.
Inline comments where you tell why this needed to change would really help

Issue #1168, opened 2024-10-27
Replace `serde_yaml` with `serde_yml` 


DenuxPlays:
The [serde_yaml](https://github.com/dtolnay/serde-yaml) crate is archived and will not be updated.
There is a fork however named [serde_yml](https://github.com/sebastienrousseau/serde_yml).

It should be replace it with the actively maintained fork.

juhaku:
Yeah, it probably is a good idea to replace the `serde_yaml` with `serde_yml` as it seems to take the initiative to continue the legacy of `serde_yaml` in yaml support for rust.

domenicquirl:
While I appreciate the intention here, there are some concerns around `serde_yml` FYI: https://github.com/rustsec/advisory-db/issues/2212. I know the deprecation of the original `serde_yaml` was a hard-hitter, but at least there was pretty high confidence for the original crate to be well-maintained and, most importantly, sound. 

I'd strongly consider looking for a different alternative (there are other forks of the original, like https://github.com/acatton/serde-yaml-ng for example).

DenuxPlays:
Yeah I agree.

I am sorry I wasn't aware of this.

Although I don't think that `serde-yaml-ng` is a good alternative as it seems to be unmaintained or atleast the last real commit was 5 months ago.

But it should be better to use this then serde_yml

Again I am sorry for my poor research :(

DenuxPlays:
I've found this replacement: https://crates.io/crates/serde_norway

Looks promising

domenicquirl:
Sounds good to me!

And no need to apologize, it's still the most downloaded fork, I get that it looks very appealing (probably even more so then).

PR #1311, opened 2025-02-07
replaced `serde_yml` with `serde_norway`

DenuxPlays:
@domenicquirl has mentioned that `serde_yml` is actually a  very bad replacement for `serde_yaml`.

I've chosen `serde_norway` as the replacement.
A fork of the original `serde_yaml` and way better.

For more information see the discussion here: https://github.com/juhaku/utoipa/issues/1168#issuecomment-2642797978

I am very sorry for introducing this dependency.
This pr should fix my error.

I've `serde_norway` is not wanted we could also go with https://github.com/acatton/serde-yaml-ng


juhaku (approved):
:+1:

PR #1313, opened 2025-02-13
fix(todo-actix): examples

crassicus:
According to the documentation, the /api/todo endpoint should return a 201 status when a new resource is created. Fixed the response handling to comply with this expectation.


juhaku (approved):
Thanks :+1:

PR #1328, opened 2025-03-02
Remove serde(skip_deserializing) from being skipped

### LINKED ISSUE #868: Keep serde skipped field in API Spec
Say I have a struct like
```rust
#[derive(Debug, Serialize, Deserialize, Clone, ToSchema)]
pub struct Client {
    #[serde(skip_deserializing)]
    pub test: Option<bool>,
}
```
the field `test` is being ignored because of the serde `skip_deserializing` macro, is there a way to keep it in the spec?

LynithDev:
This should resolve https://github.com/juhaku/utoipa/issues/868.

If a field has `skip_deserializing`, it should still be serialized. And all serialized fields should be in the generated OpenAPI document.

### DIFF
--- a/utoipa-gen/src/component/serde.rs
+++ b/utoipa-gen/src/component/serde.rs
@@ -47,8 +47,7 @@ impl SerdeValue {
                 match tt {
                     TokenTree::Ident(ident)
                         if ident == "skip"
-                            || ident == "skip_serializing"
-                            || ident == "skip_deserializing" =>
+                            || ident == "skip_serializing" =>
                     {
                         value.skip = true
                     }


brocaar:
I'm not sure if this 100% solves the issue?

In my use-case, there are some fields that get set server-side, e.g. `created_at` and `updated_at`. This means that I will have `#[serde(skip_deserializing)]` for these fields, as the server will set these on create / update. So on `POST` / `PUT` / `PATCH` I don't want to have these fields in my schema, but on `GET` I want because I would like to know when each object was created and last updated.

With this change, on `skip_deserializing`, these fields are always included. Using an OpenAPI client (eg. converting the API definition to Typescript), it might even mean I will be forced to enter a value for these fields or I will get a compile-time error that these fields are not present in the payload.

**Update:** I think what is needed is knowing if the object is part of the request or response: 

* deserialize request + `skip_deserializing` = exclude
* serialize response + `skip_deserializing` = include
* deserialze request + `skip_serializing` = include
* serialize response + `skip_serializing` = exclude

Issue #1234, opened 2024-12-12
Update axum examples with state


happysalada:
Could you update the axum code examples of the nested routers to include state ?
As soon as i have a handler that requires state that us nested, thw compiler infers that my router types are stateful and i can never get it to a Router<()> type later to serve it with axum.

Kimishu:
> Could you update the axum code examples of the nested routers to include state ? As soon as i have a handler that requires state that us nested, thw compiler infers that my router types are stateful and i can never get it to a Router<()> type later to serve it with axum.

If I understood ur question correctly, I think u should do something like [this](https://stackoverflow.com/questions/79247308/deadpool-postgres-connection). The main thing is to add a generic parameter to Router.

litcc:
I'm not sure if the utoipa_axum::routes! macro can handle the following scenarios:
Handler functions that substitute generic parameters

![image](https://github.com/user-attachments/assets/4fc832e5-2a0e-4620-9e2d-1b594002f157)


Includes middleware added to handler functions when writing routes

![image](https://github.com/user-attachments/assets/5df8dc01-70cc-4ba3-b34d-4a2b65ed9328)


Previous attempts to use utoipa_axum::routes! , have invariably failed


By the way I would like to make a point that is not related to this issue, please don't be offended:



😄 If possible, please add a little bit more examples of generalizations;
I've noticed that the current support for generalization doesn't seem to be particularly well developed (probably because I don't know enough about the project)

I have a basic response body based on generics, and currently only the `ToSchema` macro can be used directly, the rest of the code generated by `utoipa::IntoResponses, utoipa::ToResponse` have generic constraints and can't be used directly.

![image](https://github.com/user-attachments/assets/410a62f0-cb08-4df5-b2d6-30271b27ea61)

RobbieMcKinstry:
I'm currently encountering the same problem as @litcc : substituting generic parameters in handlers signatures.

For example, in this code, my `create_workspace_handler` handler takes a generic. In instantiate it with a concrete `PostgresRepository` instance in production, but swap that for a `MockRepository` when testing. This lets me practice the hexagonal architecture as described in [this guide](https://www.howtocodeit.com/articles/master-hexagonal-architecture-rust#the-repository-pattern-in-rust).

```rust
pub struct WorkspacesController;

impl WorkspacesController {
    pub fn router() -> OpenApiRouter<Deps> {
        OpenApiRouter::new()
            .route(
                "/",
[...]

PR #1329, opened 2025-03-03
add support for generic handlers in the routes macro

tlowerison:
code is a little sloppy but this provides support for generic handlers

addresses #1234

### DIFF
--- a/utoipa-axum/src/lib.rs
+++ b/utoipa-axum/src/lib.rs
@@ -123,30 +123,96 @@ pub use paste::paste;
 /// ```
 #[macro_export]
 macro_rules! routes {
-    ( $handler:path $(, $tail:path)* $(,)? ) => {
+    ( $( $gen:ty | )? @code $schemas:tt: $router:ident: $paths:ident: $( $handler:tt )* ) => {
         {
-            use $crate::PathItemExt;
+            let (path, item, types) = $crate::routes!(@resolve_types $($handler)* : $schemas );
+            let router = types.iter().by_ref().fold($router, |router, path_type| {
+                router.on(path_type.to_method_filter(), $($handler)* $( :: < $gen > )? )
+            });
+            $paths.add_path_operation(&path, types, item);
+            ($schemas, $paths, router)
+        }
+    };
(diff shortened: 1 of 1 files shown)


tlowerison:
Hi @litcc, I think there's a bug in your test code which is preventing compilation of my updated `routes` macro

For this section here near the end of your test code:

```rs
.routes(routes2!(browse_store::<String, HashMap<String,String>>.layer(
    TestLayer {
        param1: "a".to_string(),
        param2: Arc::new("b".to_string()),
    }
)))
```
should this code actually be the code below?
```rs
.routes(routes!(browse_store::<String, HashMap<String, String>>))
.layer(TestLayer {
    param1: "a".to_string(),
    param2: Arc::new("b".to_string()),
})
```
If so, this code block compiles using the changes in my revision

Let me know if you have any other tests you'd like to add or aspects of this you would change!

tlowerison:
@juhaku any interest in merging this?

litcc:
You're right, my understanding of the relationship between routing and middleware in axum was incorrect;
the middleware added by .layer only applies to the previous service, not all preceding services,
so there's no need to place .layer inside routes!;
thanks for the reminder

wer1st:
@juhaku please merge this PR if you have no concerns here

litcc:
> You're right, my understanding of the relationship between routing and middleware in axum was incorrect; the middleware added by .layer only applies to the previous service, not all preceding services, so there's no need to place .layer inside routes!; thanks for the reminder

There might still be a slight issue; I'm not sure where I went wrong.  
.layer applies to all previously defined routes,  
and the same goes for .router_layer. In this case,  
if I have multiple routes with different middleware parameter requirements,  

For example: Each endpoint has a fixed permission identifier, such as `system:config:read`. I have a pre-written middleware for permission validation to check whether the user has the permission to call that endpoint. This middleware has a parameter to control the permission identifier for each endpoint.  
If used as you suggested,
image

image

When calling /3, the middleware for /4 is still triggered,

So I believe this type of middleware should be placed on the handler_fn
image

Only when placed after the handler function does it become middleware for a single functionality.  
I'm not sure if my understanding is correct, or if there are points I haven't considered.

PR #1334, opened 2025-03-07
Add support for NonZeroU8, NonZeroU16, NonZeroU32, and NonZeroU64

mcronce:
These output integers matching their primitive counterparts, but with `minimum` set to 1 instead of 0


mcronce:
Realized I'm incorrectly producing `"format": "int32"` for `NonZeroU8` and `NonZeroU16` when `non_strict_integers` is enabled; need to fix that before this is ready

mcronce:
OK, that wasn't the problem.  Fixed up the issues...testing still fails with `--features=non_strict_integers`, but the failures I'm seeing existed prior to my commit.  I matched the new test cases with their primitive counterparts because I _assume_ the expected output for the primitives is what's desired.  I can change them if that's incorrect, and take a whack at fixing the failing test cases if needed.

mcronce:
I _feel like_ there was a reason I didn't include `NonZeroUsize` but I can't for the life of me remember what it was.  I'll add it and if no reason jumps out at me in the process, push it up to the branch

PR #1343, opened 2025-03-24
fix: mismatch error from InvalidArchive

mbukeRepo:
Fixes #1341

PR #1348, opened 2025-03-26
[swagger-ui]◘Pin zip dependency

Akanoa:
The PR fix the issue https://github.com/juhaku/utoipa/pull/1343, but doesn't pin the version.

I propose to define explicitly the zip crate version to avoid further complication.

### DIFF
--- a/utoipa-swagger-ui/Cargo.toml
+++ b/utoipa-swagger-ui/Cargo.toml
@@ -49,7 +49,7 @@ no-default-features = true
 rustdoc-args = ["--cfg", "doc_cfg"]
 
 [build-dependencies]
-zip = { version = "2", default-features = false, features = ["deflate"] }
+zip = { version = "2.5.0", default-features = false, features = ["deflate"] }
 regex = "1.7"
 
 # used by cache feature


anietieasuquo:
I was just about to create a PR for this before I saw this one 👍

LockedThread (approved):
LGTM, working for me, same as my fork.

juhaku:
Note this still does not exactly pin the version since the Rust default version definition resolves to  semver ^ meaning with a version like `2.5.0` it still allows `2.5.x` versions to be used when available. Though this might be exactly what is wanted. Loosely pinning like this will definitely cause errors for users who are using an earlier version of zip than 2.5. It will break their code where older zip dependency might be passed down from another dependency which has not been updated to the latest for one reason or another. 

If pinning will be done it might be better to add 2 new feature flags where one can optionally choose to use the latest 2.5.x or then 2.4.x or perhaps define the semver so that anything before 2.5 can be used if possible.

Akanoa:
oh! I didn't know about this implicit caret behavior, I thought I was defining an exact version. Thanks to let me know it :) 

you're right, someone with the zip = "=2.4.0" in its Cargo.toml will face a conflict version error .

```
error: failed to select a version for `zip`.
    ... required by package `utoipa-swagger-ui v9.0.0 (https://github.com/Akanoa/utoipa.git?branch=pin-zip-dependency#74fe10f8)`
    ... which satisfies git dependency `utoipa-swagger-ui` (locked to 9.0.0) of package `oas v0.1.0 (/home/noa/Documents/projects/rust/oas)`
versions that meet the requirements `^2.5.0` are: 2.5.0

all possible versions conflict with previously selected packages.

  previously selected package `zip v2.4.0`
    ... which satisfies dependency `zip = "=2.4.0"` of package `oas v0.1.0 (/home/noa/Documents/projects/rust/oas)`

failed to select a version for `zip` which could resolve this conflict
```
I'll submit another submission based on your feature flags proposition.

Thanks to the careful review ^^

juhaku:
Yup, thanks

juhaku:
It might be worth scraping this for now because there has been another update to zip with more breaking changes and to keep things simple utoipa is now loosely pinned to the 2.6 version.

Akanoa:
They started to become angry toward

https://github.com/zip-rs/zip2/issues/333

Akanoa:
wait what ??
https://github.com/zip-rs/zip2/issues/332

PR #1363, opened 2025-04-04
fix(swagger-ui): fix build for breaking zip crate 2.6.0 change

tversteeg:
This fixes the build for the latest zip crate breaking change, which is not behind a different major version.

The difference is that `zip::ZipFile` now takes the generic parameter `<R: Read>` which it didn't before.


juhaku (approved):
Super, thanks: :tada: Just cant comprehend what are they doing with the `zip` crate. :exploding_head:

PR #1364, opened 2025-04-04
feat: optimize enum schema generation by grouping unit variants

domenkozar:
This is just a prototype I've put together to see if I can unblock myself.

The general idea is that even if one of the enums is special, we generate the rest of enums as one.


domenkozar:
I think it's better to solve this using simplifier pass: https://github.com/OpenAPITools/openapi-generator/pull/21041

Issue #1307, opened 2025-02-04
Add support DateTimeWithTimeZone, DateTimeUtc from sea-orm


anhnmt:
I had some problems using DateTimeWithTimeZone or DateTimeUtc from sea-orm

https://github.com/SeaQL/sea-orm/blob/e0679785ba4aff93b60e91e2cf20196a41d37570/src/entity/prelude.rs#L34-L43

My struct is generated from sea-orm-cli

```
#[derive(
    Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize, utoipa :: ToSchema,
)]
#[sea_orm(table_name = "users")]
pub struct Model {
    pub created_at: DateTimeWithTimeZone,
    pub updated_at: DateTimeWithTimeZone,
    #[sea_orm(primary_key, auto_increment = false)]
    pub id: Uuid,
    pub name: String,
    #[sea_orm(unique)]
    pub email: String,
    pub password: String,
    pub role: UserRole,
    pub deleted_at: Option<DateTimeWithTimeZone>,
}
```

nyvid:
@anhnmt did you tried to use the `chrono` feature from the `utoipa` crate?

prateekkumarweb:
I am facing the same issue. Even when using feature `chrono`, it still gives the same error:

```
the trait `ToSchema` is not implemented for `chrono::DateTime<Utc>`
```

lu-zero:
even with the feature on, ToSchema is not really implemented for the types, so 

```rust
#[derive(ToSchema)]
struct S<T> {
  t: T
}
type DS = S<chrono::NaiveDate>;
```

would fail.

WENPIN1:
Hi all,
Is this derive macro with utoipa 5.4 chrono feature  ?

#[schema(value_type = String, format="date-time", example = "2023-10-11T10:00:00Z")]
pub updated_at: DateTime<Utc>,


Thank you very much !

-WENPIN

PR #1367, opened 2025-04-09
feat: Add support for chrono::DateTimeWithTimeZone

haitrr:
Add support for chrono::DateTimeWithTimeZone


jtwaleson:
I believe there is no `DateTimeWithTimeZone` , instead it comes from `Sea-ORM`  which has:

```
#[cfg(feature = "with-chrono")]
pub type DateTimeWithTimeZone = chrono::DateTime<chrono::FixedOffset>;
```

This patch works, but is not the proper solution. In https://github.com/Aleph-Alpha/ts-rs this does work, so I think the problem in Utoipa is due to the types not being converted down to their actual primitives, instead the type name is used for the comparison.

This also relates to https://github.com/juhaku/utoipa/issues/1307

I'm not good enough at Rust to come up with a fix for this, but we could maybe look at how ts-rs solved this.

PR #1376, opened 2025-04-21
Add support for DateTimeUtc

Ns2Kracy:


Ns2Kracy:
This is to address the fact that some libraries use DateTimeUtc rather than DateTime<Utc>

ap-1:
@juhaku any chance this can be merged?

Issue #1197, opened 2024-11-08
Nested generic parameters must be manually inlined


JMLX42:
## Problem

Follow up on #1182 / #1184

The new behavior for generic types is as follow:

> - If it's a native type, force `inline = true` except if `#[schema(inline = false)]` is specified (because if it is specified, maybe someone does actually want to make their own `u32` schema?).
> - If `#[schema(inline = ...)]` is specified, use it.
> - Otherwise, assume `#[schema(inline = true)]` (in order to keep the existing behavior).

But an edge case remains (cf [this comment](https://github.com/juhaku/utoipa/pull/1184#issuecomment-2455718288)).

Nested generics need to be inlined manually in some cases.

Here is an example:

```rust
   #[derive(ToSchema)]
   #[schema(as = path::MyType<T>)]
   struct Type<T> {
       t: T,
   }

   #[derive(ToSchema)]
   struct Person<T: Sized, P> {
       field: T,
       t: P,
   }
```

In this case:

1. `Person<String, Type<MyStruct>` will work as expected.
2. `Person<String, Type<i32>` will not work, because it will generate an invalid reference to the the `i32` schema (`"$ref": "#/components/schemas/i32"`) instead of inlining the `i32` schema.

## Workaround

To workaround case 2., `#[schema(inline)]` must be added manually on the nested generic's field(s):

```rust
   #[derive(ToSchema)]
   #[schema(as = path::MyType<T>)]
   struct Type<T> {
       #[schema(inline)]
       t: T,
   }

   #[derive(ToSchema)]
   struct Person<T: Sized, P> {
       field: T,
       #[schema(inline)]
       t: P,
   }
```

## To Do

- [ ] Find an actual fix

lu-zero:
Right now I'm having the opposite problem, the generic type is inlined even if I use `#[schema(inline = false)]` on it.

JMLX42:
> Right now I'm having the opposite problem, the generic type is inlined even if I use #[schema(inline = false)] on it.

This issue is about nested generics specifically. Can you post a complete example?

lu-zero:
I have a situation like
```rust
#[derive(ToSchema)]
struct A<T> {
    #[schema(inline = false)]
    a: Vec<B<T>> 
}

#[derive(ToSchema)]
struct B<T> {
    #[schema(inline = false)]
    b: T,
}
```

T is inlined in `b`

Tomorrow once I'm awake I'll try to give more details.

lu-zero:
[...]

PR #1381, opened 2025-04-28
feat: Do not try to inline if explicitly asked not to

lu-zero:
Make sure `inline = false` works as described in https://github.com/juhaku/utoipa/issues/1197

### DIFF
--- a/utoipa-gen/src/component.rs
+++ b/utoipa-gen/src/component.rs
@@ -1192,6 +1192,7 @@ impl ComponentSchema {
             }
             ValueType::Object => {
                 let is_inline = features.is_inline();
+                let is_not_inline = features.is_not_inline();
 
                 if type_tree.is_object() {
                     let nullable_schema_type = ComponentSchema::get_schema_type_override(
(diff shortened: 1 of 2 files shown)


utoipa-gen/src/component.rs
```diff
@@ -1192,6 +1192,7 @@ impl ComponentSchema {
             }
             ValueType::Object => {
                 let is_inline = features.is_inline();
+                let is_not_inline = features.is_not_inline();
```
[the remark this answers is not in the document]
lu-zero: Ah, sure.

utoipa-gen/src/component.rs
```diff
@@ -1306,16 +1307,18 @@ impl ComponentSchema {
...
                             path: Cow::Borrowed(&rewritten_path),
                             nullable,
                         };
-                        let index =
-                            if !schema_type.is_primitive() || type_tree.generic_type.is_none() {
-                                container.generics.get_generic_type_param_index(type_tree)
-                            } else {
-                                None
-                            };
+                        let index = if !is_not_inline
```
CommanderStorm, line 1315: Here is the line which I in particular wanted to get rid of.
The double negation which is not the same as the positive is quite confusing. Inlining would help here.

utoipa-gen/src/component.rs
```diff
@@ -1306,16 +1307,18 @@ impl ComponentSchema {
 
                         schema.to_tokens(tokens);
                     } else {
+                        // not inline or unset
```
CommanderStorm, line 1310:
```suggestion
                        // unused or explicitly set inline=false
```

CommanderStorm (changes requested):
Sorry, but the design is a bit clunky.
Let's inline this properly ^^

CommanderStorm:
Also adding a test case against regressions would be really nice

PR #1389, opened 2025-05-14
Make property_names accept also a Ref and not just a schema

lu-zero:
This would make possible to describe maps with enums as keys in a more faithful way.


CommanderStorm (changes requested):
Could you add a testcase to ensure that there are no regressions in this area?

PR #1393, opened 2025-05-20
fix: avoid cloning ApiDoc on every request for axum

twuebi:
@xugy99 identified a performance bottleneck in Lakekeeper that was causing excessive memory usage. His detailed analysis in https://github.com/lakekeeper/lakekeeper/issues/1106 revealed that clones of `utoipa::ApiDoc` were being performed for every incoming request, even when accessing unrelated endpoints.

The fix was quite straightforward, wrapping the `ApiDoc` in an `Arc` avoids cloning the value and only pays the serialization upon requests to the handler.

With this change, we could reduce memory consumption in a simple and very unscientific load test with 100 parallel connections hitting a GET on our service from 290.4MB to 51MB.

### DIFF
--- a/utoipa-swagger-ui/CHANGELOG.md
+++ b/utoipa-swagger-ui/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## Unreleased
 
-* avoid cloning ApiDoc on every request for axum (https://github.com/juhaku/utoipa/pull/1393)
+### Fixed
+
+* Avoid cloning ApiDoc on every request for axum (https://github.com/juhaku/utoipa/pull/1393)
 
 ### Changed
 


Teebor-Choka (commented):
How large is the object? Or depends on the schema?

dayvejones:
On the left are RAM usage graphs of 2 pods with utoipa (axum) enabled. On the right - with it disabled. Without changing of load.

image

dayvejones:
It's surprising that we haven't seen an elephant for so long 😁

Teebor-Choka:
That's an amazing find. Thank you.

twuebi:
Thanks, I've added the changelog entry

PR #1402, opened 2025-05-30
Add `#[automatically_derived]` to all macro-generated impls

SabrinaJewson:
This allows certain lints to be silenced. For example, the following code now compiles:

```rs
#![deny(single_use_lifetimes)]
#[derive(ToSchema)]
struct Demo<'a> {
    a: &'a str,
}
```


Huliiiiii:
This is quite annoying, currently we can only rely on module-level lint suppression.

PR #1403, opened 2025-05-31
Iterate all configured PathItem::operations

xitep:
A convenient utility to iterate all configured operations of a path item


CommanderStorm (changes requested):
Could you add a quick testcase below in `mod tests`?
This ensures that there are not regressions

CommanderStorm (approved):
(not a maintainer)
Test case looks good

Issue #1377, opened 2025-04-21
ToSchema panicked when the field is Result<(), Error>


Erisli:
Hi, not sure if it goes here.

This used to be an issue https://github.com/juhaku/utoipa/issues/448, and was implemented in https://github.com/juhaku/utoipa/pull/476.

```
#[derive(Serialize, Deserialize, Debug, Clone, Default, TS, ToSchema)]
#[ts(export)]
pub struct MyType {
    #[serde(rename = "MyField")]
    pub myField: Option<Result<(), Error>>,
}
```

It detects ```()``` as a tuple in the ```Result<>``` therefore it tried to find the tuple elements, hence the error:

``` message: TypeTree must have a path ```

My guess was they seem to be the same root cause, and I am not sure if this is intended after version 5.x.x. Any help would be appreciated, thanks in advance.

Issue #1312, opened 2025-02-12
How to support unit type in generics?


daladim:
Hello,

Thanks for utoipa. I'm very new to utoipa (and to OpenAPI) but it looks super powerful.
I'm hitting an error at derive time. I _think_ that it is due to missing support in utoipa, but I may be wrong.

How to reproduce: put this file as a binary in the `axum-utoipa-bindings` example crate and `cargo build` it.
```
use axum::Json;

#[derive(utoipa::ToSchema)]
struct Wrapper<T> {
    inner: T
}

#[utoipa::path(
    get,
    path = "/unit_type",
)]
async fn unit_type(my_param: Wrapper<Struc<()>>) -> u32 {
    0
}

#[tokio::main]
async fn main() {}
```

utoipa will fail because `TypeTree must have a path: Type::Tuple { paren_token: Paren, elems: [] }`, as if the unit type was not supported.
Change `my_param`'s type to `Json<Wrapper<String>>`, and utoipa generates code correctly.

I tried to implement the missing piece in utoipa, but I know too little about parsing/lexing in Rust, and about OpenAPI to be really helpful here :-/

Thanks

daladim:
Edit: due to a failed copy-and-paste, my minimal reproducer did not exhibit any issue. Now it is

daladim:
For those interested in the future (and since I gave up the rabbit hole of figuring out exactly what when wrong), I replaced `()` with `NoArgs` defined as followed, and this worked.

```
#[derive(utoipa::ToSchema)]
pub struct NoArgs{}
```

PR #1421, opened 2025-06-25
More helpful error message

daladim:
This change is related to https://github.com/juhaku/utoipa/issues/1312 and https://github.com/juhaku/utoipa/issues/1377.

Every so often, I'm hitting a `TypeTree must have a path` error. Usually, it can be worked around by creating a newtype that somehow is better supported by utoipa (see [an example here](https://github.com/juhaku/utoipa/issues/1312#issuecomment-2672692588)).
I'm not versed enough in proc-macros to fix this myself. However, I have been using this trick, that at least points me at what type is not "understood" by utoipa.

Example of error message before:
> TypeTree must have a path

Example of error message after:
>  utoipa does not understand this type, for which type_tree must have a path:: Type::Tuple { paren_token: Paren, elems: [Type::Path { qself: None, path: Path { leading_colon: None, segments: [PathSegment { ident: Ident { ident: "SomeCustomType", span: #0 bytes(174539..174557) }, arguments: PathArguments::None }] } }, Comma, Type::Path { qself: None, path: Path { leading_colon: None, segments: [PathSegment { ident: Ident { ident: "T", span: #0 bytes(174559..174560) }, arguments: PathArguments::None }] } }] }

### DIFF
--- a/utoipa-gen/src/component.rs
+++ b/utoipa-gen/src/component.rs
@@ -160,7 +160,7 @@ impl<'p> SynPathExt for &'p Path {
                             let path = type_tree
                                 .path
                                 .as_ref()
-                                .expect("TypeTree must have a path")
+                                .expect(&format!("utoipa does not understand this type, for which type_tree must have a path: {ty:?}"))
                                 .as_ref();
 
                             if let Some(default_type) = PrimitiveType::new(path) {


Drifter-J:
cc @juhaku I think this can be merged non-controversially

Issue #841, opened 2024-01-10
IntoParams does not respect `#[serde(flatten)]`


jelacious:
Hiya!

I am implementing some pagination for a few of my axum endpoints and am not able to properly get the Pagination fields into my structs. In my twisted mind it'd make sense for the `IntoParams` derive macro to respect `#[serde(flatten)]` to inline all fields from the flattened structure into the other.

Here's an example of what I mean using `utoipa-swagger-ui`:

![image](https://github.com/juhaku/utoipa/assets/103415686/6c213986-f51c-4f01-ad5b-d0843e19ceef)

I'd expect both `page` and `per_page` to be included into the parameters as normal queries instead of a json object.

Relevant code:

```rust
#[derive(Serialize, Deserialize, Debug, IntoParams)]
#[into_params(style = Form, parameter_in = Query)]
pub struct DeviceQuery {
    /// Device names to search for.
    #[serde(default)]
    #[param(style = Form, explode, allow_reserved)]
    name: Vec<String>,
    /// Profile names to search for.
    #[serde(default)]
    #[param(allow_reserved)]
    r#type: Option<String>,
    /// Pagination desired.
    #[serde(flatten, default)]
    #[param(inline)] // this has to be included or Pagination isn't found 
    pagination: Option<Pagination>,
}
```

```rust
#[derive(Deserialize, Serialize, Copy, Clone, Debug, ToSchema)] // should I be deriving IntoParams (doesn't do anything in my experience)?
#[serde_as]
#[serde(default)]
pub struct Pagination {
    /// Which "page" of the results desired.
    #[serde_as(as = "DisplayFromStr")]
    page: usize,
    /// Number of results per page.
    #[serde_as(as = "DisplayFromStr")]
    per_page: usize,
}

impl Default for Pagination {
    fn default() -> Self {
        Self {
            page: 0,
            per_page: 50,
        }
    }
}
```

Am I missing something. or is this working as intended?

jelacious:
Adding `IntoParams` or `ToSchema` on both has not effect. Using `#[param(inline)]` or `#[schema(inline)]` or both has no effect. I am unsure what `inline` is even supposed to do at this point.

KeeganMyers:
+1 I'm currently in the same situation as of version 4.2.0

juhaku:
@jelacious 
[...]

PR #1426, opened 2025-07-09
Support for serde(flatten) in IntoParams

litcc:
I also encountered this issue and tried to solve it: https://github.com/juhaku/utoipa/issues/841


0xJepsen:
would be nice to have this

litcc:
It's been too long, probably not needed anymore.

litcc:
Alright, currently this change can only be considered a minimal fix to solve the problem, but it might not be very elegant. If the author doesn't want to do it this way or has other ideas, please let me know.

PR #1241, opened 2024-12-16
Add support for `#[schema(pattern = "...")]` on new type structs

wyatt-herkamp:
This adds pattern for new type structs

Forgot about this issue I had a busy semester.
Closes #1150

PR #1427, opened 2025-07-12
Expand support for validation features in `#[schema(...)]` on new type structs

### LINKED ISSUE #572: Patterns for structs with single unnamed field
For the following struct,
```
struct Thing(T);
```
I would like to restrict `T` using some of the options available for structs with named fields, such as restricting a `String` to a pattern, or a numerical type to a `minimum` and `maximum`. Is there a way to specify this using the `#[schema(...)]` macro?

future-highway:
This is essentially a follow up to #1241, adding the rest of the validation related features for  newtype structs. 

Closes #572


Drifter-J:
Thanks! looks clean and the test coverage is solid 👍🏻 merged.

PR #1457, opened 2025-09-05
Implement bigdecimal (needed for diesel &etc.)

### LINKED ISSUE #1449: impl ToScheme for Diesel default types
After some research i discovered that utoipa implements decimal number with the crate rust_decimal, under the features of decimal and decimal_float, but when using [diesel](https://diesel.rs/) crate, decimal number are represented using [BigDecimal](https://docs.rs/bigdecimal/latest/bigdecimal/) crate. 
Api that interfaces with databases may use diesel to do db queries, and have to use data types not implemented for the utoipa crate, is possible to add the `ToScheme` implementation for diesel's default Types?

SamuelMarks:
Closes #1449

FWIW: Both [bigdecimal](https://github.com/akubera/bigdecimal-rs) and [rust-decimal](https://github.com/paupino/rust-decimal) were last updated 3 weeks ago, so both are maintained.


theelderbeever:
Is this not going to be fixed/added? I am running into this with bigdecimal as well.

PR #1458, opened 2025-09-06
Pass through for axum::Router::has_routes

xitep:
Expose `OpenApiRouter`'s underlying `axum::Router::has_routes` to be able to make decisions without decomposing via `OpenApiRouter::split_for_parts`.


Drifter-J:
Hi @xitep, the change looks good but there are a couple of things.
1. could you add a description
2. this MR bundles 2 unrelated changes in 1 MR, it would be nice to separate them out. 
3. plz add at least a one line assert test on pass through
4. add an entry in the changelog

Thanks

Issue #1205, opened 2024-11-15
bug: the trait bound `BigDecimal: PartialSchema` is not satisfied


francis2tm:
Hello,
Amazing project! Thank you for your work. 
I've just upgraded my utopia crate from `5.0.0-beta` to the latest version `5.2.0` and now I have a problem:

```bash
error[E0277]: the trait bound `BigDecimal: PartialSchema` is not satisfied
   --> server/src/db/file.rs:193:30
    |
193 |     pub price_amount: Option<BigDecimal>,
    |                              ^^^^^^^^^^ the trait `ComposeSchema` is not implemented for `BigDecimal`, which is required by `BigDecimal: PartialSchema`
    |
    = help: the following other types implement trait `ComposeSchema`:
              &[T]
              &mut [T]
              &str
              Account
              AccountsMembership
              BTreeMap<K, T>
              BTreeSet<K>
              BillingCustomer
            and 51 others
    = note: required for `BigDecimal` to implement `PartialSchema`
note: required by a bound in `utoipa::ToSchema::name`
   --> /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/utoipa-5.2.0/src/lib.rs:364:21
    |
364 | pub trait ToSchema: PartialSchema {
    |                     ^^^^^^^^^^^^^ required by this bound in `ToSchema::name`
...
395 |     fn name() -> Cow<'static, str> {
    |        ---- required by a bound in this associated function

For more information about this error, try `rustc --explain E0277`.
error: could not compile `server` (bin "server") due to 12 previous errors
```
In `5.0.0-beta` I didn't have this problem. 
BigDecimal is from the https://github.com/akubera/bigdecimal-rs crate, `v0.4.6`

Thanks in advance!

francis2tm:
Got it, thanks a lot! 

Cheers

PR #1487, opened 2025-10-13
feat: bigdecimal support

hdbg:
Another take on bigdecimal support.

This is mainly for the databases: to avoid creating models, where just the `BigDecimal` is changed to `rust_decimal::Decimal`


CommanderStorm (approved):

Drifter-J:
This is related to https://github.com/juhaku/utoipa/issues/1205 and https://github.com/juhaku/utoipa/issues/1449. 

I feel uneasy about this MR as we add an extra dependency which contradicts a philosophy of maintainer (@juhaku) based on his recent comment in https://github.com/juhaku/utoipa/issues/1565#issuecomment-4695453268 

Screenshot 2026-06-13 at 23 56 43

I still believe `value_type` is sufficient but it's up to @juhaku to decide whether adding a new dependency is acceptable in this case (as 2 peeps requested)

juhaku:
Well there are already 3 different time libraries supported so why not a second big decimal one. Seems like it is used quite a lot. This does not add a a runtime dependency, only for development, which is okayish. 

Though note for future, it would be worth to reassess the current hard coded types implementation whether there is another way around this which is more flexible and would need less hard coding in utoipa side. Perhaps the `utoipa-config` could have a facelift and be leveraged here as well.

juhaku (changes requested):
For this to merge, we still need CHAGELOG.md entry and the docs need to be updated including README.md 🙂

Drifter-J:
> For this to merge, we still need CHAGELOG.md entry and the docs need to be updated including README.md 🙂

I pushed the doc changes :) and FYI, the examples test is failing due to unrelated issue! @juhaku

juhaku (approved):
🏆

juhaku:
> > For this to merge, we still need CHAGELOG.md entry and the docs need to be updated including README.md 🙂
> 
> I pushed the doc changes :) and FYI, the examples test is failing due to unrelated issue! @juhaku

Yeah, I am aware of that, that is super weird as the local build works ( or I have some different set of dependencies) :thinking: Anyhow it is related to some time crate compatibility with actix-web crate. Need to investigate at some time.

Drifter-J:
> > > For this to merge, we still need CHAGELOG.md entry and the docs need to be updated including README.md 🙂
> 
> > 
> 
> > I pushed the doc changes :) and FYI, the examples test is failing due to unrelated issue! @juhaku
> 
> 
> 
> Yeah, I am aware of that, that is super weird as the local build works ( or I have some different set of dependencies) :thinking: Anyhow it is related to some time crate compatibility with actix-web crate. Need to investigate at some time.

I can have a look today if you are not looking at it any time soon.

PR #1490, opened 2025-10-20
feat: support `bytes::Bytes`

hdbg:
basic feature support for `bytes::Bytes`

### DIFF
--- a/utoipa/src/lib.rs
+++ b/utoipa/src/lib.rs
@@ -800,11 +800,14 @@ impl PartialSchema for serde_json::Value {
 impl ToSchema for serde_json::Value {}
 
 #[cfg(feature = "bytes")]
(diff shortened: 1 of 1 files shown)


utoipa/src/lib.rs
```diff
@@ -799,6 +799,16 @@ impl PartialSchema for serde_json::Value {
...
 
+#[cfg(feature = "bytes")]
+const _: () = {
+    impl PartialSchema for bytes::Bytes {
+        fn schema() -> openapi::RefOr<openapi::schema::Schema> {
+            <Vec<u8> as PartialSchema>::schema()
+        }
+    }
+    impl ToSchema for bytes::Bytes {}
+};
```
juhaku: Why wrapping it to `const _: () { } ` and not just block directly as `{ }`?

PR #1494, opened 2025-11-14
Make value field in SecurityRequirement `pub`

marlonbaeten:
In our project we use the `OpenApi` data structure to test our implementation corresponds to our documentation. However we cannot acces the the value of `SecurityRequirement` without serializing the struct.

We propose to make the value field `pub`.


CommanderStorm (changes requested):
- Can you change this to `impl Iterator` instead or would this not solve the issue at hand for you?
- Can you add a quick test so that if there are regressions, a test catches this?

PR #1496, opened 2025-11-15
Use desynt for r# stripping of ident

jayvdb:
Sits on top of #1497 

Use https://github.com/jayvdb/desynt


juhaku:
While this seems fascinating peace of work, it seems a bit overkill just to get `r#...` stripped from the struct fields. What was previously there. Though this can strip away `r#...`  not just from named struct fields but also from the types itself which itself could be useful but not sure whether in the utoipa context. 

Nonetheless as for now this need to be blocked, utoipa is still in edition 2021 and is committed to work with rust 1.75. Thus there are certain features that are not available.

PR #1500, opened 2025-11-20
ignore really ignores!

### LINKED ISSUE #1499: `ignore` doesn't really ignore!
It doesn't ignore and still care about the property!

For example, in the following case:
```rust
struct Y;

#[derive(ToSchema)]
struct X {
    #[schema(ignore)]
    field: Y
}
```

`Y` still needs to have `ToSchema` and `PartialSchema` implemented.

And I'll be able to see `Y` object in the final OpenAPI documentation.

omid:
Fixes #1499


ianthetechie:
I am not a maintainer of utoipa, but I can independently confirm that this fixes the issue :)

utoipa-gen/src/component/schema.rs
```diff
@@ -368,7 +368,15 @@ impl NamedStructSchema {
...
-                        Some(Ok((field_options, field_rules, field_name, field)))
+                        let should_always_ignore = match &field_options.ignore {
+                            Some(LitBoolOrExprPath::LitBool(bool)) => bool.value(),
+                            _ => false,
+                        };
+                        if should_always_ignore {
+                            None
+                        } else {
+                            Some(Ok((field_options, field_rules, field_name, field)))
+                        }
```
juhaku, line 379: While this seems to work when only used with `#[schema(ignore)]` syntax. it however does not work with syntax like `#[schema(ignore = my_ignore_fn)]`.

Docs: https://docs.rs/utoipa/latest/utoipa/derive.ToSchema.html#named-fields-optional-configuration-options-for-schema

Thus this implementation is broken in this state, and either needs fixing, or the current ignore definition should be changed to not to support the function call. 

Fact is that the function call expression makes this bit more spicy, because the evaluation does not happen when the macro builds but afterwards during runtime. Thus this will need some serious thought to get it right.

juhaku (changes requested):

juhaku:
I label this as breaking because this is potentially breaking if the current implementation will be changed not to support `fn ignore() -> bool` syntax. 

Nonetheless it might be to only option for this.

utoipa-gen/src/component/schema.rs
omid: Thanks @juhaku  

That's a good point.

I searched GitHub (public repositories) and just found utopia tests. I know it's not everything, but will give some estimation of the importance and dependency.

If it's runtime as you said, what's the use case? In some cases it will return ignore and when I refresh again it returns not ignore? I cannot easily imagine a use case 🤔

omid: @juhaku
Anyway, you tell me what to do? 

1. Only ignore, just when _ignore_ is a boolean
2. Remove runtime function from _ignore_
3. _something else_

Drifter-J: I agree on @juhaku here about it's architecturally not sound. I think the cleanest path is we deprecate `ignore = fn` tbh. we can emit deprecation warning when used for now. and we can move the ignore check inside `get_named_struct_field_options` function in utoipa-gen/src/component/schema.rs, the same file, before type resolution, so `#[schema(ignore)]` skips all processing.

as a long term we remove `ignore = fn` in the next major version. Wdyt?

the current PR shouldn't be merged as it is.

utoipa-gen/tests/snapshots/schema_derive_test__derive_schema_with_ignore_eq_call_field.snap.new
```diff
@@ -0,0 +1,16 @@
+---
```
Drifter-J, line 1: q. why was this pushed? with .new? (not familiar with this)

Drifter-J:
@omid Do you believe `into_params.rs` needs to be updated as well like `ToSchema` here?

utoipa-gen/tests/snapshots/schema_derive_test__derive_schema_with_ignore_eq_call_field.snap.new
juhaku: That was actually due to me doing testing, and left the snapshots in pending review state. Thanks for noticing anyhow :smile:

utoipa-gen/src/component/schema.rs
juhaku: @omid 
> If it's runtime as you said, what's the use case? In some cases it will return ignore and when I refresh again it returns not ignore? I cannot easily imagine a use case 

I can't remember exactly how it was, but the feature was most likely built before the generics where in their current state. Seems like the generics do not suit well with the ignore functionality. But the devil is in the details. It generally generates the following AST:
```rust
// ...
if !ignore {
   // the component magic happens here
}

// or if function call

if !ignore() {
  // the component magic happens here
}
````

This is exactly to support the functionality for the rust compiler to allow evaluating the schemas based on a function call. Proc-macros cannot execute arbitrary functions, all they see is a token stream representation of code. 

I am not sure why this support was initially introduced, but perhaps some people prefer to use a single function reference which returns boolean instead of repetitive boolean definition. Though I am not sure whether the current expression support allows definition of `const fn` which creates an identical AST output and will allow the compiler to optimize the ignore statement the same way as it does with a boolean value to, possibly  to completely ignore the code branch. 

But the actual root issue in the current ignore implementation is the recursive schema collection pattern. The ignore does not affect the schema collection, which then manifests as errors like **T does not implement PartialSchema** etc. while having the `#[schema(ignore)]` attribute defined over the field.

juhaku:
> @omid Do you believe `into_params.rs` needs to be updated as well like `ToSchema` here?

I'd say it would be great if the same support would also land in the `IntoParams` macro as well. But depending on whether the IntoParams macro implementation uses the named schema implementation behind the scenes it might come as a bonus for free.

Just leaving this only for the `ToSchema` does seem a little like doing half the work and then would be inconsistent with the `IntoParams` implementation what should have identical support anyways.

utoipa-gen/src/component/schema.rs
juhaku: @omid @Drifter-J 

I'd say go for:
1. Only ignore boolean support. That is the simplest and probably only way to go forward with this for now. 

I'd also look into the possibility to emit the compiler warning. At the time of implementation it was not yet landed stable compiler feature, but the compiler might have it by now. I haven't checked the situation recently. But if it's just recently added as stable it should not be adopted yet. Utoipa should support at least last 5 stable rust releases. Currently set to 1.75 thus there is plenty to catch up, perhaps we could easily support 1.90. 

Also the docs should be updated that the `fn -> bool` will be sunset and removed in next major release. 

If emitting the compiler warning is not possible for some strange reason. I guess we just need to live it that. Perhaps a println would do in such case, would require testing though. 

Eventually the old ignore functionality should be removed, but should not be done in this PR; otherwise we are forced to do a major release, and we do not want to do that lightly. Though there are know changes that need a breaking change so to take most out of it would be to group all of them together, but that is some future thing.

omid:
@juhaku @Drifter-J 

It's not straight forward to print a deprecation warning, I did a workaround. If you don't like it, we can go the normal printing way.

I updated the CHANGELOG.md of utoipa-gen crate.

And updated the docs.

Also implement it for `IntoParams`.

juhaku (approved):

juhaku:
Nice work, let's get this merged :tada:

PR #1501, opened 2025-11-22
Add tagged_discriminator feature to generate cleaner OpenAPI schemas …

### LINKED ISSUE #1456: Improve support for serde tagged enums / OneOf generation
Hi,

we are currently designing an API that can return one of multiple different structs from an endpoint.

The structs contain some shared fields, so we want to use a discriminator field so that the generated REST clients can clearly distinguish between the different types.

We also want to enforce type safety as much as possible in our rust code as well as for generated REST clients, so we want to avoid optional fields and use unions/enums instead.

However, we noticed that it is a bit difficult at the moment to use utoipa together with #[serde(tag ...)] because it creates new, unnamed objects for each enum variant, which are very tedious to use on client side. Also, the #[schema(discriminator... )] macro can only be used for untagged enums.

We first tried the following apporach:

```rust
#[derive(Debug, Deserialize, Serialize, Clone, ToSchema)]
pub struct OperationStateCompleted {
    #[schema(value_type = String, format = "date-time")]
    pub created_at: String,
    #[schema(value_type = String, format = "date-time")]
    pub completed_at: String,
}

#[derive(Debug, Deserialize, Serialize, Clone, ToSchema)]
pub struct OperationStatePending {
    #[schema(value_type = String, format = "date-time")]
    pub created_at: String,
}

/// Represents the entire state of a long-running operation.
#[derive(Debug, Deserialize, Serialize, Clone, ToSchema)]
#[serde(tag = "status")]
#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
pub enum OperationState {
    Pending(OperationStatePending),
    Completed(OperationStateCompleted),
}
```

This works fine in Rust and serde serializes/deserializes exactly what we want, but Utoipa generates the following Schema:

```yaml
OperationState:
      oneOf:
      - allOf:
        - $ref: '#/components/schemas/OperationStatePending'
        - type: object
          required:
          - status
          properties:
            status:
              type: string
              enum:
              - PENDING
      - allOf:
        - $ref: '#/components/schemas/OperationStateCompleted'
        - type: object
          required:
          - status
          properties:
            status:
              type: string
              enum:
              - COMPLETED
```

Even though serde serializes the status tag into the enum variants directly, Utoipa creates a new, anonymous object via allOf for each enum variant. This results in unusable generated clients, because instead of generating a union of (OperationStatePending | OperationStateCompleted) clients will now generate a union of (OperationStateOneOf | OperationStateOneOf1) or similar, with nested types inside.

Instead, we think that Utoipa should automatically recognize the serde(tag) and treat this field as discriminator for OpenAPI, so it should generate something like:

```yaml
OperationState:
      oneOf:
        - $ref: '#/components/schemas/OperationStatePending'
        - $ref: '#/components/schemas/OperationStateCompleted'
      discriminator:
          propertyName: status
```

This would also have to include the status member for the OperationState schemas and may also use the advanced dsicriminator syntax that maps between property value and Schema.

To solve the issue, we now had to create some suboptimal rust enums and structs to manually build the desired behavior:

```rust
#[derive(Debug, Deserialize, Serialize, Clone, ToSchema)]
#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
pub enum OperationStateCompletedDiscriminator {
    Completed,
}

#[derive(Debug, Deserialize, Serialize, Clone, ToSchema)]
#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
pub enum OperationStatePendingDiscriminator {
    Pending,
}

#[derive(Debug, Deserialize, Serialize, Clone, ToSchema)]
pub struct OperationStateCompleted {
    #[schema(inline)]
    pub status: OperationStateCompletedDiscriminator,
    #[schema(value_type = String, format = "date-time")]
    pub created_at: String,
    #[schema(value_type = String, format = "date-time")]
    pub completed_at: String,
}

#[derive(Debug, Deserialize, Serialize, Clone, ToSchema)]
pub struct OperationStatePending {
    #[schema(inline)]
    pub status: OperationStatePendingDiscriminator,
    #[schema(value_type = String, format = "date-time")]
    pub created_at: String,
}

/// Represents the entire state of a long-running operation.
#[derive(Debug, Deserialize, Serialize, Clone, ToSchema)]
#[serde(untagged)]
#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
#[schema(discriminator = "status")]
pub enum OperationState {
    Pending(OperationStatePending),
    Completed(OperationStateCompleted),
}
```

While this works, this makes the rust code very verbose and error prone. We also experimented with ideas like manually implementing ToSchema or custom derive macros, but all of these approaches only resulted in even more complexity and sometimes removed compile time guarantees.

I am not sure if we are missing a special combination of options/macros to make this possible. If not, we think this would be a great addition to Utoipa for increased type safety and comfort when using discriminators and oneOf types.

juliuslipp:
…for internally tagged enums (fixes #1456)

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ and the `ipa` is _api_ reversed. Aaand... `ipa` is also an awesome type of beer 
 - **`rc_schema`**: Add `ToSchema` support for `Arc<T>` and `Rc<T>` types. **Note!** serde `rc` feature flag must be enabled separately to allow
   serialization and deserialization of `Arc<T>` and `Rc<T>` types. See more about [serde feature flags](https://serde.rs/feature-flags.html).
 - **`config`** Enables [`utoipa-config`](./utoipa-config/README.md) for the project which allows defining global configuration options for `utoipa`.
+- **`tagged_discriminator`**: Changes the schema generation for internally tagged enums. Instead of wrapping variants in an anonymous `allOf` object to inject the discriminator tag, it generates a cleaner schema with `oneOf` and a top-level `discriminator` field. This provides better compatibility with some client generators (e.g. openapi-generator for TypeScript/Java) which struggle with the default nested structure.
 
 ### Default Library Support
 
--- a/utoipa-gen/Cargo.toml
+++ b/utoipa-gen/Cargo.toml
@@ -75,6 +75,7 @@ smallvec = []
 repr = []
 indexmap = []
 rc_schema = []
+tagged_discriminator = []
 config = ["dep:utoipa-config", "dep:once_cell"]
 
 # EXPERIEMENTAL! use with cauntion
--- a/utoipa-gen/src/component/schema/enums.rs
+++ b/utoipa-gen/src/component/schema/enums.rs
@@ -21,7 +21,7 @@ use crate::{
     },
     doc_comment::CommentAttributes,
     schema_type::SchemaType,
-    Array, AttributesExt, Diagnostics, ToTokensDiagnostics,
+    as_tokens_or_diagnostics, Array, AttributesExt, Diagnostics, ToTokensDiagnostics,
 };
 
 use super::{features, serde, NamedStructSchema, Root, UnnamedStructSchema};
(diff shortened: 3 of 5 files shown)


baughmann:
Any idea when this might be merged? Looking forward to this

PR #1503, opened 2025-11-26
Allow named field descriptions

### LINKED ISSUE #1466: Description on struct field
The documentation states that it is possible to have `description` on the field. However when trying to annotate a struct field with:
```
#[derive(Deserialize, Debug)]
#[serde(rename_all = "camelCase")]
#[derive(ToSchema)]
pub struct SomeRequest {
    #[schema(
        example = "John",
        description = "First name for the requester",
    )]
    name: String,
}

```

It results in an error, when compiling:
```
error: unexpected attribute: description, expected any of: example, examples, value_type, format, default, write_only, read_only, xml, inline, nullable, rename, multiple_of, maximum, minimum, exclusive_maximum, exclusive_minimum, max_length, min_length,
 pattern, max_items, min_items, schema_with, additional_properties, required, deprecated, content_encoding, content_media_type, ignore, no_recursion
  --> src/lib.rs:79:9
   |
79 |         description = "a game for which the optimization should be done"
```

ivan770:
Closes #1466

### DIFF
--- a/utoipa-gen/src/component/schema.rs
+++ b/utoipa-gen/src/component/schema.rs
@@ -627,7 +627,14 @@ impl NamedStructSchema {
             .as_ref()
             .map_try(|value_type| value_type.as_type_tree())?;
         let comments = CommentAttributes::from_attributes(&field.attrs);
-        let description = &ComponentDescription::CommentAttributes(&comments);
+
+        let description: Option<Description> =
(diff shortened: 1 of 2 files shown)


CommanderStorm (changes requested):
THanks for the contribution.
Can you add a testcase for your code to prevent regressions?

PR #1509, opened 2026-01-05
Add `const` parameter to Object

paolobarbolini:


CommanderStorm (changes requested):
Thank you for the contribution.
Could you attach a testcase so that we know that we don't regress on this functiality? 😉

PR #1511, opened 2026-01-08
Add `#[schema(title_variants)]` attribute for mixed enums

### LINKED ISSUE #865: [Feature Request] Automatically derive "title" for Enum Variants
I'm adding functionality to another code generator that consumes the OpenAPI spec, and I realized that it would be nice to propagate the enum names from Rust into other languages. I'm probably going to do this by pulling the `title` field if present.

In my environment, I have a lot of enums and adding `schema(title="X")` for each variant just to have the same name is tedious. Is there any reason not to automatically set `title=<enum name>` automatically?

leovalais:
closes #865 

Hi and thanks for utoipa!

Adds an enum-level attribute to generate the `title` of its variants. By default it is the concatenation of the enum's and the variant's name. If the enum has an explicit `title` set, it is used as prefix instead. All variant titles can be overridden using the field-level `title` attribute.

Like the issue author I've encountered some issues with code generators on rich enums. I tried to keep the new feature as simple as possible but let me know if there's missing functionality.

I'm also not sure how (or even if) it should relate to `serde` attributes.


leovalais:
Hi @juhaku, any chance of seeing this merged? I'd be happy to work on it more if something's not right. Or just to know if the PR should be closed because it's not a right fit for utoipa. Thanks!

PR #1514, opened 2026-01-26
Added the vendored version of scalar

vaytea:
now the scalar shouldn't need the internet connection to load :))


ghost:
this will be a nice upgrade over swagger considering how easy it is for scalar ui to configure header authorization. good luck! @vaytea

CommanderStorm (changes requested):
can you add the ability to update this arefact?

Here is how swagger does it and this way it is clear what this blob on disk is:

https://github.com/juhaku/utoipa/blob/2492086d40ad2b488b00db39724d45a92cb7863a/justfile#L83-L105

vaytea:
sorry about the delay,
I live in Iran with almost no internet

PR #1515, opened 2026-01-27
Add customizable page title to `utoipa-scalar`

terror:
This diff adds a `title` method to Scalar that allows customizing the HTML page title. Previously, the title was hardcoded to "Scalar" and the only way to change it was to override the entire HTML template with `custom_html()`.

The default HTML template now uses a `$title` placeholder which is replaced alongside `$spec` during `to_html`. The default title remains "Scalar" for backwards compatibility.

### DIFF
--- a/utoipa-scalar/CHANAGELOG.md
+++ b/utoipa-scalar/CHANAGELOG.md
@@ -1,5 +1,11 @@
 # Changelog - utoipa-scalar
 
+## Unreleased
+
+### Added
+
+* Add `Scalar::title` for customizing the HTML page title (https://github.com/juhaku/utoipa/pull/1515)
+
 ## 0.3.0 - Thu 16 2025
 
 ### Changed
@@ -45,4 +51,3 @@
 
 * Update docs and next versions
 * Use same licences for scalar as well
-
--- a/utoipa-scalar/res/scalar.html
+++ b/utoipa-scalar/res/scalar.html
@@ -1,7 +1,7 @@
 <!doctype html>
 <html>
 <head>
-    <title>Scalar</title>
+    <title>$title</title>
     <meta charset="utf-8"/>
     <meta
             name="viewport"
--- a/utoipa-scalar/src/lib.rs
+++ b/utoipa-scalar/src/lib.rs
@@ -239,14 +239,12 @@ impl<S: Spec> Scalar<S> {
     /// At this point in time, it is not possible to customize the HTML template used by the
     /// [`Scalar`] instance.
     pub fn to_html(&self) -> String {
-        self.html
-            .replace("$title", &self.title)
-            .replace(
-                "$spec",
-                &serde_json::to_string(&self.openapi).expect(
-                    "Invalid OpenAPI spec, expected OpenApi, String, &str or serde_json::Value",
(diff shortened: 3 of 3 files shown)


CommanderStorm (approved):

MarkusZoppelt (approved):
Looks like a clean, nice UX win for a popular UI. Can we merge this?

terror:
> Looks like a clean, nice UX win for a popular UI. Can we merge this?

cc @juhaku 

Would love to hear your thoughts on this :)

juhaku (approved):
Looks good. Thanks for PR. 💯

juhaku:
@terror Though missing `CHANGELOG.md` entry. That should be added. Should follow the same convention as already exists, if there is no `Unreleased` title in the beginning of the file, then new one should be added.

juhaku (approved):
🎉

juhaku:
Great :+1:

terror:
> @terror Though missing `CHANGELOG.md` entry. That should be added. Should follow the same convention as already exists, if there is no `Unreleased` title in the beginning of the file, then new one should be added.

Ah, for sure. I just added one. It looks like theres a typo in the existing changelog filename, and its missing from the root file as well 🫠

juhaku:
> > @terror Though missing `CHANGELOG.md` entry. That should be added. Should follow the same convention as already exists, if there is no `Unreleased` title in the beginning of the file, then new one should be added.
> 
> Ah, for sure. I just added one. It looks like theres a typo in the existing changelog filename, and its missing from the root file as well 🫠

Oh, well, I can fix the typos :slightly_smiling_face: , since the change did not alter any files other than the `utoipa-scalar` there is no need to address this in other crates `CHANGELOG` files.

terror:
> > > @terror Though missing `CHANGELOG.md` entry. That should be added. Should follow the same convention as already exists, if there is no `Unreleased` title in the beginning of the file, then new one should be added.
> > 
> > 
> > Ah, for sure. I just added one. It looks like theres a typo in the existing changelog filename, and its missing from the root file as well 🫠
> 
> Oh, well, I can fix the typos 🙂 , since the change did not alter any files other than the `utoipa-scalar` there is no need to address this in other crates `CHANGELOG` files.

Ah, I meant its missing from the root changelog, i.e. [this file](https://github.com/juhaku/utoipa/blob/master/CHANGELOG.md) as a reference.

juhaku:
> > > > @terror Though missing `CHANGELOG.md` entry. That should be added. Should follow the same convention as already exists, if there is no `Unreleased` title in the beginning of the file, then new one should be added.
> > > 
> > > 
> > > Ah, for sure. I just added one. It looks like theres a typo in the existing changelog filename, and its missing from the root file as well 🫠
> > 
> > 
> > Oh, well, I can fix the typos 🙂 , since the change did not alter any files other than the `utoipa-scalar` there is no need to address this in other crates `CHANGELOG` files.
> 
> Ah, I meant its missing from the root changelog, i.e. [this file](https://github.com/juhaku/utoipa/blob/master/CHANGELOG.md) as a reference.

Ah, :laughing: good catch, need to add that

Issue #856, opened 2024-02-05
SwaggerUI routing for subpath


dvu714:
Is there a setting for me to set to properly route to swagger ui hosted in a subpath of a domain with utoipa swagger ui for axum?  SwaggerUI seems to route calls for` /api-docs/openapi.json` to the root domain, i.e. `https://domain.com/api-docs/openapi.json`.

My api service is hosted in a subpath of a domain, i.e. `https://domain.com/foo/`. 
I'm looking for swagger ui to get `/api-docs/openapi.json` from `https://domain.com/foo/api-docs/openapi.json`

juhaku:
You should be able to provide a custom path for the api docs urls via config for the Swagger UI:
```rust
let config = Config::from("/api/foobar/api-docs/openapi.json");
let swagger = SwaggerUi::new("/swagger-ui/{_:.*}")
     .url("/api-docs/openapi.json", ApiDoc::openapi())
     .config(config); // <-- change the default config
```
See more details here: https://github.com/juhaku/utoipa/issues/731#issuecomment-1688043311

juhaku:
Closing as solved. Also this issue might be useful: #842. There is lot of discussion about serving the `utoipa` via reverse proxy. This can be opened if there is more to this.

Issue #842, opened 2024-01-11
Is there any way to configure utoipa behind reverse proxy?


qrilka:
OpenAPI supports servers but reverse proxy could be configured independently for a particular service sitting behind that proxy. `utoipa` has support for servers but it looks like the trait `Modify` is static. Is there some way to use X-Forwarded headers like for example it's done in https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/1801 ?

plang-arista:
Hi, I've also run into the same issue. 

My setup: Axum server running behind Reverse Proxy, with URL rewrite: "/api" -> "/"

I could modify the OpenAPI documentation with this trick:
```
struct PathPrefixAddon;

impl Modify for PathPrefixAddon {
    fn modify(&self, openapi: &mut utoipa::openapi::OpenApi) {
        match env::var("PATH_PREFIX") {
            Some(prefix) => openapi.servers = Some(vec![Server::new(prefix)]),
            None => ()
        }
    }
}

async fn main() {
    #[derive(OpenApi)]
    #[openapi(modifiers(&PathPrefixAddon))]
    struct ApiDoc;
}
```

However, I could not trick utoipa to host the actual `openapi.json` resource without the PathPrefix, while configuring the url with the PathPrefix for the `swagger-ui`.

I looked at the source code, it seems this use-case was not considered at all when setting up swagger-ui.

lpfy:
> tricks utoipa to use "/{prefix}" to access the server, which essentially adds a prefix to every endpoint.

Hi @plang-arista, thanks for your workaround. I can confirm it also works for the Swagger UI part on Actix-web, the next problem I am facing is the "Try it out" part still using the wrong endpoint.

for example, my server is http://www.abcd.com/, set Apache reverse proxy to http://www.abcd.com/rusttest/ for Rust Actix-web
The swagger UI URL becomes http://www.abcd.com/rusttest/swagger/
The actual API endpoint is http://www.abcd.com/rusttest/api/xxxxABC
The "Try it out" in utoipa swagger UI send request to http://www.abcd.com/api/xxxxABC, the prefix "/rusttest" is missing.

Could you please help to see what's wrong in my code? thanks heaps

```
use actix_cors::Cors;
use actix_web::{App, HttpServer, web::Data};
use std::{error::Error, env};
use utoipa::{OpenApi, Modify, openapi::server::Server};
use utoipa_swagger_ui::{SwaggerUi, Config};

[...]

PR #1530, opened 2026-03-12
fix(swagger-ui): Use relative redirect to add trailing slash

boblehest:
Given SwaggerUi::new("/path"), we would previously redirect requests for `/path` to `/path/` (adding the trailing slash). Using an absolute redirect path like this can cause issues when the HTTP server is behind a proxy doing path rewriting.

Example scenario detailing the issue:

Say you have a Rust application which exposes a HTTP server with `utoipa-swagger-ui` at path `/path/to/swagger/`. The server is available at hostname `my.server`.

You also have a proxy server at `my.proxy`, which handles incoming requests to the path `/api/...` by stripping the path prefix `/api`, and forwarding the request to `my.server`.

Then you do the following:

1. You send a request to `my.proxy/api/path/to/swagger`
2. The server at `my.proxy` forwards the request to `my.server/path/to/swagger` (note the stripped `/api` prefix)
3. The server at `my.server` redirects it to `/path/to/swagger/` (to "add a trailing slash")
4. The client then follows this redirect, sending a new request to `my.proxy/path/to/swagger/`
5. The request fails, because the proxy does not serve anything at this path (the path has no `/api/` prefix).

Solution:

Redirecting to the relative path `swagger/`* should be more robust, as it more precisely expresses the intent of simply adding a slash to the end of the path, instead of replacing the entire path.

*Or more generally, redirecting to `X/` where `X` is the last path segment of the configured swagger-ui path.

### DIFF
--- a/utoipa-swagger-ui/src/axum.rs
+++ b/utoipa-swagger-ui/src/axum.rs
@@ -63,7 +63,8 @@ where
             };
             debug_assert!(!path.is_empty());
 
-            let slash_path = format!("{}/", path);
+            let last_segment = path.rsplit('/').next().unwrap_or(path);
+            let slash_path = format!("{}/", last_segment);
             router
                 .route(
                     path,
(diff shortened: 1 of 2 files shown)


boblehest:
There are some previous issues which discuss serving SwaggerUi behind a proxy: #842 and #856 .

I've personally had two issues when trying to serve it being a proxy that does path rewriting.

This PR fixes one of them (people forgetting to add the trailing slash, getting redirected to a broken URL, then coming to ask me why my API doesn't host SwaggerUi).

The other problem I've run into is SwaggerUi not finding `openapi.json`, which I work around by following the hint from [this comment in issue 856](https://github.com/juhaku/utoipa/issues/856#issuecomment-2395441603):

```
SwaggerUi::new("/swagger-ui")
    .url("/api-docs/openapi.json", ApiDoc::openapi())
    .config(utoipa_swagger_ui::Config::new(["../api-docs/openapi.json"])),
                                           // ^-- override to use relative path
```

juhaku (commented):
This  does this for axum and rocket, but not for actix-web. Was it already implemented or is it missing by design? Also a CHANGELOG.md entry could be added to `utiopa-swagger-ui` You can look example from utoipa-gen.

Issue #1538, opened 2026-04-27
`repr` crate feature can cause unintended schema breakage


zackyancey:
The `#[repr()]` attribute is a language-level attribute that affects the
internal representation of the enum. It doesn't on its own say anything about
how the type is serialized, so we need some option to tell us whether to use
the `#[repr()]`

The current way of doing that, using the `repr` crate feature to control this
functionality, has a problem. If I have an enum definition like this:

```rust
// Note serde::Serialize, not serde_repr::Serialize, so I need strings, not
// ints, in the schema
#[derive(serde::Serialize, ToSchema)] 
#[repr(u8)] // repr is specified, but nothing to do with serialization
enum StringEnum {
    A = 0,
    B = 1
}
```

What I need to do is not enable the `repr` feature. However, because features
are addative and unified across all crates, if *any other* dependency of mine
enables the repr feature, then the schema for `StringEnum` will be wrong.

Here's how I think this can be solved (PR [here](https://github.com/juhaku/utoipa/pull/1539)):

This is something that needs to be specified per-schema, eg

```rust
#[derive(serde::Serialize, ToSchema)]
#[schema(repr)]
#[repr(u8)]
enum IntEnum {
    A = 0,
    B = 1
}

// To avoid the ambiguity caused by the feature, you can specify repr = false
// to ignore the `repr` feature if it's enabled.
#[derive(serde::Serialize, ToSchema)]
#[schema(repr = false)]
#[repr(u8)]
enum StringEnum {
    A = 0,
    B = 1
}
```

This change leaves the `repr` crate feature unchanged if no `#[schema(repr)]`
is specified, but if it is then that value will be used regardless of the state
of the `repr` feature. This lets new code use the unambiguous syntax without
breaking anything that depends on the feature. I think the `repr` crate feature
should probably be removed in the next major release.

esavier:
i made some pr comments, technically it looks good, but i am wondering whenever full path is covered, especially i am worried about `ToToken` path

PR #1539, opened 2026-04-27
Add a `repr` option for enums instead of relying on a feature

zackyancey:
This is to address [this issue](https://github.com/juhaku/utoipa/issues/1538):

The `#[repr()]` attribute is a language-level attribute that affects the internal representation of the enum. It doesn't on its own say anything about how the type is serialized, so we need some option to tell us whether to use the `#[repr()]`

The current way of doing that, using the `repr` crate feature to control this functionality, has a problem. If I have an enum definition like this:

```rust
// Note serde::Serialize, not serde_repr::Serialize, so I need strings, not
// ints, in the schema
#[derive(serde::Serialize, ToSchema)] 
#[repr(u8)] // repr is specified, but nothing to do with serialization
enum StringEnum {
    A = 0,
    B = 1
}
```

What I need to do is not enable the `repr` feature. However, because features are addative and unified across all crates, if *any other* dependency of mine enables the repr feature, then the schema for `StringEnum` will be wrong.

This is something that needs to be specified per-schema, eg

```rust
#[derive(serde::Serialize, ToSchema)]
#[schema(repr)]
#[repr(u8)]
enum IntEnum {
    A = 0,
    B = 1
}

// To avoid the ambiguity caused by the feature, you can specify repr = false
// to ignore the `repr` feature if it's enabled.
#[derive(serde::Serialize, ToSchema)]
#[schema(repr = false)]
#[repr(u8)]
enum StringEnum {
    A = 0,
    B = 1
}
```

This change leaves the `repr` crate feature unchanged if no `#[schema(repr)]` is specified, but if it is then that value will be used regardless of the state of the `repr` feature. This lets new code use the unambiguous syntax without breaking anything that depends on the feature. I think the `repr` crate feature should probably be removed in the next major release.


esavier:
@zackyancey can you update branch?

esavier:
minimally some docs are misleading, i..e repr is still there and working. I would suggest legacy/deprecation

utoipa-gen/src/lib.rs
```diff
@@ -605,6 +609,24 @@ static CONFIG: once_cell::sync::Lazy<utoipa_config::Config> =
 ///  }
 /// ```
 ///
+/// **Note!**: The deprecated `repr` feature enables this functionality
+/// automatically for any unit enum annotated with `#[repr(_)]`. Since features
+/// are unified across all dependencies, it is recommended to specify
+/// `#[schema(repr = false)` for any enum that has `#[repr(_)]` but doesn't use
```
esavier: `#[schema(repr = false)` -> `#[schema(repr = false)]`
`]` is missing

esavier:
we have `ToTokens` in `utoipa/utoipa-gen/src/component/features.rs`
that means if Feature::Repr is not taken out from pop_feature! code will try to call .repr(Some(true)) on schema which is not a builder method

this may work, or may not,
I would suggest treating it as As/RenameAll i.e. if it goes in `ToTokens` i would return some error

zackyancey:
@esavier Thanks for taking a look. I've merged in master and updated the docs, and made `Repr` return an error in ToTokensDiagnostics.

esavier:
@zackyancey 
I am not a maintainer, i just want to help a bit so treat this as triage.

Your solution is sound, imo, lets see what maintainers will do with it. 
btw kudos for fast response :+1: 

I guess its time to summon @juhaku

Issue #1165, opened 2024-10-25
Auto schema collection for params


dittos:
Auto schema collection added in utoipa 5 is really nice! 👍 
However it seems that schema in params are not collected. Is it just not implemented yet or not included intentionally?

```rust
use utoipa::{IntoParams, OpenApi, ToSchema};

#[derive(ToSchema)]
enum EnumValue {
    Apple,
    Orange,
}

#[derive(IntoParams)]
struct Params {
    x: EnumValue,
}

#[utoipa::path(
    get,
    path = "",
    params(
        ("value" = EnumValue, Query),
        Params,
    ),
)]
fn handler() {}

#[derive(OpenApi)]
#[openapi(
    paths(handler)
)]
struct Api;

fn main() {
    println!("{}", serde_json::to_string_pretty(&Api::openapi()).unwrap());
}
```

Generates the following without EnumValue in components.

```js
{
  "openapi": "3.1.0",
  "info": {
    "title": "utoipa-bug",
    "description": "",
    "license": {
      "name": ""
    },
    "version": "0.1.0"
  },
  "paths": {
    "": {
      "get": {
        "tags": [],
        "operationId": "handler",
        "parameters": [
          {
            "name": "value",
            "in": "query",
            "required": true,
            "schema": {
              "$ref": "#/components/schemas/EnumValue"
            }
          },
          {
            "name": "x",
            "in": "path",
            "required": true,
            "schema": {
              "$ref": "#/components/schemas/EnumValue"
            }
          }
        ],
        "responses": {}
      }
    }
  },
  "components": {} // <--
}
```

juhaku:
> However it seems that schema in params are not collected. Is it just not implemented yet or not included intentionally?

It was not implemented by design and party forgotten. This is something that need to be addressed in future.

juhaku:
Related #1271

diefans:
Is there a way to force this type into the schema?

PR #1545, opened 2026-05-06
Auto schema collection for params

reneklacan:
Implements https://github.com/juhaku/utoipa/issues/1165

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ Refer to the existing [examples](./examples) to find out more.
     Tuples, arrays and slices cannot be used as generic arguments on types. Types implementing `ToSchema` manually should not have generic arguments, as
     they are not composeable and will result compile error.
 * Automatic schema collection from usages recursively. 
+  * Schemas referenced by parameters defined with `params` or `IntoParams`.
   * Request body from either handler function arguments (if supported by framework) or from `request_body` attribute.
   * Response body from response `body` attribute or response `content` attribute.
 * Various OpenAPI visualization tools supported out of the box.
--- a/utoipa-gen/tests/openapi_derive.rs
+++ b/utoipa-gen/tests/openapi_derive.rs
@@ -166,6 +166,65 @@ fn derive_openapi_with_components_in_different_module() {
 }
 
 #[test]
+fn derive_openapi_collects_parameter_schema_references() {
+    use utoipa::IntoParams;
+
+    #[derive(ToSchema)]
+    #[allow(dead_code)]
+    enum ExplicitOrder {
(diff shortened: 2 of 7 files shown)


utoipa-gen/tests/openapi_derive.rs
```diff
@@ -165,6 +165,67 @@ fn derive_openapi_with_components_in_different_module() {
...
+    fn get_items(_params: SearchParams) {}
+
+    #[derive(OpenApi)]
+    #[openapi(paths(get_items))]
+    struct ApiDoc;
+
+    let doc = serde_json::to_value(ApiDoc::openapi()).unwrap();
+
+    assert!(doc.pointer("/components/schemas/ExplicitOrder").is_some());
+    assert!(doc.pointer("/components/schemas/DerivedOrder").is_some());
```
CommanderStorm: already tested below

```suggestion
```

README.md
```diff
@@ -50,6 +50,7 @@ Refer to the existing [examples](./examples) to find out more.
     Tuples, arrays and slices cannot be used as generic arguments on types. Types implementing `ToSchema` manually should not have generic arguments, as
     they are not composeable and will result compile error.
 * Automatic schema collection from usages recursively. 
+  * Path parameters from the `params` attribute and `IntoParams` types.
```
CommanderStorm: I have no clue what you mean here.
Could you reword this

CommanderStorm (approved):
Nice job resolving the papercut.

I have a small question how we communicate this feature to the outside, but otherwise LGTM

README.md
reneklacan: It was meant as continuation of from parent bullet point, so: "Automatic schema collection..." (of) "Path parameters from the `params` attribute and `IntoParams` types."

But I agree it sounds a bit off after re-reading it, rewrote it!

utoipa-gen/tests/openapi_derive.rs
reneklacan: I thought it provides extra layer of protection but it's fine either way, removed it

reneklacan:
@CommanderStorm thank you for the review!

utoipa-gen/tests/openapi_derive.rs
CommanderStorm: it is not extra protection because `None` also would error on that, and the error message is a bit better..

reneklacan:
@CommanderStorm anything else to make it mergeable?

CommanderStorm (approved):
LGTM from me as said. 
This need someone with review rights now

PR #1547, opened 2026-05-07
Replaced no longer maintained paste with pastey

dr0ps:
https://rustsec.org/advisories/RUSTSEC-2024-0436


CommanderStorm:
I would prefer we merge https://github.com/juhaku/utoipa/pull/1483 instead, because that PR replaces more of the usages

dr0ps:
See #1483

PR #1550, opened 2026-05-10
feat: readonly and writeonly dropped on ref fields

### LINKED ISSUE #596: Read Only is omitted for named struct fields
Currently when using `toSchema` on a struct that has another struct as a named field, the `read_only` attribute is omitted in the resulting schema for this type. Here is a full example:

```rust
use serde::Serialize;
use utoipa::ToSchema;

#[derive(Serialize, ToSchema)]
pub struct Child {
    pub name: String,
    pub description: String,
}

#[derive(Serialize, ToSchema)]
pub struct Parent {
    pub name: String,
    #[schema(read_only)]
    pub description: String,
    #[schema(read_only)]
    pub child: Child,
}

fn main() {
    let schema = Parent::schema();
    println!("{}", serde_json::to_string_pretty(&schema).unwrap());
}
```

This prints:
```json
[
  "Parent",
  {
    "type": "object",
    "required": [
      "name",
      "description",
      "child"
    ],
    "properties": {
      "name": {
        "type": "string"
      },
      "description": {
        "type": "string",
        "readOnly": true
      },
      "child": {
        "$ref": "#/components/schemas/Child"
      }
    }
  }
]
```

Note that the child has no readOnly Attribute.
The correct way to render this would be for OAS 3.0.X:

```json
"child": {
    "readOnly": true,
    "allOf": [
         "$ref": "#/components/schemas/Child"
    ]
}
```

This actually is supported by Swagger - the field does not show up for POST methods.
However in the `Schemas` section of the webpage, this seems to not be rendered.

In OAS 3.1 I believe that sibling keywords are supported out of the box.

Do you agree on this? Thanks for some feedback :)

Drifter-J:
Issue #596 stated that when struct has `#[schema(read_only)]` or `#[schema(write_only)]` and the field type is another struct (generating `$ref`), the attribute is dropped. Since utoipa uses openapi 3.1 now, sibling keywords alongside `$ref` are supported so I made a change to fix the existing issue.

Added 2 tests 

Fixes #596


utoipa/src/openapi/schema.rs
```diff
@@ -1451,6 +1481,16 @@ impl RefBuilder {
...
     }
+
+    /// Add or change read only flag for the reference.
+    pub fn read_only(mut self, read_only: Option<bool>) -> Self {
+        set_value!(self read_only read_only)
+    }
+
+    /// Add or change write only flag for the reference.
+    pub fn write_only(mut self, write_only: Option<bool>) -> Self {
+        set_value!(self write_only write_only)
```
CommanderStorm: I think this has a bit better as an API

```suggestion
    pub fn read_only(mut self, read_only: bool>) -> Self {
        set_value!(self read_only Some(read_only))
    }

    /// Add or change write only flag for the reference.
    pub fn write_only(mut self, write_only: bool) -> Self {
        set_value!(self write_only Some(write_only))
```

CommanderStorm (approved):
one minor nit, otherwise LGTM

utoipa/src/openapi/schema.rs
Drifter-J: ~~fixing the rest atm~~ actually nvm, I won't touch other code. I was referring to `ObjectBuilder` and whatnot. but will just push this change only.

juhaku: @Drifter-J @CommanderStorm A small note regarding this `Option<bool>` vs `bool`. It can be simply a `bool`, but that has one drawback. The logic behind many `Option<T>`  values there is to allow users have 3 states:

1. `Some(true)` to explicitly set a value to true.
2. `Some(false)` to explicitly set a value to false.
3. `None` to reset the value to the unset state. E.g. not to render it in the final OpenAPI json.

This is also true in many OpenAPI spec fields as they either can be set or ignored. Most of the cases the ignoring equals to `false` but that is just a matter of visual preference not change in functionality. 

By using only `bool` it removes the 3. state resetting option from the `Builder`. Of course it can be set to `None` when editing the type manually which is just fine but is a some price to pay in the ergonomics.

juhaku (approved):
Great 💯

Don't mind the examples not passing, they need some love, seems like time has eroded some of them.

utoipa-gen/tests/schema_derive_test.rs
```diff
@@ -1482,6 +1482,55 @@ fn derive_struct_with_read_only_and_write_only() {
...
+    };
+
+    assert_value! {user=>
+        "properties.address.$ref" = r###""#/components/schemas/Inner""###, "User address $ref"
+        "properties.address.readOnly" = r###"true"###, "User address read only"
+        "properties.address.writeOnly" = r###"null"###, "User address write only"
+        "properties.secret.$ref" = r###""#/components/schemas/Inner""###, "User secret $ref"
+        "properties.secret.writeOnly" = r###"true"###, "User secret write only"
+        "properties.secret.readOnly" = r###"null"###, "User secret read only"
+    }
```
juhaku: You could use the snapshot tests with the `insta` here also instead of this old `assert_value!` macro, there are plenty of examples. It provides better visual diff, but this will do it as well. 🙂

utoipa/src/openapi/schema.rs
```diff
@@ -1397,6 +1417,16 @@ builder! {
...
         pub summary: String,
+
+        /// Declares the property as "read only" alongside the `$ref`.
+        /// In OAS 3.1 sibling keywords next to `$ref` are allowed.
+        #[serde(rename = "readOnly", skip_serializing_if = "Option::is_none")]
+        pub read_only: Option<bool>,
+
+        /// Declares the property as "write only" alongside the `$ref`.
+        /// In OAS 3.1 sibling keywords next to `$ref` are allowed.
+        #[serde(rename = "writeOnly", skip_serializing_if = "Option::is_none")]
```
juhaku, line 1432: These comments could state that these can only be set withing a Schema object for sibling properties and when used with standalone `Reference` type these values should be omitted.

utoipa/src/openapi/schema.rs
Drifter-J: I agree that it can be simply bool - we need to change all three builders, is that what you want? @juhaku

utoipa-gen/tests/schema_derive_test.rs
Drifter-J: > You could use the snapshot tests with the `insta` here also instead of this old `assert_value!` macro, there are plenty of examples. It provides better visual diff, but this will do it as well. 🙂

will have a look after work :)

utoipa/src/openapi/schema.rs
juhaku: All three `Builder`s? Am I blind or what? :smile: Doesn't this only concern the `RefBuilder`? @Drifter-J

utoipa/src/openapi/schema.rs
Drifter-J: revised the comment :) thanks

utoipa-gen/tests/schema_derive_test.rs
Drifter-J: changed

PR #1552, opened 2026-05-14
ci: stop installing nightly in `test (examples)` job

### LINKED ISSUE #1551: CI: `test (examples)` job fails on master — rustix 0.37.28 (via tide) rejected by current nightly rustc
## Summary

The `test (examples)` job in `Utoipa build` (`.github/workflows/build.yaml`) has been failing on every push to `master` since early May 2026. The last green run was on commit `ignore really ignores! (#1500)` (2026-05-02). Every subsequent push, including `Emit nullable_item last for OneOfBuilder (#1299)` and `Add customizable page title to utoipa-scalar (#1515)`, shows the same failure.

This also makes PRs (e.g. #1463) appear red on a check that is unrelated to the PR's own changes.

## Failing run

Most recent failure: https://github.com/juhaku/utoipa/actions/runs/25870359184/job/76025...

```
Checking example: ./examples/todo-tide...
error: attributes starting with `rustc` are reserved for use by the `rustc` compiler
  --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rustix-0.37.28/src/backend/linux_raw/io/errno.rs:28:25
   |
28 | #[cfg_attr(rustc_attrs, rustc_layout_scalar_valid_range_start(0xf001))]
   |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: cannot find attribute `rustc_layout_scalar_valid_range_start` in this scope
error: cannot find attribute `rustc_layout_scalar_valid_range_end`   in this scope
error: could not compile `rustix` (lib) due to 4 previous errors
error: recipe `validate-examples` failed with exit code 101
```

## Root cause

The `test-examples-compile` job installs **two** toolchains via `dtolnay/rust-toolchain@master`:

```yaml
- name: Install stable Rust
  uses: dtolnay/rust-toolchain@master
  with: { toolchain: stable, components: clippy, rustfmt }

- name: Install nightly Rust
  uses: dtolnay/rust-toolchain@master
  with: { toolchain: nightly, components: clippy, rustfmt }
```

That action runs `rustup default <toolchain>` after each install, so **the second call wins** and nightly becomes the default. `just validate-examples` then invokes plain `cargo`, which means examples are clippied with the latest nightly rustc.

The `examples/todo-tide` example transitively depends on `rustix 0.37.28`:

```
rustix v0.37.28
└── async-io v1.13.0
    └── async-h1 v2.3.4
        └── tide v0.16.0
            └── todo-tide v0.1.0
```

`rustix 0.37.x` uses `#[cfg_attr(rustc_attrs, rustc_layout_scalar_valid_range_*)]` and detects nightly to enable the `rustc_attrs` cfg via its build script. A recent nightly rustc has tightened how `rustc_*` attribute names are validated, and now rejects them outside `rustc-internal` builds.

Confirmed locally:

- `cargo +stable clippy --all-features --all-targets --workspace` inside `examples/todo-tide` ✅ passes
- `cargo +nightly clippy --all-features --all-targets --workspace` inside `examples/todo-tide` ❌ fails the same way

`tide 0.16.0` (Aug 2021) is the last `tide` release and is effectively unmaintained, and `async-io 1.x`/`async-h1 2.x` are pinned to `rustix 0.37`, so we can't trivially bump out of this dependency tree.

## Proposed fix

The `test-examples-compile` job doesn't actually need nightly — `just validate-examples` only runs `cargo fmt --check` and `cargo clippy …` with the default toolchain, and the nightly-only justfile recipes (`coverage`, `doc`) prefix `cargo +nightly` explicitly. Either:

1. **Remove the `Install nightly Rust` step from the `test-examples-compile` job** (simplest, restores green build).
2. Reorder so stable is installed *after* nightly (also fine).
3. Add a `rust-toolchain.toml` pinning stable inside the `todo-tide` example.

Option 1 is the minimal, intent-preserving fix and is what I'd suggest.

## Disclosure

This issue and the accompanying fix PR were drafted with assistance from Anthropic Claude (Claude Code). All diagnostic work — log inspection, dependency tree analysis, local toolchain reproduction — was verified before filing.

JMLX42:
## Summary
- Removes the unused `Install nightly Rust` step from the `test-examples-compile` job.
- Restores green CI on `master` by ensuring `just validate-examples` runs under stable rustc.

## Why

The `test (examples)` job has been failing on every push to `master` since early May 2026 (see #1551). `dtolnay/rust-toolchain@master` runs `rustup default <toolchain>` after each install, so installing nightly *after* stable made nightly the default. `just validate-examples` then invoked plain `cargo`, compiling every example under the latest nightly rustc.

A recent nightly tightened how `rustc_*` attribute names are validated, which trips `rustix 0.37.28` (pulled in transitively by `tide 0.16` → `async-h1 2.3.4` → `async-io 1.13.0`):

```
error: attributes starting with `rustc` are reserved for use by the `rustc` compiler
  --> rustix-0.37.28/src/backend/linux_raw/io/errno.rs:28:25
   | #[cfg_attr(rustc_attrs, rustc_layout_scalar_valid_range_start(0xf001))]
error: could not compile `rustix` (lib) due to 4 previous errors
```

Stable rustc still accepts the gated attribute, so dropping the unused nightly install is the minimal fix and avoids reaching into the unmaintained `tide` dependency tree.

Nothing in the workflow or `just validate-examples` recipe needs nightly — the only nightly-only recipes (`coverage`, `doc`) prefix `cargo +nightly` explicitly.

## Test plan

- [x] Reproduced the failure locally: `cargo +nightly clippy --all-features --all-targets --workspace --quiet` inside `examples/todo-tide` fails with the rustix errors above.
- [x] Verified the fix locally: `just validate-examples` succeeds under `stable-x86_64-unknown-linux-gnu` (rustc 1.94.1), printing `All examples are valid!`.
- [x] Verified `cargo +stable clippy --all-features --all-targets --workspace` on `examples/todo-tide` returns warnings only, exit 0.

Closes #1551

---

**Disclosure:** Diagnosis and patch drafted with assistance from Anthropic Claude (Claude Code). All findings — log inspection, dependency-tree reverse lookup, local toolchain reproduction — were verified before submitting.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -126,12 +126,6 @@ jobs:
           toolchain: stable
           components: clippy, rustfmt
 
-      - name: Install nightly Rust
-        uses: dtolnay/rust-toolchain@master
-        with:
-          toolchain: nightly
-          components: clippy, rustfmt
-
       - uses: actions/cache@v3
         with:
           path: |


juhaku (approved):
Yeah, let's remove that, because we are anyways aiming the support for stable. For if some example explicitly needs nightly or something else that should be then managed by the `rust-toolchain.toml`

PR #1555, opened 2026-05-16
Add OpenAPI 3.2 support

### LINKED ISSUE #686: Support for Callbacks and Webhooks as defined in OpenAPI v3.1
Hello guys,

are there any plans to implement support for Callbacks and Webhooks?

### LINKED ISSUE #1480: sse support for openapi 3.2
https://spec.openapis.org/oas/v3.2#server-sent-event-streams

pjankiewicz:
## Summary

This PR adds first-class OpenAPI 3.2 support to utoipa and updates the generated default version from OpenAPI 3.1.0 to 3.2.0.

The implementation is based on the official OpenAPI 3.2 specification and the official OpenAPI 3.2 JSON Schema, then mapped onto utoipa's existing Rust data model and builder APIs.

## Sources Used

- OpenAPI 3.2.0 specification: https://spec.openapis.org/oas/v3.2.0.html
- Official OpenAPI 3.2 JSON Schema, 2025-11-23 revision: https://spec.openapis.org/oas/3.2/schema/2025-11-23
- The official schema was used as the fixed-field checklist for OpenAPI Object, Components Object, Media Type Object, Encoding Object, Path Item Object, Operation Object, Parameter Object, Response Object, Security Scheme Object, Server Object, Tag Object, Example Object, Header Object, XML Object, and Link Object.

## Main Changes

- `OpenApiVersion` now defaults to `Version32`, serializes as `"3.2.0"`, and keeps `Version31` as deprecated compatibility.
- Added OpenAPI 3.2 root fields:
  - `$self`
  - `webhooks`
  - `jsonSchemaDialect`
- Expanded Components Object support:
  - `mediaTypes`
  - `parameters`
  - `examples`
  - `requestBodies`
  - `headers`
  - `links`
  - `callbacks`
  - `pathItems`
- Added or aligned OpenAPI 3.2 operation/path support:
  - Path Item `$ref`
  - `query` operation
  - `additionalOperations`
  - callbacks as structured callback maps instead of the previous placeholder string
  - `RefOr` support for parameter/request body/header/content/reference-capable locations where the 3.2 schema allows references
- Added streaming/sequential media support:
  - Media Type `itemSchema`
  - Media Type `description`
  - Media Type `itemEncoding`
  - Media Type `prefixEncoding`
  - nested Encoding `encoding`, `itemEncoding`, and `prefixEncoding`
- Added Server-Sent Events and typed streaming payload modeling:
  - SSE item schema with `event`, `data`, `id`, and `retry`
  - typed payload references from stream items to reusable component schemas
  - embedded encoded content via `contentMediaType` + `contentSchema`
  - examples for `text/event-stream`, `application/jsonl`, `application/json-seq`, and `multipart/mixed`
- Added other OpenAPI 3.2 fixed-field updates:
  - Tag `summary`, `parent`, and `kind`
  - Server `name`
  - XML `nodeType`
  - Response `summary`
  - Example `dataValue` and `serializedValue`
  - Parameter location `querystring`
  - Parameter style `cookie`
  - OAuth2 `deviceAuthorization`
  - OAuth2 `oauth2MetadataUrl`
  - `deprecated` support on security schemes
- Corrected Link Object serialization to use OpenAPI camelCase fixed fields:
  - `operationId`
  - `operationRef`
  - `requestBody`

## Macro / Derive Updates

- `#[derive(OpenApi)]` tag parsing now accepts and emits `summary`, `parent`, and `kind`.
- Path parameter parsing supports `querystring` and `cookie` style.
- Path examples support `data_value` and `serialized_value`.
- Response header generation was updated for optional schemas and content-capable headers.
- Generated media type tokens were adjusted to avoid ambiguous `Into` inference after content maps gained reference support.

## New Test Coverage

Added OpenAPI 3.2-focused tests in `utoipa/src/openapi.rs`:

- `serialize_deserialize_openapi_version_success`
  - Verifies 3.2 serialization/deserialization and 3.1 compatibility.
- `openapi_32_root_self_and_webhooks_serialize`
  - Covers `$self`, `jsonSchemaDialect`, and `webhooks`.
- `openapi_32_components_and_callbacks_use_spec_data_model`
  - Covers reusable components added/expanded in 3.2, callback maps, path items, media types, references, and corrected Link Object field names.
- `openapi_32_tags_server_xml_and_examples_serialize`
  - Covers hierarchical tags, server name, XML `nodeType`, and richer examples.
- `openapi_32_query_and_additional_operations_serialize`
  - Covers Path Item `query` and `additionalOperations`.
- `openapi_32_querystring_parameter_and_cookie_style_serialize`
  - Covers `querystring` parameters, `cookie` style, and parameter content/examples.
- `openapi_32_components_media_types_streaming_and_response_summary_serialize`
  - Covers `mediaTypes`, `itemSchema`, streaming response summaries, SSE event shapes, typed event payload refs, embedded encoded JSON content, JSON Lines, JSON Text Sequences, and multipart streams.
- `openapi_32_oauth_device_authorization_and_deprecated_security_serialize`
  - Covers OAuth2 device authorization, OAuth2 metadata URL, and security scheme deprecation.

Existing snapshots were updated where the OpenAPI output intentionally changed, especially:

- generated `openapi` version from `3.1.0` to `3.2.0`
- empty response descriptions omitted where 3.2 allows summary-only/no-description response shapes
- Link Object field names corrected to OpenAPI camelCase

## Verification

Ran successfully:

```bash
cargo test -p utoipa -p utoipa-gen
```

The command passes locally. Remaining output is warnings only from existing warning categories.

Resolves #1480 Resolves #686


pjankiewicz:
Fixed the failing tests.

pjankiewicz:
Conflicts with `master` resolved in 27d6c52.

The conflict was in `utoipa/src/openapi/header.rs` between #1556 (expanded `Header` struct with `deprecated`, `style`, `explode`, `example`, `examples`, `content`, `extensions`) and this PR's OpenAPI 3.2 changes (`RefOr<Content>` for the header content map, `Option<I>` `schema` builder, and the per-item `.content()` builder). Resolved by taking the full upstream field set while keeping the OpenAPI 3.2 builder signatures and `RefOr<Content>` typing; the new `content_from_iter` was updated to accept `Into<RefOr<Content>>` accordingly, and the upstream test that constructs `Header` directly was updated to wrap its `Content` with `.into()`.

CI-equivalent local verification:
- `cargo fmt --check -p utoipa -p utoipa-gen` — clean
- `cargo clippy -p utoipa -p utoipa-gen` — no errors
- `just test utoipa` — 87 passed
- `just test utoipa-gen` — all suites green
- `cargo build --workspace` — clean

juhaku:
Thanks, this is a massive PR. I need to take a look at this when I have time.

CommanderStorm (changes requested):
Can you please split this up into multiple PRs?
1.5k of macro code is a lot...

(not a maintainer)

pjankiewicz:
> Can you please split this up into multiple PRs? 1.5k of macro code is a lot...
> 
> (not a maintainer)

1500 LOC for this PR is not a lot and most of it is not macro related code. At least 500 LOC are new tests. And the majority of the other changes are related to data structures.

I don't want to split it artificially because it can get more complex to track it later on. Maybe it could be a good idea to merge it into some major release candidate so people start testing it without breaking the official releases.

utoipa/src/openapi.rs
```diff
@@ -119,13 +124,21 @@ builder! {
         #[serde(skip_serializing_if = "Option::is_none")]
         pub external_docs: Option<ExternalDocs>,
 
+        /// The default JSON Schema dialect used by Schema Objects in this OpenAPI document.
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub json_schema_dialect: Option<String>,
+
         /// Schema keyword can be used to override default _`$schema`_ dialect which is by default
         /// “<https://spec.openapis.org/oas/3.1/dialect/base>”.
```
poliorcetics, line 132: This may need an update to 3.2, though simply changing the version gives 404

utoipa/src/openapi.rs
```diff
@@ -119,13 +124,21 @@ builder! {
...
         ///
         /// All the references and individual files could use their own schema dialect.
         #[serde(rename = "$schema", default, skip_serializing_if = "String::is_empty")]
         pub schema: String,
 
+        /// URI identifying this OpenAPI document.
+        #[serde(rename = "$self", skip_serializing_if = "Option::is_none")]
+        pub self_url: Option<String>,
```
poliorcetics:
```suggestion
        pub self_uri: Option<String>,
```

To use the same wording for docs and rust name and follow the spec wording too (in https://spec.openapis.org/oas/v3.2.0.html#fixed-fields)

utoipa/src/openapi.rs
```diff
@@ -373,6 +391,16 @@ impl OpenApiBuilder {
...
+    /// Add or change the default JSON Schema dialect for Schema Objects.
+    pub fn json_schema_dialect<S: Into<String>>(mut self, json_schema_dialect: Option<S>) -> Self {
+        set_value!(self json_schema_dialect json_schema_dialect.map(Into::into))
+    }
+
+    /// Add or change the URI identifying this OpenAPI document.
+    pub fn self_url<S: Into<String>>(mut self, self_url: Option<S>) -> Self {
+        set_value!(self self_url self_url.map(Into::into))
+    }
 }
```
poliorcetics:
```suggestion
    pub fn self_uri<S: Into<String>>(mut self, self_uri: Option<S>) -> Self {
        set_value!(self self_uri self_uri.map(Into::into))
    }
}
```

utoipa/src/openapi.rs
```diff
@@ -381,10 +409,14 @@ impl OpenApiBuilder {
...
-    /// Will serialize to `3.1.0` the latest released OpenAPI version.
+    /// Will serialize to `3.1.0`.
+    #[deprecated(note = "OpenAPI 3.1 is superseded by 3.2. Use Version32.")]
     #[serde(rename = "3.1.0")]
-    #[default]
     Version31,
+    /// Will serialize to `3.2.0` the latest supported OpenAPI version.
+    #[serde(rename = "3.2.0")]
+    #[default]
+    Version32,
```
poliorcetics: changing the default is a breaking change that I'm not sure we want to do before having real life testing from downstream users

utoipa/src/openapi.rs
```diff
@@ -418,9 +450,15 @@ impl<'de> Deserialize<'de> for OpenApiVersion {
                     .collect::<Vec<_>>();
 
                 if version.len() == 3 && version.first() == Some(&3) && version.get(1) == Some(&1) {
+                    #[allow(deprecated)]
                     Ok(OpenApiVersion::Version31)
+                } else if version.len() == 3
+                    && version.first() == Some(&3)
+                    && version.get(1) == Some(&2)
+                {
+                    Ok(OpenApiVersion::Version32)
```
poliorcetics:
```suggestion
                if matches!(&version[..], [Some(3), Some(1), _]) {
                    #[allow(deprecated)]
Ok(OpenApiVersion::Version31)
                } else if matches!(&version[..], [Some(3), Some(2), _]) 
                {
                    Ok(OpenApiVersion::Version32)
```

I think that works, the syntax may be slightly wrong around references though

utoipa/src/openapi/security.rs
```diff
@@ -347,6 +367,13 @@ impl HttpBuilder {
 
         self
     }
+
+    /// Add or change deprecated status.
+    pub fn deprecated(mut self, deprecated: Option<Deprecated>) -> Self {
+        self.deprecated = deprecated;
+
+        self
+    }
```
poliorcetics, line 375:
```suggestion
        self.deprecated = deprecated;
        self
    }
```

utoipa/src/openapi/path.rs
```diff
@@ -356,10 +373,25 @@ impl PathItem {
...
+        if path_item.query.is_some() && self.query.is_none() {
+            self.query = path_item.query;
+        }
+        if !path_item.additional_operations.is_empty() {
+            for (method, operation) in path_item.additional_operations {
+                self.additional_operations
+                    .entry(method)
+                    .or_insert(operation);
+            }
+        }
```
poliorcetics, line 383: the non-empty check is redundant here

utoipa/src/openapi/path.rs
```diff
@@ -586,9 +662,21 @@ impl OperationBuilder {
     /// Append parameter to [`Operation`] parameters.
     pub fn parameter<P: Into<Parameter>>(mut self, parameter: P) -> Self {
         match self.parameters {
-            Some(ref mut parameters) => parameters.push(parameter.into()),
+            Some(ref mut parameters) => parameters.push(parameter.into().into()),
```
poliorcetics: why do we need the two intos here ?

utoipa-gen/src/lib.rs
```diff
@@ -1769,7 +1769,7 @@ pub fn derive_to_schema(input: TokenStream) -> TokenStream {
 /// ```
 ///
 /// _**Define context path for endpoint. The resolved **path** shown in OpenAPI doc will be `/api/pet/{id}`.**_
-/// ```rust
+/// ```rust,ignore
```
poliorcetics: if the tests were passing before and they don't know, that's not a good change ?

poliorcetics (commented):
Did a first pass without reading through the corresponding spec changes, just commenting on the Rust side of it

utoipa/src/openapi.rs
pjankiewicz: Renamed the field to `self_uri` to match the spec wording and the doc comment.

utoipa/src/openapi.rs
pjankiewicz: Done - renamed the builder method to `self_uri` as well.

utoipa/src/openapi/security.rs
pjankiewicz: Applied.

utoipa/src/openapi/path.rs
pjankiewicz: Removed — iterating an empty map is a no-op, so the guard wasn't buying anything.

utoipa/src/openapi.rs
pjankiewicz: That's a fair concern, and I'd rather not flip the default output silently before 3.2 tooling support is broad. Reverted the default back to `Version31` (3.1.0) with no deprecation, and made `Version32` opt-in. Added `OpenApiBuilder::openapi(..)` so 3.2 can be selected explicitly, and rolled the affected snapshots back to 3.1.0. We can revisit flipping the default once downstream tooling has caught up.

utoipa/src/openapi.rs
pjankiewicz: Applied. `version` is a `Vec<i8>` so the elements aren't `Option`; using `matches!(version.as_slice(), [3, 1, _])` and `[3, 2, _]` — reads much cleaner.

utoipa/src/openapi/path.rs
pjankiewicz: `parameters` is `Vec<RefOr<Parameter>>`, so the first `.into()` was `P -> Parameter` (from the `P: Into<Parameter>` bound) and the second `Parameter -> RefOr<Parameter>`. Made it explicit as `RefOr::T(parameter.into())` here and in the bulk `parameters` setter.

utoipa/src/openapi.rs
pjankiewicz: Good point. Since the default output version is now reverted to 3.1.0 (see the `OpenApiVersion` thread), the documented `https://spec.openapis.org/oas/3.1/dialect/base` default is still accurate, so I've left the `$schema` doc as-is. The new `json_schema_dialect` field is there for anyone who wants to point at a 3.2 dialect explicitly once a stable URL is published.

utoipa-gen/src/lib.rs
pjankiewicz: These do run and pass — but only under the `actix_extras` feature, which the CI `just test utoipa-gen` step enables via `utoipa/actix_extras`. Without it, the HTTP method can't be inferred from `#[get(...)]` and they fail to build, which is why a plain `cargo test --doc -p utoipa-gen` shows them red. Reverted them back to `rust` rather than hiding them. Verified locally: `cargo test --doc -p utoipa-gen --features actix_extras` → 75 passed, 0 failed.

pjankiewicz:
Thanks for the careful pass, @poliorcetics - pushed a commit (a6bd044) addressing the Rust-side feedback:

- **Default version**: reverted the default back to `Version31` (3.1.0); `Version32` is now opt-in with no deprecation, so the generated output stays non-breaking. Added `OpenApiBuilder::openapi(..)` to select 3.2 explicitly, and rolled the affected snapshots back to 3.1.0.
- **`self_url` → `self_uri`**: renamed field and builder method to match the spec wording.
- **`OpenApiVersion` deserialize**: simplified to slice `matches!` patterns (`[3, 1, _]` / `[3, 2, _]`).
- **`additional_operations` merge**: dropped the redundant non-empty guard.
- **Double `.into()`**: made the `RefOr<Parameter>` conversion explicit via `RefOr::T(parameter.into())`.
- **`HttpBuilder::deprecated`**: formatting tidied.
- **Doc examples**: reverted the actix `#[utoipa::path]` examples from `rust,ignore` back to `rust` - they pass under the `actix_extras` feature that CI enables (`cargo test --doc -p utoipa-gen --features actix_extras` → 75 passed); they only failed without the feature, so hiding them was the wrong fix.
- Left the `$schema` 3.1 dialect doc as-is, since the default is back to 3.1.0.

`cargo test -p utoipa -p utoipa-gen` passes locally with the CI feature sets. Replied inline on each thread too.

utoipa/src/openapi/encoding.rs
```diff
@@ -29,7 +29,7 @@ builder! {
         /// Content-Disposition. Content-Type is described separately and SHALL be ignored in this
         /// section. This property SHALL be ignored if the request body media type is not a multipart.
         #[serde(skip_serializing_if = "BTreeMap::is_empty")]
-        pub headers: BTreeMap<String, Header>,
+        pub headers: BTreeMap<String, RefOr<Header>>,
```
juhaku, line 32: This itself already makes this a breaking change. Though it is fine since there already are breaking changes in the mastser.

utoipa/src/openapi/path.rs
```diff
@@ -396,7 +462,12 @@ impl PathItemBuilder {
 
     /// Append list of [`Parameter`]s common to all [`Operation`]s to this [`PathItem`].
     pub fn parameters<I: IntoIterator<Item = Parameter>>(mut self, parameters: Option<I>) -> Self {
-        set_value!(self parameters parameters.map(|parameters| parameters.into_iter().collect()))
+        set_value!(self parameters parameters.map(|parameters| parameters.into_iter().map(Into::into).collect()))
+    }
+
+    /// Append list of parameter references common to all [`Operation`]s to this [`PathItem`].
+    pub fn parameter_refs<I: IntoIterator<Item = Ref>>(mut self, parameters: Option<I>) -> Self {
+        set_value!(self parameters parameters.map(|parameters| parameters.into_iter().map(RefOr::Ref).collect()))
```
juhaku, line 470:
```suggestion
    pub fn parameter_refs<I: IntoIterator<Item = Ref>>(mut self, refs: Option<I>) -> Self {
        set_value!(self parameters refs.map(|refs| refs.into_iter().map(RefOr::Ref).collect()))
```
Nothing alarming, but perhaps refs is the correct term, since it is accepting refs as an argument. 🙂

utoipa/src/openapi/security.rs
```diff
@@ -753,6 +819,64 @@ impl AuthorizationCode {
...
+    #[serde(flatten)]
+    pub scopes: Scopes,
+
+    /// Optional extensions "x-something".
+    #[serde(skip_serializing_if = "Option::is_none", flatten)]
+    pub extensions: Option<Extensions>,
+}
+
+impl DeviceAuthorization {
+    /// Construct a new device authorization OAuth2 flow.
```
juhaku, line 848: The docs could be more elaborative here and many places, It is always a good practice to provide examples. Users would love to know how to use the crate without guessing. 🙂

juhaku (approved):
Overall looks good, thanks for hard work and putting the effort into making this. ❤️ 

If something, the docs could be improved a bit. Not everything needs an example, but some e.g. the device authentication  would benefit from it.

Drifter-J:
> Overall looks good, thanks for hard work and putting the effort into making this. ❤️
> 
> If something, the docs could be improved a bit. Not everything needs an example, but some e.g. the device authentication would benefit from it.

@pjankiewicz and @juhaku I have my branch fixing docs, adding changelog and renaming `parameters` to `refs`. I will merge this PR and follow up before we release 6.0.0

Here we go - https://github.com/juhaku/utoipa/pull/1590 (I merged cause test in CI pipeline was broken after merging PR-1555)

PR #1556, opened 2026-05-25
feat: expand header struct

### LINKED ISSUE #819: Missing some fields in Header struct
The Header struct misses the fileds inherited from Parameter struct. For example, "required", "deprecated", "style", "explode", "examples", "example", "content" and "extensions".

Please verify the resources you supplied.
[ https://spec.openapis.org/oas/latest.html#header-object](url)
It saies "The Header Object follows the structure of the [Parameter Object](https://spec.openapis.org/oas/latest.html#parameterObject)", not only the two fields of the example.

    /// Implements [OpenAPI Header Object][header] for response headers.
    ///
    /// [header]: https://spec.openapis.org/oas/latest.html#header-object
    #[non_exhaustive]
    #[derive(Serialize, Deserialize, Clone, PartialEq)]
    pub struct Header {
        /// Schema of header type.
        pub schema: RefOr<Schema>,

        /// Additional description of the header value.
        pub description: Option<String>,
    }

Drifter-J:
## Description

Following up with https://github.com/juhaku/utoipa/issues/819

Added missing openAPI Header Fields:
* deprecated
* style
* explode
* example
* examples
* content
* extensions

Updated `schema` as `Option<RefOr<Schema>>` to allow spec compliant content-based headers.

Added unit tests 

Closes: #819

### DIFF
--- a/utoipa-gen/tests/schema_derive_test.rs
+++ b/utoipa-gen/tests/schema_derive_test.rs
@@ -3498,7 +3498,7 @@ fn test_new_type_struct_min_items() {
     let schema = <Tags as PartialSchema>::schema();
     let value = serde_json::to_value(schema).expect("schema is JSON serializable");
 
-    assert_json_snapshot!(value); 
+    assert_json_snapshot!(value);
 }
 
 #[test]
--- a/utoipa/src/openapi/header.rs
+++ b/utoipa/src/openapi/header.rs
@@ -3,8 +3,13 @@
 //! [header]: https://spec.openapis.org/oas/latest.html#header-object
 
 use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use std::collections::BTreeMap;
 
-use super::{builder, set_value, Object, RefOr, Schema, Type};
+use super::{
+    builder, content::Content, example::Example, extensions::Extensions, path::ParameterStyle,
+    set_value, Deprecated, Object, RefOr, Schema, Type,
+};
 
 builder! {
     HeaderBuilder;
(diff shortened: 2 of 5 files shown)


juhaku:
Looks good, :tada: missing a CHANGELOG.md for the utoipa. Also the fmt fix is not a biggie but could be a separate mini PR so it would not pollute the context with unrelated content.

utoipa-gen/tests/schema_derive_test.rs
```diff
@@ -3498,7 +3498,7 @@ fn test_new_type_struct_min_items() {
     let schema = <Tags as PartialSchema>::schema();
     let value = serde_json::to_value(schema).expect("schema is JSON serializable");
 
-    assert_json_snapshot!(value); 
+    assert_json_snapshot!(value);
```
Drifter-J, line 3501:
```suggestion
    assert_json_snapshot!(value); 
```

Drifter-J:
> Looks good, 🎉 missing a CHANGELOG.md for the utoipa. Also the fmt fix is not a biggie but could be a separate mini PR so it would not pollute the context with unrelated content.

change log is added and 
https://github.com/juhaku/utoipa/pull/1557 is raised :)

ty!

juhaku (approved):
Seems like green, 🥇

utoipa/src/openapi/header.rs
```diff
@@ -17,11 +22,40 @@ builder! {
     #[cfg_attr(feature = "debug", derive(Debug))]
     pub struct Header {
         /// Schema of header type.
-        pub schema: RefOr<Schema>,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub schema: Option<RefOr<Schema>>,
```
juhaku, line 26: This is actually breaking change. We can merge this though, but then the next release must be major. Unless this PR will be ignored from the next release.

Drifter-J: There are a couple of changes that we can squeeze if we release a major e.g. serde_norway -> yaml_serde https://github.com/juhaku/utoipa/pull/1559

PR #1559, opened 2026-05-31
Add yaml_serde feature for yaml deserialization

BjarneSeger:
The current `yaml` feature uses the `serde_norway`-crate, which was not updated since nearly 1.5 years. This commit introduces the new `yaml_serde`-feature, which provides nearly the same interface as serde_norway does, but with a crate that is more actively maintained and owned by the official yaml github organization.

The current implementation is about 90% done, but it does not (really) cover what happens if both yaml features are enabled since they both provide the same function but with a different signature. Depending on your preference, I can just make a cfg block that checks if both are enabled and emits a more helpful message or rename the function, depending on preference.

### DIFF
--- a/utoipa/src/openapi.rs
+++ b/utoipa/src/openapi.rs
@@ -173,7 +173,7 @@ impl OpenApi {
     /// Converts this [`OpenApi`] to YAML String. This method essentially calls [`yaml_serde::to_string`].
     #[cfg(feature = "yaml_serde")]
     #[cfg_attr(doc_cfg, doc(cfg(feature = "yaml")))]
-    pub fn to_yaml(&self) -> Result<String, yaml_serde::Error> {
+    pub fn to_yaml_serde(&self) -> Result<String, yaml_serde::Error> {
         yaml_serde::to_string(self)
     }
 
--- a/utoipa/src/openapi.rs
+++ b/utoipa/src/openapi.rs
@@ -170,6 +170,13 @@ impl OpenApi {
         serde_norway::to_string(self)
     }
 
+    /// Converts this [`OpenApi`] to YAML String. This method essentially calls [`yaml_serde::to_string`].
+    #[cfg(feature = "yaml_serde")]
+    #[cfg_attr(doc_cfg, doc(cfg(feature = "yaml")))]
+    pub fn to_yaml(&self) -> Result<String, yaml_serde::Error> {
+        yaml_serde::to_string(self)
+    }
+
     /// Merge `other` [`OpenApi`] moving `self` and returning combined [`OpenApi`].
     ///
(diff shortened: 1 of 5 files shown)


Drifter-J:
Hi @BjarneSeger, thanks for raising this. Can we completely move away from `serde_norway`? What are the risks? It's just creating a valid yaml syntax 🤔

BjarneSeger:
The to_yaml function currently returns `serde_norway::Error`. I could try and hack together a mapping function, but that seems a little excessive.
You could probably just deprecate the feature and just have it enable yaml_serde in a future breaking release

utoipa/src/openapi.rs
```diff
@@ -170,6 +170,13 @@ impl OpenApi {
         serde_norway::to_string(self)
     }
 
+    /// Converts this [`OpenApi`] to YAML String. This method essentially calls [`yaml_serde::to_string`].
+    #[cfg(feature = "yaml_serde")]
+    #[cfg_attr(doc_cfg, doc(cfg(feature = "yaml")))]
+    pub fn to_yaml(&self) -> Result<String, yaml_serde::Error> {
```
CommanderStorm: Please make this feature additive.

Non-aadditive features means that if one of your dependencys activates the `yaml` feature, you now will not be able to enable the `yaml_serde` feature.

naming it by what it really is fixes this, even though the naming containing the dependency is arguably also not very pretty

```suggestion
    pub fn to_yaml_serde(&self) -> Result<String, yaml_serde::Error> {
```

docs -> https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#feature-unification

juhaku:
This is fine, though, I would prefer to completely remove the yaml norway thingy and have the method `fn to_yaml() -> Result<Yaml, Error>` instead. But that can be done in another PR before release. There is already breaking one breaking change in the master.

juhaku (approved):

juhaku:
Here is the follow up issue: #1565

juhaku:
@BjarneSeger  Damn :facepalm: I broke your branch. I wanted up update over the master but pushed wrong branch on top of it, I am not able to push my local changes to the branch anymore which is weird. You should be able to restore it from the reflog. Sorry, alternatively I can create PR with your changes. I can add you to the co authros? How do you want to proceed?

juhaku:
@BjarneSeger  here is your branch changes: https://github.com/juhaku/utoipa/compare/serde-yaml-bjarne-seger

juhaku:
The actix-web is broken, nothing releated to this PR though. Need to fix that in another instance.

juhaku:
@BjarneSeger thanks for quick fix :1st_place_medal: :partying_face:

PR #1568, opened 2026-06-14
ci: constrain brotli-decompressor to fix actix-web build

MarkusZoppelt:
`brotli-decompressor 5.0.2` accidentally allows `alloc-no-stdlib 3.0.0`, which doesn't play nice with `brotli 8.x`'s `alloc-no-stdlib 2.0` requirement.

Result: trait-mismatch build failures in all actix-web CI jobs.

`5.0.3` tightens things back up, so require it anywhere `actix-web` / `brotli` creeps in.

Added as dev-dependencies for library crates and regular dependencies for examples.

### DIFF
--- a/examples/actix-web-multipart/Cargo.toml
+++ b/examples/actix-web-multipart/Cargo.toml
@@ -9,5 +9,6 @@ actix-multipart = "0.7"
 utoipa = { path = "../../utoipa", features = ["actix_extras"] }
 utoipa-swagger-ui = { path = "../../utoipa-swagger-ui", features = ["actix-web"] }
 utoipa-actix-web = { path = "../../utoipa-actix-web" }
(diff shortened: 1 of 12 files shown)


juhaku (commented):
While this is good and makes the build working. we might need to later to remove these dependency locks when the library authors will actually fix their semantic versioning. It would be good to actually raise issue to those projects that they are breaking the build of downstream crates if not created already.

juhaku:
@MarkusZoppelt The brotli `8.0.4` has been released 12 hours ago: https://crates.io/crates/brotli/versions (since writing this comment), and when the build issue originally happened it was not yet released. This release renders this dependency locking obsolete. Thus I'd suggest revert this PR because it is a ticking time bomb to pending to explode once newer versions will be released and actix-web moves on to those.

PR #1572, opened 2026-06-15
feat: Support serde(flatten) in IntoParams derive

### LINKED ISSUE #1531: How to flatten a nested struct for display
“How can I make this structure

pub struct UserListQuery {
    #[validate(nested)]
    #[serde(flatten)]
    // #[param(inline)]
    pub pagination: PaginationParams,
    /// Name keyword
    #[param(example = "")]
    pub name: Option<String>,
}

generate a Swagger document where pagination is expanded, so that the final fields are page, size, and name?”

Image

esavier:
fixes: #1531

Expand flattened fields that implement IntoParams into the parent parameter list instead of generating a parameter for the flattened field itself.

### DIFF
--- a/utoipa-gen/CHANGELOG.md
+++ b/utoipa-gen/CHANGELOG.md
@@ -19,6 +19,7 @@
 ### Fixed
 
 * Avoid degenerate `oneOf` for `Option<_>` with `nullable = false` and `default`/`title`/`description` (https://github.com/juhaku/utoipa/pull/1380)
+* Support `#[serde(flatten)]` fields that implement `IntoParams` in `IntoParams` derive.
 
 ## 5.5.0 - May 5 2026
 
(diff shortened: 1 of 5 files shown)


esavier:
i hope i did this correctly, please add your comments if we think this should be done differently

CommanderStorm:
Can you compare your PR with the other open PR implementing this?
Which one is better and why?
- https://github.com/juhaku/utoipa/pull/1426/changes

Your code does not have the weird `utoipa-gen/tests/into_params_test.rs` but your code is also a lot shorter.
IF there is a difference, make sure to also add a testcase for that difference, as that would mean that the tests are not 100% 😉?

esavier:
Yup, i will do what i can.
at a glance both options are okay. I do not think the other PR is necessarily disadventagious, but i am new here, i will check how the rest of the code looks like and what we can do. Please give me time until tomorrow.

PR #1575, opened 2026-06-28
fix: simplify utoipa-gen feature gates to fix jiff_0_2

siku2:
Enabling the "jiff_0_2" feature on its own would not correctly enable the more
elaborate is_primitive branch because it was mistakenly omitted from the feature gate list.
Rather than merely fixing the feature gate list, this commit simplifies the
logic to make such a mistake impossible in the future.


juhaku (approved):

juhaku:
Could you add CHANGELOG.md entry to `utoipa-gen` project

PR #1554, opened 2026-05-15
Fix unsigned integer known format serialization

MukundaKatta:
## Summary
- serialize unsigned integer known formats as `uint8`, `uint16`, `uint32`, and `uint64`
- update non-strict integer schema snapshots for `u32` and `u64`
- add regression coverage for unsigned known format serialization

Closes #1527

## Verification
- `cargo test -p utoipa test_partial_schema_strict_integers`
- `cargo test -p utoipa test_partial_schema_non_strict_integers --features non_strict_integers`
- `cargo test -p utoipa serialize_unsigned_integer_known_formats --features non_strict_integers`
- `cargo fmt --all --check`
- `git diff --check`

PR #1578, opened 2026-07-03
feat: added support for num types + dev tools fixes

Dzordzu:
### Added support for 

```rust
std::num::{NonZeroI8, NonZeroI16, NonZeroI32, NonZeroI64, NonZeroU8, NonZeroU16, NonZeroU32, NonZeroU64, NonZeroIsize, NonZeroUsize};
```
### Added `non_strict_integers` to test targets
It was failing on the master. Better to have it tested.

### Added `bacon + nextest` to configs
It's a better dev experience ;) 

## Why? 
Because these types are used by `typify`. And they are in the stdlib 

## Notes
* Similar to #1334 (didn't see that tbh), but includes NonZeroI* variants and dev tools fixes
* Includes same fixes as #1554 (again, didn't see that PR)

### DIFF
--- a/utoipa/src/lib.rs
+++ b/utoipa/src/lib.rs
@@ -1687,6 +1687,17 @@ mod tests {
         assert_compact_json_snapshot!(u16::schema(), @r#"{"type": "integer", "format": "uint16", "minimum": 0}"#);
         assert_compact_json_snapshot!(u32::schema(), @r#"{"type": "integer", "format": "uint32", "minimum": 0}"#);
         assert_compact_json_snapshot!(u64::schema(), @r#"{"type": "integer", "format": "uint64", "minimum": 0}"#);
+
+        assert_compact_json_snapshot!(NonZeroI8::schema(), @r#"{"type": "integer", "format": "int8"}"#);
+        assert_compact_json_snapshot!(NonZeroI16::schema(), @r#"{"type": "integer", "format": "int16"}"#);
+        assert_compact_json_snapshot!(NonZeroI32::schema(), @r#"{"type": "integer", "format": "int32"}"#);
+        assert_compact_json_snapshot!(NonZeroI64::schema(), @r#"{"type": "integer", "format": "int64"}"#);
+        assert_compact_json_snapshot!(NonZeroI128::schema(), @r#"{"type": "integer"}"#);
+        assert_compact_json_snapshot!(NonZeroIsize::schema(), @r#"{"type": "integer"}"#);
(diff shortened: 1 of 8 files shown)


utoipa/src/lib.rs
CommanderStorm, line 1679: Can you add the respective new cases here too?
I know that this is not going to change the correctness, but more as a way to know that we are not going to regress on this 😉

Dzordzu: Done

utoipa/src/lib.rs
```diff
@@ -1640,8 +1685,19 @@ mod tests {
...
-        assert_compact_json_snapshot!(u64::schema(), @r#"{"type": "integer", "format": "int64", "minimum": 0}"#);
+        assert_compact_json_snapshot!(u32::schema(), @r#"{"type": "integer", "format": "uint32", "minimum": 0}"#);
+        assert_compact_json_snapshot!(u64::schema(), @r#"{"type": "integer", "format": "uint64", "minimum": 0}"#);
+
+        assert_compact_json_snapshot!(NonZeroI8::schema(), @r#"{"type": "integer", "format": "int8"}"#);
+        assert_compact_json_snapshot!(NonZeroI16::schema(), @r#"{"type": "integer", "format": "int16"}"#);
+        assert_compact_json_snapshot!(NonZeroI32::schema(), @r#"{"type": "integer", "format": "int32"}"#);
+        assert_compact_json_snapshot!(NonZeroI64::schema(), @r#"{"type": "integer", "format": "int64"}"#);
+        assert_compact_json_snapshot!(NonZeroI128::schema(), @r#"{"type": "integer"}"#);
+        assert_compact_json_snapshot!(NonZeroIsize::schema(), @r#"{"type": "integer"}"#);
```
CommanderStorm, line 1696: On second thought:
Would this be better modeled by the union of an integer with an exclusive minimum and one with an exclusive maxium of 0?

I don't have a strong oppinion on this..

CommanderStorm (approved):
LGTM, thanks

utoipa/src/lib.rs
Dzordzu: Hm. Not a bad idea tbh. Will try to do it today!

Dzordzu: Sorry for the delay. I'll try to implement this in the upcoming days

CommanderStorm: Just mark the pr as a draft to Indicate that you are planning to make changes ^^

Issue #1505, opened 2025-12-04
Why this stack overflow?


frederikhors:
Using this code:

```rust
#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, utoipa::ToSchema)]
#[serde(untagged)]
pub enum PlayerFilters {
    And(Vec<Self>),
    Id(Id),
    CreatedAt(DateTime),
    CreatedBy(NullId),
    Number(Number<i64>),
    WithEquipment(Box<EquipmentFilters>),
    HasSkill(bool),
}
```

if I generate schema from utoipa I get this error:

```
thread 'main' (19548) has overflowed its stack
```

if I remove the `And(Vec<Self>),` it works! Why?

`Vec<>` moves things already on the heap: I don't need Box.

If I remove `#[serde(untagged)]` is the same.

https://github.com/juhaku/utoipa/pull/1504 does not fix.

zdcthomas:
I'm getting a similar error but only when trying to use the swagger ui. The struct I was seeing the problem with was self referential as well.

zdcthomas:
It turns out mine was also during generation, but I fixed it using this tag https://github.com/juhaku/utoipa/blob/cdff0074eb65b23ede02f68a965f51ad53a8f643/utoipa-gen/src/lib.rs#L337

PR #1504, opened 2025-11-29
Fix stack overflow

qarmin:
Fixes #1454

Temporary array was changed into temporary vector

Issue #1454, opened 2025-09-02
Bigger structs(but with small stack size) shows clippy warnings about `allocating a local array larger than 16384 bytes`


qarmin:
```
use chrono::{NaiveDate, NaiveDateTime};
use utoipa::{IntoParams};
#[derive(IntoParams)]
pub struct Bbb {
    b1: Option<f32>,
    b2: Option<i32>,
    b3: Option<bool>,
    b4: Option<String>,
    b5: f32,
    b6: i32,
    b7: bool,
    b8: String,
    b9: NaiveDateTime,
    b10: Option<NaiveDateTime>,
    b11: Vec<String>,
    b12: Option<f32>,
    b13: Option<i32>,
    b14: Option<bool>,
    b15: Option<String>,
    b16: f32,
    b17: i32,
    b18: bool,
    b19: String,
    b20: NaiveDateTime,
    b21: Option<NaiveDateTime>,
    b22: Vec<String>,
    b23: Option<f32>,
    b24: Option<i32>,
    b25: Option<bool>,
    b26: Option<String>,
    b27: f32,
    b28: i32,
    b29: bool,
    b30: String,
}
```
clippy compiled with
```
 cargo clippy -- -Dclippy::large_stack_arrays
```
shows this error
```
error: allocating a local array larger than 16384 bytes
  --> src/pdf_creator/model_exam.rs:19:10
   |
19 | #[derive(IntoParams)]
   |          ^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_stack_arrays
   = note: requested on the command line with `-D clippy::large-stack-arrays`
   = note: this error originates in the derive macro `IntoParams` (in Nightly builds, run with -Z macro-backtrace for more info)
```

This is a bit surprising — this struct alone takes roughly 400 bytes on the stack.
Can the generated code reduce the size of these structs?

frederikhors:
@qarmin did you find a way?

Drifter-J:
https://github.com/juhaku/utoipa/pull/1580

PR #1580, opened 2026-07-05
fix: emit vec![] instead of array literals to avoid large stack arrays

Drifter-J:
## Summary

Fixes the `clippy::large_stack_arrays` warning and runtime stack overflow when deriving `ToSchema` / `IntoParams` on types with many fields (#1454, #1505).

Builds on #1504 (qarmin's `IntoParams` fix, included here rebased onto latest `master`) and extends it to **all** generated code, including `ToSchema` - the derive users actually hit the overflow with.

## Changes

- **`utoipa-gen/src/lib.rs`** - the shared `Array` tokenizer now emits `vec![...]` instead of a stack array literal `[...]`. One central change covering every call site (`ToSchema`, `IntoParams`, path/openapi builders).
- **`utoipa-gen/src/component/schema/enums.rs`** - `PlainSchema` used `.enum_values::<[T; N], T>(...)`, a turbofish pinning the type to a fixed-size array; updated to `Vec<T>`.
- **`Cargo.toml` + per-crate manifests** - enforce `clippy::large_stack_arrays = "deny"` via `[workspace.lints]`; crates inherit with `[lints] workspace = true`.
- **`utoipa-gen/tests/schema_derive_test.rs`** - regression test deriving a wide (100-field) struct.
- **`utoipa-gen/CHANGELOG.md`** - changelog entry.

Credit to @qarmin for the original `IntoParams` fix in #1504.

### DIFF
--- a/utoipa-gen/tests/schema_derive_test.rs
+++ b/utoipa-gen/tests/schema_derive_test.rs
@@ -3651,3 +3651,35 @@ fn derive_inline_option_ref_with_nullable_false_and_default() {
 
     assert_json_snapshot!(schema);
 }
+
+// Deriving on a wide type must not emit a large stack array (juhaku/utoipa#1454).
+#[test]
+fn derive_schema_with_many_fields_does_not_use_large_stack_arrays() {
+    #![allow(unused)]
+
+    macro_rules! wide_struct {
+        ($name:ident; $($field:ident),+ $(,)?) => {
+            #[derive(ToSchema)]
+            struct $name {
+                $($field: String,)+
+            }
+        };
+    }
+
+    wide_struct!(WideStruct;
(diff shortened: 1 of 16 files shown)


utoipa-gen/tests/schema_derive_test.rs
```diff
@@ -3651,3 +3651,35 @@ fn derive_inline_option_ref_with_nullable_false_and_default() {
 
     assert_json_snapshot!(schema);
 }
+
+// Deriving on a wide type must not emit a large stack array (juhaku/utoipa#1454).
+#[test]
+fn derive_schema_with_many_fields_does_not_use_large_stack_arrays() {
```
Drifter-J: ~~doesn't really test the scenario, fixing it~~ rename to reflect the reality

PR #1581, opened 2026-07-12
Refactor documentation in lib.rs

Its-Just-Nans:
Updated documentation comments to remove commented-out code.

The goal is to show `use utoipa::openapi::OpenApi;`

Because if we use `use utoipa::OpenApi;` (a trait) this code produces an error


CommanderStorm (approved):

Drifter-J:
I checked doc with `cargo doc -p utoipa-axum --no-deps` and it lgtm. merged :) thanks

PR #1582, opened 2026-07-21
Recursively discover flattened fields in maps

### LINKED ISSUE #1330: #[serde(flatten)] Fields in HashMap Not Recursively Discovered
**Description**
When using `#[serde(flatten)]` on fields inside a container type like `HashMap<String, Bar>`, the `Bar` schema is not recursively discovered and doesn't appear in `component.schemas`. This causes issues when trying to generate correct OpenAPI specifications with the utoipa library.

**Example**
```rust
#[derive(Deserialize, ToSchema)]
pub struct Foo {
    #[serde(flatten)]
    pub data: HashMap<String, Bar>,
}
#[derive(ToSchema)]
pub struct Bar {
    value: i64,
}
fn main() {
    #[derive(OpenApi)]
    #[openapi(components(schemas(Foo)))]
    struct ApiDoc;
    let spec = serde_json::to_value(ApiDoc::openapi()).unwrap();
    assert_eq!(
        spec,
        json!(
            {
              "components": {
                "schemas": {
                  "Foo": {
                    "additionalProperties": {
                      "$ref": "#/components/schemas/Bar"
                    },
                    "type": "object"
                  }
                }
              },
              "info": {
                "description": "",
                "license": {
                  "name": ""
                },
                "title": "bug-example",
                "version": "0.1.0"
              },
              "openapi": "3.1.0",
              "paths": {}
            }
        )
    );
}
```

**What's wrong**
I expect assert to fail because there should be a `Bar` schema inside `components.schemas`, but there isn't.

**Workaround**
A workaround is to explicitly list the missing schema using the `#[openapi(components(schemas(...)))]` macro. However, this approach is inconvenient for large projects where missing schemas may be deeply nested inside models. It's inconsistent to have to manually mention each schema in the components section.

**Environment**
- utoipa 5.3.1
- serde 1.0.218
- cargo 1.84.0 nightly

Any help or suggestions on how to address this would be greatly appreciated. Thanks!

nipunn1313:
This allows the discovered fields to appear in component.schemas without having to explicitly list it as a workaround.

Ran into this issue in our codebase, and confirmed that patching in this fixes.

Fixes #1330

### DIFF
--- a/utoipa-gen/src/component.rs
+++ b/utoipa-gen/src/component.rs
@@ -1576,7 +1576,7 @@ impl<'a, T> Iterator for ChildRefIter<'a, T> {
 #[cfg_attr(feature = "debug", derive(Debug))]
 pub struct FlattenedMapSchema {
     tokens: TokenStream,
-    pub schema_references: Vec<SchemaReference>,
+    schema_references: Vec<SchemaReference>,
 }
 
 impl FlattenedMapSchema {
--- a/utoipa-gen/src/component.rs
+++ b/utoipa-gen/src/component.rs
@@ -1576,6 +1576,7 @@ impl<'a, T> Iterator for ChildRefIter<'a, T> {
 #[cfg_attr(feature = "debug", derive(Debug))]
 pub struct FlattenedMapSchema {
     tokens: TokenStream,
+    pub schema_references: Vec<SchemaReference>,
 }
 
 impl FlattenedMapSchema {
@@ -1598,7 +1599,7 @@ impl FlattenedMapSchema {
         // additionalProperties denoting the type
         // maps have 2 child schemas and we are interested the second one of them
         // which is used to determine the additional properties
-        let schema_property = ComponentSchema::new(ComponentSchemaProps {
+        let mut schema_property = ComponentSchema::new(ComponentSchemaProps {
             container,
             type_tree: type_tree
                 .children
@@ -1609,6 +1610,7 @@ impl FlattenedMapSchema {
             features,
             description: None,
         })?;
+        let schema_references = std::mem::take(&mut schema_property.schema_references);
         let schema_tokens = schema_property.to_token_stream();
 
         tokens.extend(quote_diagnostics! {
@@ -1621,7 +1623,10 @@ impl FlattenedMapSchema {
         example.to_tokens(&mut tokens)?;
         nullable.to_tokens(&mut tokens)?;
 
-        Ok(Self { tokens })
+        Ok(Self {
+            tokens,
+            schema_references,
+        })
     }
 }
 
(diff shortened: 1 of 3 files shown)


utoipa-gen/src/component.rs
```diff
@@ -1576,6 +1576,7 @@ impl<'a, T> Iterator for ChildRefIter<'a, T> {
 #[cfg_attr(feature = "debug", derive(Debug))]
 pub struct FlattenedMapSchema {
     tokens: TokenStream,
+    pub schema_references: Vec<SchemaReference>,
```
CommanderStorm: why not:

```suggestion
    schema_references: Vec<SchemaReference>,
```

CommanderStorm: likey because utoipa-gen/src/component/schema.rs, or?

utoipa-gen/tests/schema_derive_test.rs
```diff
@@ -104,6 +104,61 @@ fn derive_flattened_map_ref_property() {
...
+        "type": "object"
+      },
+      "Foo": {
+        "additionalProperties": {
+          "$ref": "#/components/schemas/Bar"
+        },
+        "type": "object"
+      }
+    }
+    "###);
```
CommanderStorm: I think the "established" way of doing this testing is to use `assert_value` and non-inline snapshots.

Could you change it to that?

utoipa-gen/src/component.rs
nipunn1313: was mistake. I assumed it was necessary because of schema.rs - but it's not because it's a submodule. Made it private.

utoipa-gen/tests/schema_derive_test.rs
nipunn1313: good suggestion. Much better. Thanks.

utoipa-gen/tests/schema_derive_test.rs
```diff
@@ -104,6 +104,36 @@ fn derive_flattened_map_ref_property() {
...
+
+    #[derive(OpenApi)]
+    #[openapi(components(schemas(Foo)))]
+    struct ApiDoc;
+
+    let doc = serde_json::to_value(ApiDoc::openapi()).unwrap();
+
+    // `Bar` referenced through the flattened `HashMap` must be discovered recursively
+    // and appear in `components.schemas` without being listed explicitly.
+    assert_value! { doc =>
```
CommanderStorm: don't think this adds much value...

```suggestion
    assert_value! { doc =>
```

CommanderStorm (approved):
lgtm now, if you remove the comment

nipunn1313:
removed comment. Good call.

CommanderStorm (approved):
lgtm

Drifter-J:
Thanks @nipunn1313 & @CommanderStorm, the change lgtm, merging now :)

Drifter-J:
Ah forgot to ask for changelog, I will raise it 😅 - https://github.com/juhaku/utoipa/pull/1596

nipunn1313:
thx

PR #1583, opened 2026-07-30
Attempt to fix tests for 'repr' feature

fnimick:
Note that the tests do not pass as is, as the api docs generated for a 'tagged repr enum' do not appear to match serde_repr's runtime behavior, as validated inline in the test.

Please **do not merge!**

### DIFF
--- a/utoipa-gen/tests/snapshots/schema_derive_test__derive_schema_for_tagged_repr_enum.snap.new
+++ b/utoipa-gen/tests/snapshots/schema_derive_test__derive_schema_for_tagged_repr_enum.snap.new
@@ -1,0 +1,51 @@
+---
+source: utoipa-gen/tests/schema_derive_test.rs
+assertion_line: 2397
+expression: value
+---
+{
+  "oneOf": [
+    {
+      "properties": {
+        "tag": {
+          "enum": [
+            0
+          ],
+          "type": "integer"
+        }
+      },
+      "required": [
+        "tag"
+      ],
+      "type": "object"
+    },
+    {
+      "properties": {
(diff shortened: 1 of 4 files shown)


utoipa-gen/tests/snapshots/schema_derive_test__derive_schema_for_tagged_repr_enum.snap.new
CommanderStorm, line 1: please `insta accept` the snapshots and commit the result. `.new` is just the proto-snapshot and needs to be renamed first

fnimick: This is the one where the generated openapi doesn't match the serde_repr runtime behavior, so the snapshot output here is *wrong* (potentially). That's why I left it in this state.

fnimick: @CommanderStorm  I wasn't sure what to do here in the case where I'm not sure whether an existing behavior is a bug or not, as I've never used repr enums but the output seems to not make sense. Should I open an issue?

CommanderStorm: if the behaviour is wrong, that would be best.

fnimick: I forgot that I already did: #1584

PR #1585, opened 2026-07-30
Support partially tagged enums

fnimick:
Heavily based on #1521 - merge conflicts fixed, with additional tests added for the internally tagged and adjacently tagged serde options.

In addition, this now supports untagged members on plain enums, which were not implemented in the initial PR.

---

### Original PR Message

See here: https://serde.rs/variant-attrs.html#untagged

> Irrespective of the enum representation, serialize and deserialize this variant as untagged, i.e. simply as the variant's data with no record of the variant name.

For a simple example:

```rust
enum SomeEnum {
    Tagged(u32)
    #[serde(untagged)]
    Untagged(u32)
}
```

This would serialize to an optionally tagged integer:

```json
{"Tagged": 3}
3
```

With resulting JSON Schema:

```json
{
    "oneOf": [
        {"type": "object", "properties": {"Tagged": {"type": "integer"} }, "required": ["Tagged"] },
        {"type": "integer"}
    ]
}
```

---

Somewhat related, but __NOT__ implemented in this PR (as it seems like it will take some deeper refactoring of the "plain enum" handling):

```rust
enum SomeEnum {
    One,
    Two,
    #[serde(untagged)]
    Null
}
```

For this enum, serde would emit JSON values `"One"`, `"Two"` and `null`, but utoipa still generates a JSON schema for an enum:

```json
{
    "enum": ["One", "Two", "Null"],
    "type": "string"
}
```

### DIFF
--- a/utoipa-gen/src/component/schema.rs
+++ b/utoipa-gen/src/component/schema.rs
@@ -855,15 +855,18 @@ impl<'e> EnumSchema<'e> {
         parent: &'e Root<'e>,
         variants: &'e Punctuated<Variant, Comma>,
     ) -> Result<Self, Diagnostics> {
-        let all_unit_not_partially_tagged = variants
+        let variants_without_skipped = variants
             .iter()
-            .map(|variant| {
-                Ok(matches!(variant.fields, Fields::Unit)
-                    && !serde::parse_value(&variant.attrs)?.untagged)
+            .filter_map(|variant| match serde::parse_value(&variant.attrs) {
+                Ok(rules) if rules.skip => None,
+                Ok(rules) => Some(Ok((variant, rules))),
+                Err(diagnostics) => Some(Err(diagnostics)),
             })
-            .collect::<Result<Vec<bool>, Diagnostics>>()?
-            .into_iter()
-            .all(|is_plain| is_plain);
+            .collect::<Result<Vec<_>, Diagnostics>>()?;
+
+        let all_unit_not_partially_tagged = variants_without_skipped
+            .iter()
+            .all(|(variant, rules)| matches!(variant.fields, Fields::Unit) && !rules.untagged);
 
         if all_unit_not_partially_tagged {
             #[cfg(feature = "repr")]
@@ -909,7 +912,7 @@ impl<'e> EnumSchema<'e> {
             }
 
             Ok(Self {
-                schema_type: EnumSchemaType::Plain(PlainEnum::new(parent, variants, features)?),
+                schema_type: EnumSchemaType::Plain(PlainEnum::new(parent, variants_without_skipped, features)?),
                 schema_as,
                 schema_references: Vec::new(),
                 bound,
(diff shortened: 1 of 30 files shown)


Drifter-J:
Hi, thanks for contribution - https://github.com/fnimick/utoipa/blob/58eb1dbda4a04831e899eadd1be943e6d03296cb/utoipa-gen/src/lib.rs#L489 would you mind changing this?

Plus please add the changelog :)

utoipa-gen/src/component/schema.rs
```diff
@@ -855,10 +855,17 @@ impl<'e> EnumSchema<'e> {
...
         variants: &'e Punctuated<Variant, Comma>,
     ) -> Result<Self, Diagnostics> {
-        if variants
+        let all_unit_not_partially_tagged = variants
             .iter()
-            .all(|variant| matches!(variant.fields, Fields::Unit))
-        {
+            .map(|variant| {
+                Ok(matches!(variant.fields, Fields::Unit)
+                    && !serde::parse_value(&variant.attrs)?.untagged)
```
Drifter-J: suggestion: I read the whole related code more carefully so from what I understood https://github.com/fnimick/utoipa/blob/58eb1dbda4a04831e899eadd1be943e6d03296cb/utoipa-gen/src/component/schema/enums.rs#L63C13-L63C70 is executed the duplicated per-variant parse. Do you think we can eliminate this? If not, we can keep it as it is.

fnimick: Done

fnimick:
@Drifter-J  I realized a potential issue in the output. An enum with the format:
```
enum Foo {
    One,
    #[serde(untagged)]
    Two(String),
    #[serde(untagged)]
    Three(String),
}
```

generates an openapi schema of:
```
{
  "oneOf": [
    {
      "enum": [
        "One"
      ],
      "type": "string"
    },
    {
      "type": "string"
    }
  ]
}
```

Is this an issue, that the untagged string branch is duplicated?

A similar issue happens with partially untagged unit enums, where there are duplicate `null` members in the ouput `oneOf` if multiple untagged variants are present.

Drifter-J:
> @Drifter-J I realized a potential issue in the output. An enum with the format:
> 
> ```
> enum Foo {
>     One,
>     #[serde(untagged)]
>     Two(String),
>     #[serde(untagged)]
>     Three(String),
> }
> ```
> 
> generates an openapi schema of:
> 
> ```
> {
>   "oneOf": [
>     {
>       "enum": [
>         "One"
>       ],
>       "type": "string"
>     },
>     {
>       "type": "string"
>     }
>   ]
> }
> ```
> 
> Is this an issue, that the untagged string branch is duplicated?
> 
> A similar issue happens with partially untagged unit enums, where there are duplicate `null` members in the ouput `oneOf` if multiple untagged variants are present.

Yes, it doesn't look like a valid schema and it seems like Three is collapsed as well. I would suggest fixing duplication at codegen time

PR #1588, opened 2026-08-07
feat: support generic types with `IntoParams`

### LINKED ISSUE #1546: `IntoParams` does not support types with generics
On version 5.5, the following example

```rust
#[derive(utoipa::ToSchema, utoipa::IntoParams)]
struct Foo<R: utoipa::ToSchema> {
    x: Option<R>,
}
```

produces the compiler error:

```
error[E0425]: cannot find value `generics` in this scope
 --> src/main.rs:8:15
  |
8 |     x: Option<R>,
  |               ^ not found in this scope
```

I'm not sure if this is intended and generics are not supported, or if this is a bug.

jack-ullery:
Fixes: https://github.com/juhaku/utoipa/issues/1546

This PR implements support for generic types with `IntoParams`.

For example, with the following code:
```rust
#[derive(IntoParams)]
struct QueryParams<T: ToSchema> {
    param: T,
}
```

The previous implementation would not compile with the following error:
```rust
error[E0425]: cannot find value `generics` in this scope
 --> src/main.rs:8:12
  |
8 |     param: T,
  |            ^ not found in this scope
```

This PR makes a minor fix to support that use-case.


utoipa-gen/src/component/into_params.rs
```diff
@@ -157,9 +157,11 @@ impl ToTokensDiagnostics for IntoParams {
             })
             .collect::<Result<Vec<TokenStream>, Diagnostics>>()?;
 
+        let generic_schemas = get_generic_schemas(&self.generics);
         tokens.extend(quote! {
             impl #impl_generics utoipa::IntoParams for #ident #ty_generics #where_clause {
                 fn into_params(parameter_in_provider: impl Fn() -> Option<utoipa::openapi::path::ParameterIn>) -> Vec<utoipa::openapi::path::Parameter> {
+                    let mut generics: Vec<utoipa::openapi::RefOr<utoipa::openapi::Schema>> = vec![ #(#generic_schemas),* ];
```
Drifter-J, line 164: Could you double check if when no generics is used, no warning is surfaced on the user's side? if yes, mind if you add something like

`#[allow(unused_mut, unused_variables)]`

Drifter-J:
thanks for working on this :) @jack-ullery could you add changelog plz?

utoipa-gen/src/component/into_params.rs
jack-ullery: Good point!

At least when I build the unit-tests, the only warnings I get are:

<details>

<summary> cargo test warnings </summary>

```rust
warning: use of deprecated function `utoipa::__dev::warn_deprecated_ignore_fn_pattern`: Using a function path for `ignore` is deprecated and will be removed in a future version. Use `ignore = true` or `ignore = false` instead.
    --> utoipa-gen/tests/schema_derive_test.rs:3170:31
     |
3170 |             #[schema(ignore = Self::ignore)]
     |                               ^^^^
     |
     = note: `#[warn(deprecated)]` on by default

warning: use of deprecated function `utoipa::__dev::warn_deprecated_ignore_fn_pattern`: Using a function path for `ignore` is deprecated and will be removed in a future version. Use `ignore = true` or `ignore = false` instead.
    --> utoipa-gen/tests/schema_derive_test.rs:3234:31
     |
3234 |             #[schema(ignore = always_true)]
     |                               ^^^^^^^^^^^

warning: use of deprecated function `utoipa::__dev::warn_deprecated_ignore_fn_pattern`: Using a function path for `ignore` is deprecated and will be removed in a future version. Use `ignore = true` or `ignore = false` instead.
    --> utoipa-gen/tests/path_derive.rs:1958:26
     |
1958 |         #[param(ignore = always_true)]
     |                          ^^^^^^^^^^^
     |
     = note: `#[warn(deprecated)]` on by default
```
</details>

And I also don't get a warning with the following code:
<details>
<summary> Modified version of  "examples/simple-axum" </summary>

```rust
#![deny(
    clippy::pedantic,
    unused_mut,
    unused_variables
)]

use std::net::SocketAddr;

use axum::{routing::get, Json};
use utoipa::{IntoParams, OpenApi, ToSchema};

#[derive(ToSchema, Debug)]
enum Bird {
    Sparrow,
}

#[derive(IntoParams, ToSchema, Debug)]
struct QueryParams {
    animal: Bird,
}

/// Return JSON version of an OpenAPI schema
#[utoipa::path(
    get,
    path = "/api-docs/birds.json",
    responses(
        (status = 200, description = "JSON file", body = ())
    ),
    params(QueryParams),
)]
async fn birds() -> Json<utoipa::openapi::OpenApi> {
    Json(ApiDoc::openapi())
}

#[derive(OpenApi)]
#[openapi(paths(birds), components(schemas(Bird)))]
struct ApiDoc {}

#[tokio::main]
async fn main() {
    let socket_address: SocketAddr = "127.0.0.1:8080".parse().unwrap();
    let listener = tokio::net::TcpListener::bind(socket_address).await.unwrap();

    let app = axum::Router::new().route("/api-docs/birds.json", get(birds));

    // Create and print a QueryParams object to avoid "unused" warnings,
    // without relying on `#[allow(...)]`
    let fake_param = QueryParams{animal: Bird::Sparrow};
    println!("{:?}", fake_param.animal);

    axum::serve(listener, app.into_make_service())
        .await
        .unwrap();
}
```
</details>

Which is really weird, because if I print the TokenStream generated by `into_params()` in [utoipa-gen/src/lib.rs](https://github.com/jack-ullery/utoipa/blob/510568955cd401703fff6b33529e456406de88de/utoipa-gen/src/lib.rs#L2595) we see:

```rust
impl utoipa::IntoParams for QueryParams {
    fn into_params(
        parameter_in_provider: impl Fn() -> Option<utoipa::openapi::path::ParameterIn>,
    ) -> Vec<utoipa::openapi::path::Parameter> {
        let mut generics: Vec<utoipa::openapi::RefOr<utoipa::openapi::Schema>> = vec![];
        vec![Some(
            utoipa::openapi::path::ParameterBuilder::new()
                .name("_animal")
                .parameter_in(parameter_in_provider().unwrap_or_default())
                .required(utoipa::openapi::Required::True)
                .schema(Some(
                    utoipa::openapi::schema::RefBuilder::new().ref_location_from_schema_name(
                        format!("{}", <Bird as utoipa::ToSchema>::name()),
                    ),
                ))
                .build(),
        )]
        .into_iter()
        .filter(Option::is_some)
        .flatten()
        .collect()
    }
}
```

Which should definitely throw that warning you mentioned, but I don't see it when using `#[derive(IntoParams)]`

jack-ullery: Happy to add that `allow` attribute to be safe though, if that's what you prefer

jack-ullery:
> thanks for working on this :) @jack-ullery could you add changelog plz?

@Drifter-J Done! https://github.com/juhaku/utoipa/pull/1588/commits/e2285eae64cae84170897312c72b32066c2beed8

If you can think of something more readable lmk

utoipa-gen/src/component/into_params.rs
Drifter-J: interesting, if this is not fired, I think it's ok to skip! thanks for checking tho :)

Paul-E:
Thank you!

PR #1592, opened 2026-08-09
feat(gen): support item_schema attribute for streaming/SSE response types

Drifter-J:
## Summary
* Adds support for the media type `item_schema = ...` attribute in `responses(...)` and `request_body(...)`.
* Maps to OpenAPI 3.2's [`itemSchema`](https://spec.openapis.org/oas/v3.2.0.html#fixed-fields-11) media type keyword for streaming, sequential, or event-based media types such as `text/event-stream`.
* Documents the new `item_schema` attribute and adds a fully compiling `openapi-3_2-sse` Server-Sent Events example using axum.
* Change `MediaTypeAttr` Docs to be prettier and informative. 

## Reverted SSE example but you can see the stream of infinite pet events.
https://github.com/user-attachments/assets/fa711c34-7542-40f2-9f91-bf44f1413904

## Current version
Screenshot 2026-08-09 at 23 34 46

Reason why I reverted infinite stream;
* Swagger-UI `Execute` button buffers the entire response body before rendering it, so it cannot display an SSE stream incrementally (it only shows a spinner until the stream completes). So now, the example emits a finite-number-of-events (10, one per-second) so, `Execute` eventually returns :)

### DIFF
--- a/examples/openapi-3_2-sse/Cargo.toml
+++ b/examples/openapi-3_2-sse/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "openapi-3_2-sse"
+description = "Axum example documenting a Server-Sent Events (SSE) endpoint using OpenAPI 3.2"
+version = "0.1.0"
+edition = "2021"
+license = "MIT"
+authors = ["Drifter-J <latenytcoffee@gmail.com>"]
+
+[dependencies]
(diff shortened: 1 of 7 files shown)


juhaku (approved):
Great work, 🚀

PR #1474, opened 2025-09-27
Support extensions in `utoipa::IntoParams` attribute macros

aybeshlikyan:
Hello! This is my PR addressing the feature I proposed in Issue #1473, adding support to specifying OpenAPI extensions   on parameters using the `#[into_params]` and `#[param]` attribute macros for structs that derive `utoipa::IntoParams`.
The implementation follows the form proposed in the issue:

```rust
#[derive(Debug, Clone, utoipa::IntoParams)]
#[into_params(parameter_in = Query, extensions(("x-on-all-params" = json!(true))))]
pub struct ParamStruct {
    /// A param.
    pub param_1: i32,

    /// Another param.
    #[param(extensions(("x-on-one-param" = json!({ "key": "value" }))))]
    pub param_2: String,
}
```

I updated the macro documentation, added unit tests for the various flavors, and updated the example introduced with the feature in #1292, `simple-x-extensions`.

I couldn't find any special docs about contribution guidelines, please let me know if I'm missing anything!

Closes issue #1473

PR #1292, opened 2025-01-15
Issue #1228 - Feature: support extensions in utoipa::path macro

juliendecharentenay:
Hi,
This is a first attempt at implementing extensions within the `utoipa::path` macro. Read the PR in conjunction with the issue #1228 .

The `extensions` is implemented for the following parts of the `utoipa::path` macro:
- At operation level;
- Operation > Params > ParamItem
- Operation > Responses
- Operation > Responses > ContentType
- Operation > RequestBody
- Operation > RequestBody > ContentType

The parsing and building of the token stream is done in a struct `Extensions` - which I located under `components/features/attributes`. I decided to split it into its own file that is loaded in `attributes.rs` as `attributes.rs` is already quite long. But not sure (a) if it makes sense as a child of `attributes` and (b) if ok to split into its own separate file.

What is included:
- Tests [for the happy path]
- An example. The example is showing both macro and Modify trait approaches [this was a way for me to check which one were replaceable] - but having both there may be too much.

PR #1593, opened 2026-08-17
Support extensions in `utoipa::ToSchema` named structs

TilBlechschmidt:
Hi! 👋 

With this PR I would like to add the support for OpenAPI extensions to the `ToSchema` macro. The functionality is limited to named structs as I was unsure about the implications of enabling it for the other kinds as well.

```rust
#[derive(ToSchema)]
struct HelloWorld {
    #[schema(extensions(
        ("x-terraform-requires-replace" = json!(true))
    ))]
    id: u32
}
```

Due to the prior work, the change is rather minimal and simply ties in the existing attribute parser into the schema macro.

---

Prior work:
- PR #1292 added support for OpenAPI extensions. This is currently limited to the `path` macro.
- PR #1474 extends this with support for the `IntoParams` macro.


juhaku:
Looks good, just missing CHANGELOG entry :100:

TilBlechschmidt:
Ah there is always something one overlooks isn't there ^^

Changelog entry added, thanks for the review! 🙂

PR #1597, opened 2026-08-27
Upgrade utoipa-gen to syn 3

jakobhellermann:
Can reduce dependencies if you have syn 3 already in the tree.
`syn` 3.0 has a MSRV of `1.71`, which is lower than this crate.


juhaku (approved):
Looks good 🎉

PR #1601, opened 2026-09-11
Add content_type attribute to IntoParams params

infiniteregrets:
- This PR adds a `content_type = "..."` attribute to `IntoParams #[param(...)]`
- Macro based on work done in `Parameter::content` model in #1555

### DIFF
--- a/utoipa-gen/CHANGELOG.md
+++ b/utoipa-gen/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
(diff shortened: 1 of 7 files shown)


juhaku (commented):
Nice, the `content_type` attribute on param works as handy shortcut to add support for `content` object, but that does not fully conform the how content schema is formed. Content should support way of adding multiple content type schemas like it is implemented in responses. 

IMHO for coherence it should be here implemented in similar way under `content( .... )` attribute like done in resopnses.

PR #1603, opened 2026-09-15
Re-enable actix auto_into_responses Result test

snaggen:
**Problem:** `path_operation_auto_types_responses` has been commented out as failing to compile since #1252. The `justfile` enables `auto_into_responses` for `utoipa-gen` but not `utoipa/auto_into_responses`, which gates `impl IntoResponses for Result<T, E>`.

**Fix:** Uncomment the test, restore its imports and enable `utoipa/auto_into_responses` in the `justfile`. The snapshot matches the original assertion (`200`, `404`, `500`).

**Related:** conflicts with #1609 (adjacent `justfile` lines). The other commented-out test is fixed in #1608.

AI-assisted


juhaku (approved):
Seems good 🎉

